PR #20, opened 2014-04-26
Tuxwave fixups

dl1kse:
Have made some changes to tlf:

-Use a S&P Callsign, so that possibly the speed of it can be controlled with +  and -

(I needed the feature of the S&P Callsign because the last time i used
tlf (the first one time also), i used a optocoupler attached to the
serial port. Also i was working at DB0RC, which callsign often got
confused with DD0RC, so i needed so put some String, like D--B++0RC, to
make the B better audible.)

-This callsign is also being used with the message key
-Added ? functionality to send incomplete (or even complete) callsigns appended with ?

Needed is:

Add S&P_CW_CALL in the rule files
Check ph_message[4] is valid (should be a sound file requesting for the callsign)

### DIFF
--- a/src/writecabrillo.c
+++ b/src/writecabrillo.c
@@ -624,6 +624,7 @@ int write_cabrillo(void)
 
     /* open logfile and create a cabrillo file */
     strcpy(cabrillo_tmp_name, call);
+    removechars(cabrillo_tmp_name, "+-");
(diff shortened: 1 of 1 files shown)


sebastiankricner:
Thanks for the nice information.

dl1jbe:
Yw. Maybe you find time to test if it works as expected for your problem.

PR #29, opened 2015-01-11
pull waedc-qtc branch please

dl1jbe:
I made a pull request so we can add comments to the code.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -70,6 +70,7 @@
 #include "lancode.h"
 #include "rtty.h"
 #include "ui_utils.h"
+#include "qtcvars.h"
 
 #define TUNE_UP 6	/* tune up for 6 s (no more than 10) */
 
@@ -138,7 +139,6 @@ char callinput(void)
     extern int keyerport;
     extern int miniterm;
     extern int no_rst;
-    extern int qtcdirection;
 
     int cury, curx;
     int i, j, ii, rc, t, x = 0, y = 0;
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -52,6 +52,7 @@
 #include <hamlib/rig.h>
 #endif
 #include "ui_utils.h"
+#include "qtcvars.h"
 
 #define MULTS_POSSIBLE(n) ((char *)g_ptr_array_index(mults_possible, n))
 
@@ -78,7 +79,6 @@ int changepars(void)
     extern int ctcomp;
     extern char *config_file;
     extern int miniterm;
-    extern int qtcdirection;
     extern int total;
 
 #ifdef HAVE_LIBHAMLIB
--- a/src/deleteqso.c
+++ b/src/deleteqso.c
@@ -34,12 +34,10 @@
 #include "clear_display.h"
 #include "scroll_log.h"
 #include "qtcutil.h"
+#include "qtcvars.h"
 
 #define QTCRECVCALLPOS 30
 #define QTCSENTCALLPOS 35
-
-extern int qtcdirection;
-extern int nr_qtcsent;
 
 void delete_qso(void)
 {
--- a/src/genqtclist.c
+++ b/src/genqtclist.c
@@ -23,6 +23,7 @@
 
 #include "globalvars.h"
 #include "genqtclist.h"
+#include "qtcvars.h"
 
 int genqtclist(char * callsign, int nrofqtc)
 {
--- a/src/getexchange.c
+++ b/src/getexchange.c
@@ -51,6 +51,7 @@
 #include "keyer.h"
 #include "rtty.h"
 #include "ui_utils.h"
+#include "qtcvars.h"
 
 #define MULTS_POSSIBLE(n) ((char *)g_ptr_array_index(mults_possible, n))
 #define LEN(array) (sizeof(array) / sizeof(array[0]))
@@ -107,7 +108,6 @@ int getexchange(void)
     extern int keyerport;
     extern int commentfield;
     extern int no_rst;
-    extern int qtcdirection;
     extern int serial_or_section;
     extern int ve_cty;
     extern int w_cty;
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -120,19 +120,6 @@ extern int three_point;
 extern int dxped;
 extern int addzone;
 
-extern int qsoflags_for_qtc[MAX_QSOS];	// array of flag to log lines of QSOs for QTC's;
-					// this is an array of flags, which marks when a QSO sent as QTC
-extern int nr_qsosflags_for_qtc;	// number of lines in qsoflags_for_qtc[]
-extern int next_qtc_qso;		// the next non-sent QSO, which can be send next as QTC
-extern t_qtclist qtclist;		// the QTC list to send
-extern int nr_qtcsent;
-extern t_qtcreclist qtcreclist;		// the QTC list which received
-
-extern int qtcdirection;				// QTC direction - 1: RECV, 2: SEND
 extern t_pfxnummulti pfxnummulti[MAXPFXNUMMULT];	// array of PFX_NUM_MULTIS parameter
 extern int pfxnummultinr;				// length of array of PFX_NUM_MULTIS parameter
 extern int continentlist_only;				// CONTINENT_LIST_ONLY parameter
-extern t_qtc_ry_line qtc_ry_lines[QTC_RY_LINE_NR];	// when QTC is set, and mode is RTTY, then the modem lines stored this array
-extern int qtc_ry_currline;				// current line of QTC RTTY modem
-extern int qtc_ry_capture;				// enable or disable QTC RTTY capturing
-extern int qtc_ry_copied;				// stores the number of copied lines from QTC RTTY terminal to QTC window
--- a/src/log_recv_qtc_to_disk.c
+++ b/src/log_recv_qtc_to_disk.c
@@ -32,6 +32,7 @@
 #include <stdlib.h>
 #include "tlf.h"
 #include "lancode.h"
+#include "qtcvars.h"
 
 extern int trx_control;
 extern float freq;
--- a/src/main.c
+++ b/src/main.c
@@ -36,6 +36,7 @@
 #endif
 #include "fldigixmlrpc.h"
 #include "ui_utils.h"
+#include "qtcvars.h"
 
 SCREEN *mainscreen;
 SCREEN *packetscreen;
@@ -332,19 +333,6 @@ char simulator_tone[5];
 /*-------------------------------the log lines-----------------------------*/
 char qsos[MAX_QSOS][LOGLINELEN+1];
 int nr_qsos = 0;
-
-int qsoflags_for_qtc[MAX_QSOS];
-int nr_qsosflags_for_qtc;
-int next_qtc_qso;
-t_qtclist qtclist;
-int nr_qtcsent = 0;
-t_qtcreclist qtcreclist;
-struct t_qtc_store_obj *qtc_temp_obj;
-int qtcdirection = 0;
-t_qtc_ry_line qtc_ry_lines[QTC_RY_LINE_NR];
-int qtc_ry_currline = 0;
-int qtc_ry_capture;
-int qtc_ry_copied;
 
 /*------------------------------dupe array---------------------------------*/
 int nr_worked = 0;		/*< number of calls in worked[] */
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -37,6 +37,7 @@
 #include "bandmap.h"
 #include "locator2longlat.h"
 #include "dxcc.h"
+#include "qtcvars.h"
 
 extern int keyerport;
 extern char tonestr[];
@@ -246,7 +247,6 @@ int parse_logcfg(char *inputbuffer)
     extern char sc_volume[];
     extern char modem_mode[];
     extern int no_rst;
-    extern int qtcdirection;
     extern int serial_or_section;
 
 /* LZ3NY mods */
@@ -293,7 +293,6 @@ int parse_logcfg(char *inputbuffer)
     extern int qtcrec_record;
     extern char qtcrec_record_command[2][50];
     extern char qtcrec_record_command_shutdown[50];
-    extern t_qtc_ry_line qtc_ry_lines[QTC_RY_LINE_NR];
 
     char commands[MAX_COMMANDS][30] = {
 	"enable",		/* 0 */		/* deprecated */
--- a/src/qtcutil.c
+++ b/src/qtcutil.c
@@ -29,9 +29,6 @@
 GHashTable* qtc_store = NULL; 	/* stores number of QTC's per callsign */
 struct t_qtc_store_obj *qtc_empty_obj = NULL;
 
-extern int qtcdirection;
-extern struct t_qtc_store_obj *qtc_empty_obj;
-
 void qtc_init() {
     if (qtc_store != NULL) {
 	g_hash_table_destroy(qtc_store);
--- a/src/qtcwin.c
+++ b/src/qtcwin.c
@@ -1,6 +1,6 @@
 /*
  * Tlf - contest logging program for amateur radio operators
- * Copyright (C) 2013-2014 Ervin Hegedüs - HA2OS <airween@gmail.com>
+ * Copyright (C) 2013-2015 Ervin Hegedüs - HA2OS <airween@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -45,6 +45,9 @@
 
 #include <sys/time.h>
 
+/* check direction clause macro
+ * direction should be RECV (1) or SEND (2), see tlf.h
+ */
 #define DIRCLAUSE (direction == RECV) || (direction == SEND && (activefield == 0 || activefield == 2))
 
 extern char hiscall[];
@@ -121,6 +124,17 @@ char * qtccallsign;
 int * qtccount;
 int * qtccurrdiretion;
 
+void fill_qtc_callsign(int direction, char * tcall) {
+    if (direction == RECV) {
+	strncpy(qtcreclist.callsign, tcall, strlen(tcall));
+	qtcreclist.callsign[strlen(tcall)] = '\0';
+    }
+    if (direction == SEND) {
+	strncpy(qtclist.callsign, tcall, strlen(tcall));
+	qtclist.callsign[strlen(tcall)] = '\0';
+    }
+}
+
 int qtc_main_panel(int direction) {
     char qtchead[32], tempc[40];
     int i, j, x;
@@ -146,24 +160,10 @@ int qtc_main_panel(int direction) {
     static int record_run = -1;
 
     if (strlen(hiscall) > 0) {
-	if (direction == RECV) {
-	    strncpy(qtcreclist.callsign, hiscall, strlen(hiscall));
-	    qtcreclist.callsign[strlen(hiscall)] = '\0';
-	}
-	if (direction == SEND) {
-	    strncpy(qtclist.callsign, hiscall, strlen(hiscall));
-	    qtclist.callsign[strlen(hiscall)] = '\0';
-	}
+	fill_qtc_callsign(direction, hiscall);
     }
     else if (strlen(lastcall) > 0) {
-	if (direction == RECV) {
-	    strncpy(qtcreclist.callsign, lastcall, strlen(lastcall));
-	    qtcreclist.callsign[strlen(lastcall)] = '\0';
-	}
-	if (direction == SEND) {
-	    strncpy(qtclist.callsign, lastcall, strlen(lastcall));
-	    qtclist.callsign[strlen(lastcall)] = '\0';
-	}
+	fill_qtc_callsign(direction, lastcall);
     }
 
     qtccurrdiretion = &direction;
--- a/src/qtcwin.c
+++ b/src/qtcwin.c
@@ -120,6 +120,7 @@ char help_send_msgs[7][26] = {
 };
 
 char * qtccallsign;
+int * qtccount;
 int qtccurrdirection;
 
 void fill_qtc_callsign(int direction, char * tcall) {
@@ -205,6 +206,7 @@ int qtc_main_panel(int direction) {
 	prevqtccall[strlen(qtcreclist.callsign)] = '\0';
 	qtccallsign = qtcreclist.callsign;
 	/* save the address of counter of receive QTC block */
+	qtccount = &qtcreclist.count;
 	/* set the QTC win header */
         sprintf(qtchead, "QTC receive");
     }
@@ -225,6 +227,7 @@ int qtc_main_panel(int direction) {
 	prevqtccall[strlen(qtclist.callsign)] = '\0';
 	qtccallsign = qtclist.callsign;
 	/* save the adress of counter of SEND qtc structure */
+	qtccount = &qtclist.count;
 	/* set the QTC win header */
 	sprintf(qtchead, "QTC send");
     }
@@ -350,7 +353,7 @@ int qtc_main_panel(int direction) {
 			showfield(activefield);
 			curpos = 0;
 		    }
-		    if (direction == SEND && ((trxmode != DIGIMODE && qtclist.totalsent < qtcreclist.count) || (trxmode == DIGIMODE && qtclist.totalsent == qtcreclist.count))) {
+		    if (direction == SEND && ((trxmode != DIGIMODE && qtclist.totalsent < *qtccount) || (trxmode == DIGIMODE && qtclist.totalsent == *qtccount))) {
 			switch (activefield) {
 			    case 0: activefield = 12;
 				    show_help_msg(activefield);
@@ -398,7 +401,7 @@ int qtc_main_panel(int direction) {
 			showfield(activefield);
 			curpos = 0;
 		    }
-		    if (direction == SEND && ((trxmode != DIGIMODE && qtclist.totalsent < qtcreclist.count) || (trxmode == DIGIMODE && qtclist.totalsent == qtcreclist.count))) {
+		    if (direction == SEND && ((trxmode != DIGIMODE && qtclist.totalsent < *qtccount) || (trxmode == DIGIMODE && qtclist.totalsent == *qtccount))) {
 			switch (activefield) {
 			    case 0: activefield = 2;
 				    showfield(0);
@@ -462,7 +465,7 @@ int qtc_main_panel(int direction) {
 				qtcreclist.qtclines[currqtc].receivedtime[15] = '\0';
 				qtcreclist.qtclines[currqtc].status = 2;
 				show_status(currqtc);
-				if (currqtc < qtcreclist.count) {
+				if (currqtc < *qtccount) {
 				    if (trxmode == DIGIMODE) {
 					qtcreclist.qtclines[currqtc].confirmed = 1; // compatibility for other modes
 					qtcreclist.confirmed++;
@@ -543,7 +546,7 @@ int qtc_main_panel(int direction) {
 			    mvwprintw(qtcwin, activefield, 30, "*");
 			    qtclist.qtclines[activefield-3].flag = 1;
 			    // scroll down if not at end of qtclist:
-			    if (activefield-3 < qtcreclist.count-1) {
+			    if (activefield-3 < (*qtccount-1)) {
 				wattrset(qtcwin, line_normal);
 				mvwprintw(qtcwin, activefield, 4, "%s", qtclist.qtclines[(activefield-3)].qtc);
 				activefield++;
@@ -606,13 +609,13 @@ int qtc_main_panel(int direction) {
 				tempc[0] = '\0';
 				strncpy(tempc, qtc_send_msgs[1], tlen-1);
 				tempc[tlen-1] = '\0';
-				sprintf(tmess, "%s %d/%d %s %d/%d\n", tempc, qtclist.serial, qtcreclist.count, tempc, qtclist.serial, qtcreclist.count);
+				sprintf(tmess, "%s %d/%d %s %d/%d\n", tempc, qtclist.serial, *qtccount, tempc, qtclist.serial, *qtccount);
 			    }
 			    timec[0] = '\0';
 			    get_time();
 			    strftime(timec, 40, "%d-%b-%y %H:%M", time_ptr);
 
-			    for(ql = 0; ql < qtcreclist.count; ql++) {
+			    for(ql = 0; ql < (*qtccount); ql++) {
 				qtclist.qtclines[ql].sent = 1;
 				strncpy(qtclist.qtclines[ql].senttime, timec, 15);
 				qtclist.qtclines[ql].senttime[15] = '\0';
@@ -645,7 +648,7 @@ int qtc_main_panel(int direction) {
 		    }
 		    break;
 	  case 19:	// CTRL-S - save QTC
-		    if (qtccurrdirection == SEND && qtcreclist.count > 0 && qtclist.totalsent == qtcreclist.count) {
+		    if (qtccurrdirection == SEND && *qtccount > 0 && qtclist.totalsent == *qtccount) {
 			if (log_sent_qtc_to_disk(nr_qsos) == 0) {
 			    wattrset(qtcwin, line_inverted);
 			    mvwprintw(qtcwin, 2, 11, "QTC's have been saved!");
@@ -719,7 +722,7 @@ int qtc_main_panel(int direction) {
 				tempc[0] = '\0';
 				strncpy(tempc, qtc_send_msgs[x - 129], tlen-1);
 				tempc[tlen-1] = '\0';
-				sprintf(tmess, "%s %d/%d ", tempc, qtclist.serial, qtcreclist.count);
+				sprintf(tmess, "%s %d/%d ", tempc, qtclist.serial, *qtccount);
 				sendmessage(tmess);
 			    }
 			    else if ((activefield-3) >= 0) {
@@ -917,7 +920,7 @@ int qtc_main_panel(int direction) {
 		    }
 		    break;
 	  case 48 ... 57:	// numbers
-		    /* if (trxmode == DIGIMODE && direction == RECV && activefield > 0 && activefield < 3 && qtcreclist.count == 0 && capturing == 0) {
+		    /* if (trxmode == DIGIMODE && direction == RECV && activefield > 0 && activefield < 3 && *qtccount == 0 && capturing == 0) {
 			capturing = 1;
 			wattrset(qtcwin, (chtype)(A_NORMAL | COLOR_PAIR(QTCRECVBG)));
 			mvwprintw(qtcwin, 1, 19, "CAPTURE ON ");
@@ -1009,7 +1012,7 @@ int showfield(int fidx) {
 	    posidx = 1;
 	}
 	else if (fidx == 2) {
-	    sprintf(fieldval, "%d", qtcreclist.count);
+	    sprintf(fieldval, "%d", *qtccount);
 	    winrow = 2;
 	    posidx = 2;
 	}
@@ -1086,7 +1089,7 @@ int modify_field(int pressed) {
 	    }
 	}
 	else if (activefield == 2 && isdigit(pressed)) {
-	    sprintf(fieldval, "%d", (qtcreclist.count)*10);
+	    sprintf(fieldval, "%d", (*qtccount)*10);
 	    fieldval[strlen(fieldval)-(1+curpos)] = pressed;
 	    if (curpos > 0) {
 		curpos--;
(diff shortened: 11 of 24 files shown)


rules/Makefile.am
```diff
@@ -1,6 +1,6 @@
 RULES_FILES = arrl10m_dx arrldx_dx arrldx_usa arrlfd arrlss contest cqww \
 	dxped eusprint focmarathon lzdx pacc_pa pacc_dx qso spdx_dx spdx_sp ssa_mt \
-	stewperry template wpx
+	stewperry template wpx waedc_dx waedc_eu aadx_dx aadx_as
 SUBDIRS =
```
dl1jbe: The waedx_[dx|eu]_rtty files are missing here.

src/qtcutil.c
```diff
@@ -0,0 +1,110 @@
...
+	 *
+	 *--------------------------------------------------------------*/
+
+#include "qtcutil.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <glib.h>
+#include "tlf.h"
+
+#define NBANDS 10
```
dl1jbe: NBANDS is already defined in tlf.h with a value of 9.

rules/Makefile.am
airween: Fixed.

src/rtty.c
```diff
@@ -202,6 +239,9 @@ int rx_rtty()
 	    ry_term[2][0] = '\0';
 	    ry_term[3][0] = '\0';
 	    ry_term[4][0] = '\0';
+	    /*for(i=0; i<13; i++) {
+		qtc_ry_lines[i][0] = '\0';
+	    }*/
```
dl1jbe: Do you need that piece of code? Please delete it or write a comment why it is here and when the comment should be dropped.

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+				activefield = 32;
+				showfield(0);
+			    }
+			    else {
+				activefield--;
+				showfield(activefield+1);
+			    }
+			}
+			else {
+			    if ((activefield < 3) || ((activefield-3)%3 > 0 && (activefield-3)%3 <= 2)) {
```
dl1jbe: x&3 will always be <= 2, so the last test is always true.

src/qtcutil.c
airween: Removed macro, and t_qtc_bands struct too.

src/checkqtclogfile.h
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include <stdio.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include "tlf.h"
+#include "startmsg.h"
```
dl1jbe: All the includes here should go to the *.c file. No other one including these file needs that knowledge.
Please add also the include guards.

src/rtty.c
airween: Removed those lines.

src/genqtclist.c
```diff
@@ -0,0 +1,100 @@
...
+	      i++;
+	  }
+	}
+	s++;
+    }
+
+    return qtclist.count;
+}
+
+int genqtcline(char * qtc, char * line) {
```
dl1jbe: Why do we return 0 here? It is the only possible return value and seems to have no meaning. So the function should return void.

src/qtcutil.c
```diff
@@ -0,0 +1,110 @@
...
+
+typedef unsigned char t_qtc_bands[NBANDS];
+
+extern GHashTable* qtc_store; // = NULL;
+extern int qtcdirection;
+extern struct t_qtc_store_obj *qtc_empty_obj;
+
+int qtc_inc(char callsign[15], int direction) {
```
dl1jbe: should return void

src/qtcutil.c
```diff
@@ -0,0 +1,110 @@
...
+	qtc_obj->received++;
+    }
+    if (direction == SEND) {
+	qtc_obj->sent++;
+    }
+
+    return 0;
+}
+
+int qtc_dec(char callsign[15], int direction) {
```
dl1jbe: should return void

src/qtcwin.c
airween: Fixed, replaced the complete clause.

src/genqtclist.c
airween: No, it could't be void. How did you test?
genqtclist() gets two parameters: callsign and number of qtc, which operator wants to send. If you don't have any "free" qso, which not sent in any previous QTC, then this function returns 0, what means you don't have any "free" QSO. May be you've tested it without previous QSO's.
Please keep this as is.

dl1jbe: Sorry, I am talking about the genqtcline() function.

src/qtcutil.c
airween: Changed.

src/qtcutil.c
airween: Changed.

src/readqtccalls.c
```diff
@@ -0,0 +1,134 @@
...
+    }
+    if (qtc_empty_obj != NULL) {
+	g_free(qtc_empty_obj);
+    }
+    qtc_store = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, NULL);
+    qtc_empty_obj = g_malloc0(sizeof (struct t_qtc_store_obj));
+    qtc_empty_obj->total = 0;
+    qtc_empty_obj->received = 0;
+    qtc_empty_obj->sent = 0;
+    nr_qsosflags_for_qtc = nr_qsos;
```
dl1jbe: I would suggest to put line 56..67 into a init_qtc_store() function or similar in qtcutil.c. Then we can move the qtc_store definition from main.c into qtcutil.c too. No one else needs to know about the hash table.
The same goes for qtc_empty_obj then.
What do you think?

src/genqtclist.c
airween: No, I'm sorry - I just see the previous "return qtclist.count", and associated to genqtclist(). Sorry again, changed return type of genqtcline(). Pushed to repo.

src/log_recv_qtc_to_disk.c
```diff
@@ -0,0 +1,191 @@
...
+    qtcreclist.count = 0;
+    qtcreclist.serial = 0;
+    qtcreclist.confirmed = 0;
+    qtcreclist.sentcfmall = 0;
+    qtcreclist.callsign[0] = '\0';
+
+    return (0);
+}
+
+int store_recv_qtc(char *loglineptr)
```
dl1jbe: Does it need to return a int? void should be enough. No one tests this return value.

src/log_recv_qtc_to_disk.c
```diff
@@ -0,0 +1,191 @@
...
+    for(i=0; i<10; i++) {
+	qtcreclist.qtclines[i].time[0] = '\0';
+	qtcreclist.qtclines[i].callsign[0] = '\0';
+	qtcreclist.qtclines[i].serial[0] = '\0';
+	qtcreclist.qtclines[i].status = 0;
+	qtcreclist.qtclines[i].confirmed = 0;
+	qtcreclist.qtclines[i].receivedtime[0] = '\0';
+    }
+    for(i=0; i<13; i++) {
+	qtc_ry_lines[i].content[0] = '\0';
```
dl1jbe: that maybe out of bounds. There are only 12 members of the qtc_ry_lines array defined but 13 (0..12) gets initialized.

src/log_recv_qtc_to_disk.c
airween: Changed and pushed.

airween: In case of store_sent_qtc() function I've changed too.

src/deleteqso.c
```diff
@@ -47,8 +57,95 @@ void delete_qso(void)
...
+                            while (look == 1) {
+                                lseek(qtcfile, ((int)qstatbuf.st_size - (96+qtclen)), SEEK_SET);
+                                rc = read(qtcfile, logline, 95);
+                                if (! (strncmp(call, logline+QTCSENTCALLPOS, strlen(call)) == 0 && strncmp(bandmode, logline, 5) == 0)) {
+                                    look = 0;
+                                }
+                                else {
+                                    qtclen += 96;
+				    qtc_dec(call, SEND);
+				    qsoflags_for_qtc[s] = 0;
```
dl1jbe: PCLint warns that the index s maybe negative. I think that comes from the test 'while(s >= 0 &&...) s--;' in line 117. Maybe it should be s>0. Please recheck.

src/log_recv_qtc_to_disk.c
airween: Changed to QTC_RY_LINE_NR macro. Pushed.

dl1jbe:
Thanks for the quick run this evening. I have to stop for now. More to come.

src/deleteqso.c
airween: As I know, the (most? all?) C compiler evaluate the && clause from left to right. If the first of left clause is false, it doesn't continue the evaulation, exists as false - but I will check it on a test code. If this theory is true, then 's' couldn't be negative. (If I would shift the two clauses, and the 's > 0' should be at right, then the index could be negative at first clause, which is wrong.) What do you think about it?

From another point: the 's > 0' wouldn't enough, cause the first line will not be checked ever.

dl1jbe: Yes it is checked from left to right.
 But 's>=0'  is even true if s == 0 and then the 's--' is executed a last time resulting in s = -1. 
Or am I wrong?

airween: My opinion: if the s == 0, then the right clause will evaulated too. If that's true, then the full clause will be true, and "s--" will be performed. Now 's' will be -1, and the next (left) 's >= 0' comparison will be false, so the right clause will not evaulated.

dl1jbe: Ah, I see were me misunderstand us. You are right, the right side will not be evaulated in these case. But the problem is in line 132. were s may still be -1.

airween: I've fixed the problem - in the inner loop, there is a new clause. See the diff, in line 123.

src/checkqtclogfile.h
airween: Pushed all request.

src/background_process.c
```diff
@@ -164,6 +164,16 @@ void *background_process(void *ptr)
...
+		case QTCRENTRY:
+
+		    store_recv_qtc(lan_message+2);
+		    break;
+
+		case QTCSENTRY:
+
+		    store_sent_qtc(lan_message+2);
+		    break;
+
```
dl1jbe: The file misses a #include for the new functions.

src/callinput.c
```diff
@@ -210,6 +213,26 @@ char callinput(void)
...
 	    }
 
+	case 17:	// CTRL+q
+	    {
+		if (qtcdirection & 1) {
+		    qtc_main_panel(RECV);
+		}
+		if (qtcdirection == 2) {
+		    qtc_main_panel(SEND);
+		}
```
dl1jbe: What happens if qtcdirection is set to 3 (BOTH)?
Do we show the qtc panel twice after each other?

src/changepars.c
```diff
@@ -582,14 +584,22 @@ int changepars(void)
 		synclog(synclogfile);
 	    scroll_log();
 	    /** \todo register return value */
+	    total = 0;
```
dl1jbe: Setting 'total' to 0 seems unneeded, as it is done already in readcalls().

src/genqtclist.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+#ifndef GENQTCLIST_H
+#define GENQTCLIST_H
+
+int genqtclist(char * callsign, int nrofqtc);
+void genqtcline(char * qtc, char * line);
```
dl1jbe: That function is only used in genqtclist() and not needed on other side. The prototype should be moved to the *.c file.

src/genqtclist.c
```diff
@@ -0,0 +1,99 @@
...
+	qtclist.qtclines[s].flag = 0;
+	qtclist.qtclines[s].saved = 0;
+	qtclist.qtclines[s].sent = 0;
+	qtclist.qtclines[s].senttime[0] = '\0';
+    }
+
+    s=next_qtc_qso;
+
+    while (qtclist.count < qtclistlen && s < nr_qsos) {
+        if (strlen(callsign) == 0 || strncmp(qsos[s]+29, callsign, strlen(callsign)) != 0) {	// exclude current callsign
```
dl1jbe: Why do you ran that loop at all if there is no call sign to look for?
You can test that before the loop.

dl1jbe: Sorry forget the above comment. I missread the code.

src/getexchange.c
```diff
@@ -161,9 +164,27 @@ int getexchange(void)
...
 	switch (x) {
 
+	case 17:	// CTRL-Q
+	    {
+		if (qtcdirection & 1) {
+		    qtc_main_panel(RECV);
+		}
+		if (qtcdirection == 2) {
+		    qtc_main_panel(SEND);
+		}
```
dl1jbe: Same problem as in callinput.c - what happens if qtcdirecttion is 3?

src/main.c
```diff
@@ -62,6 +63,7 @@ int debugflag = 0;
 int editor = EDITOR_JOE;
 char rttyoutput[120];
 int use_vk = 0;
+int use_qtc_vk = 0;
```
dl1jbe: Does that variable make sense? It is needed no where (just as the variable above).

src/main.c
```diff
@@ -315,6 +332,21 @@ char simulator_tone[5];
...
+t_qtcreclist qtcreclist;
+struct t_qtc_store_obj *qtc_temp_obj;
+int qtcdirection = 0;
+//char qtc_ry_lines[12][50] = {"", "", "", "", "", "", "", "", "", "", "", ""};
+//int qtc_ry_lines_attr[12] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+t_qtc_ry_line qtc_ry_lines[QTC_RY_LINE_NR];
+int qtc_ry_currline = 0;
+int qtc_ry_capture;
+int qtc_ry_copied;
+
```
dl1jbe: As global variables makes the code much harder to maintain, please check if the new qtc related definitions (line 335..348) are really need in a global context. Maybe they can be moved to the according *.[ch] files or make local variables.

An example for the last one seems to be 'qtc_temp_store'.

src/main.c
```diff
@@ -804,6 +854,7 @@ int main(int argc, char *argv[])
     showmsg(tlfversion);
     showmsg("");
 
+    total = 0;
```
dl1jbe: Already done in readcalls(). But maybe it should be made explicit.

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+			}
+		    }
+		    break;
+	  case 10:  		// ENTER
+		    if (activefield > 2) {
+			if (direction == RECV) {
+			    currqtc = ((activefield-3)/3);
+			    //if ((activefield-3)%3 == 2) {
+			    // TODO - is it need to restrict only for last field to allow the ENTER?
+			    if (1) {
```
dl1jbe: That check is always true. Should be cleaned up.

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+			    refreshp();
+			    sleep(1);
+			    x = 27;	// close the window
+			}
+			//data_ready = 1;
+		    }
+		    if (*qtccurrdiretion == RECV && trxmode == DIGIMODE) {
+			qtc_ry_capture = 1;
+			wattr_get(qtcwin, &attributes, &cpair, NULL);
+			wattrset(qtcwin, line_inverted);
```
dl1jbe: You use a weird mix of wattr_get and wattrset functions. Why no wattr_set?

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+		showfield(activefield);
+	    }
+
+	}
+	return 0;
+}
+
+int shift_right(char * fieldval) {
+	int i;
+	for(i=strlen(fieldval); i<strlen(fieldval)-curpos; i--) {
```
dl1jbe: As curpos is alway >=0 that logic makes no sense. In start of loop i will never be <strlen(fieldval) - curpos.

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+				mvwprintw(qtcwin, activefield, 4, "%s", qtclist.qtclines[(activefield-3)].qtc);
+				activefield++;
+				wattrset(qtcwin, line_currinverted);
+				mvwprintw(qtcwin, activefield, 4, "%s", qtclist.qtclines[(activefield-3)].qtc);
+			    }
+			    if (*qtccount > 0 && qtclist.totalsent == *qtccount) {
+				wattrset(qtcwin, line_inverted);
+				mvwprintw(qtcwin, 2, 11, "CTRL+S to SAVE!");
+				refreshp();
+				show_help_msg(6);
```
dl1jbe: index 6 is not in the field of help messages. Out of bounds....

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+	    wtoken = strtok_r(NULL, " ", &saveptr1);
+	}
+    }
+    else {
+	strcpy(sep, "/;-: ");
+	token = strtok(lline, sep);
+	while(token != NULL && t < 5) 
+	{
+	    t++;
+	    strcpy(tokens[t], token);
```
dl1jbe: out-of-bounds as t maybe 5 but the field has only indices 0..4

src/qtcwin.c
```diff
@@ -0,0 +1,1816 @@
...
+	while (x < 1) {
+
+	    usleep(1000);
+	    time_update();
+	    if (miniterm == 1 && trxmode == DIGIMODE && (keyerport == GMFSK
+	           || keyerport == MFJ1278_KEYER)) {
+		    show_rtty();
+	    }
+	    if ((qtc_ry_currline < 11 && prevline != qtc_ry_currline-1) || qtc_ry_currline == 11) {
+		if (actline > 1 && strcmp(firstline, qtc_ry_lines[0].content) != 0) {
```
dl1jbe: 'firstline' may not be initialised

src/parse_logcfg.c
```diff
@@ -422,6 +456,69 @@ int parse_logcfg(char *inputbuffer)
...
 	"POWERMULT",		/* 160 */
 	"SERIAL_OR_SECTION",
+	"QTC",
+	"CONTINENTLIST",
+	"CONTINENT_LIST_POINTS",
+	"USE_COUNTINENTLIST_ONLY",  /* 165 */
+	"BANDWEIGHT_POINTS",
+	"BANDWEIGHT_MULTIS",
+	"PFX_NUM_MULTIS",
+	"PFX_MULT_ALLB",
```
dl1jbe: PFX_MULT_ALLB needs to be documented in the man page.

src/globalvars.h
```diff
@@ -123,3 +126,20 @@ extern int dxped;
...
+extern t_qtclist qtclist;		// the QTC list to send
+extern int nr_qtcsent;
+extern t_qtcreclist qtcreclist;		// the QTC list which received
+
+extern int qtcdirection;
+extern t_pfxnummulti pfxnummulti[MAXPFXNUMMULT];
+extern int pfxnummultinr;
+extern int continentlist_only;
+//extern char qtc_ry_lines[12][50];
+extern t_qtc_ry_line qtc_ry_lines[QTC_RY_LINE_NR];
```
dl1jbe: Please add a comment what the qtc_ry_lines is for

src/main.c
```diff
@@ -315,6 +332,21 @@ char simulator_tone[5];
...
 
+int qsoflags_for_qtc[MAX_QSOS];
+int nr_qsosflags_for_qtc;
+int next_qtc_qso;
+t_qtclist qtclist;
+int nr_qtcsent = 0;
+t_qtcreclist qtcreclist;
+struct t_qtc_store_obj *qtc_temp_obj;
+int qtcdirection = 0;
+//char qtc_ry_lines[12][50] = {"", "", "", "", "", "", "", "", "", "", "", ""};
```
dl1jbe: What is about these commented out functions? Do we need them?

src/readqtccalls.c
airween: As I see, the qtc_init() had been implemented by you, so this issue could be resolved.

src/background_process.c
airween: I don't see the context between your comment, and the modified code. The diff shows only the switch statement cases, but you wrote the "file misses". Should we need to clear it, or just simple regarded as solved?

src/callinput.c
airween: No. If qtcdirection is 3, that means the (qtcdirection & 1) will be evaulated as TRUE, but (because qtcdirection = 3), the (qtcdirection == 2) will be false. That's works as well in the code.
When I tested the last code with the waedc-based config, I've found some small disfunction, but I think that's not affects this.

src/changepars.c
airween: Looks like this one also fixed in the new version.

src/genqtclist.h
airween: Fixed in the latest version.

src/genqtclist.c
airween: Ok, I just marked this conversation as "seen".

src/getexchange.c
airween: Same effect as callinput. If qtcdirection = 3, then (qtcdirection == 2) will be FALSE.

src/main.c
airween: Removed, pushed.

src/main.c
airween: Explicit is better than implicit :). In readcalls(), the total initialization is only after a condition, but the readqtccalls() also initializes the total variable. I think, now it's better to keep this.

src/qtcwin.c
airween: Yes, that's a question, but I don't know the answer. I don't know, what could be the good solution? There are 3 fields in a QTC line: time, callsign and serial. If the operator press the ENTER, the current QTC will saved. The question is what's the better solution? Only allow the ENTER if the current field is serial, or in any field? Now the last one is the used, but I don't know, should we change this?

src/qtcwin.c
airween: No, I didn't mixed. There are never used wattr_set, only wattrset (from the ncurses library). There are wattrset() and wattr_set() functions:
int wattr_set(WINDOW *win, attr_t attrs, short pair, void *opts);
int wattrset(WINDOW *win, int attrs);

Othere developers always used wattrset(), I didn't want to deviate.

But there is only wattr_get() function, no wattrget().

What could be the right way?

src/qtcwin.c
airween: Sorry, I don't understand this issue.

src/qtcwin.c
airween: The length on help_recv_msgs[] is 7, so it contains 6th element. If the direction is SEND, then the condition prevents the out of bound error.
  1381      if (*qtccurrdiretion == SEND) {
  1382          if (msgidx > 2 && msgidx < 6) {
  1383              msgidx = 3;
  1384          }
  1385          mvwprintw(qtcwin, ++j, 36, help_send_msgs[msgidx]);
  1386      }

I think there isn't any problem.

src/qtcwin.c
airween: Tokens are generated by strtok(). If the strtok() ends the loop, then the token will be NULL, and "t < 5" will not be evaulate.
This "t < 5" os only if the RTTY line contains some extra separator character. The lenght of tokens is 5:
1422.    ..., tokens[5][15] = {"", "", "", "", ""};

src/qtcwin.c
airween: Fixed, pushed.

src/parse_logcfg.c
airween: Added and pushed.

src/globalvars.h
airween: Added, pushed.

src/main.c
airween: No, we don't need them. Removed and pushed.

src/background_process.c
dl1jbe: You use some functions here and let the compilerr guess the prototype. That will sometimes go wrong.
So background_process.c needs to #iinclude "store_recv_qtc.h" and #include "store_sent_qtc.h" file to know the prototypes of the two functions here.

src/callinput.c
dl1jbe: Ah, I see. It looks a little weird but should work. It would be good to add a comment about it or use an else between the two cases. THat way the next one will not stumble upon it.

src/main.c
dl1jbe: I think we should change the readcall() funtion to made it more explicit. I will look for that myself.

src/qtcwin.c
dl1jbe: The question is what the 'if (1)' is for? There is nothing to check. So why is the test here?

src/qtcwin.c
dl1jbe: As the wattr_get does return 'attr_t'  and wattrset expects an 'int' there may be some mismatch. So it seems normal to use wattr_set for it. I am not quite sure if anyone else has used wattr_get before.

src/qtcwin.c
dl1jbe: Please check the logic in the for loop. 
You initialise 'i = strlen(fieldval)' and will execute the loop as long as 'i<strlen(fieldval)-curpos'. But as far as I know the for loop will check the test before entering the loop and the test will not be true as long as 'curpos' is >= 0. So the loop will not be executed.

src/qtcwin.c
dl1jbe: See line 99 ot the same file:
char help_rec_msgs[6][26] = {
    "Enter callsign",
    "Enter the QTC serial",
    "Enter the QTC number",
    "Enter the time",
    "Enter the CALL",
    "Enter the SERIAL"
};
The length of the help_rec_msgs is 6.

src/qtcwin.c
dl1jbe: Right. But if t < 5 you enter the while clause and increment t. Say t is 4 now you access tokens[5].

src/qtcwin.c
airween: Oh, sorry - the length of help_rec_msgs[] is 6, that's correct. But the help_send_msgs[] is 7, and there is a condition before that snippet:
```C
510.    if (direction == SEND && trxmode != DIGIMODE) {
...
537.        if (*qtccount > 0 && qtclist.totalsent == *qtccount) {
...
541.            show_help_msg(6)
```
So, I think there couldn't be overflowed. Anyway, this part of code executed when SENDER station sent all line of QTC in DIGIMODE - shows the message: "Press CTRL+S to save"

src/qtcwin.c
airween: Ah, well, that's true. Fixed and pushed.

src/callinput.c
airween: Right, I've modified the code both source file, where these events handled. Fixed and pushed.

src/qtcwin.c
airween: The fake test here is just to remember me/us to discuss, what could be final solution :). After two 3 contest I prefer here isn't need any conditions. Any field could be current in the QTC line, if operator pressed the ENTER (with empty fields), then it means he/she could't get the all QTC. What do you think about?

src/qtcwin.c
airween: May be nobody needs that. I could't solve another way to store the problem. Should I do anything at this issue?

src/qtcwin.c
airween: Ah, well, now I see - so, I've made 3 contest with this way. Now I removed the loop, but the last "shift" line stayed. Tested and pushed.

src/background_process.c
airween: Hmm, that's very interesting. Of course, I've added header files to background_process.c.
But now I'm a littlebit confused: the source tree compiled succesfully, the compiler not misses this header (and prototypes). I think, because it found the Makefile. I've tested the network mode too, so there are two clients, and they communicates with each other without problem. And as I see, there is an another function in background_process.c, which not has prototype in the C source: qsonr_to_str(). I have to read about it... :)

airween: Fixed, tested and pushed.

dl1jbe: Any time you call a function which is not previously defined (by function declaration or prototype) C guesses a funcction prototype by itself: int xyz (). Later on it may see that the function inn reality has a different prototypeand will complain.

But if the funcion is defined in a different *.c file the compiler has no chance to check for that and therefore has no chance to check it out.

In our case it was not critical because the compiler thoughed the function will return an int which we ignored. But the function did return void. But there may be subtle bugs - I am bitten more than once by it.

src/qtcwin.c
```diff
@@ -0,0 +1,1814 @@
...
+			    break;
+		}
+		showfield(activefield);
+	    }
+
+	}
+	return 0;
+}
+
+int shift_right(char * fieldval) {
```
dl1jbe: Does the 'shift_right' funtion now make sense at all?
All it is doing is to add a second \0 after the one which terminates the string already.

src/qtcwin.c
```diff
@@ -0,0 +1,1814 @@
...
+WINDOW * qtcwin;
+PANEL * qtc_panel;
+WINDOW * ry_win;
+PANEL * ry_panel;
+WINDOW * ry_help_win;
+PANEL * ry_help_panel;
+int activefield;
+// array values: hor position, cursor position, field len to fill with spaces
+int pos[6][3] = {{3, 3, 15}, {3, 6, 4}, {8, 8, 2}, {3, 3, 4}, {8, 8, 14}, {24, 24, 4}};
+int curpos = 0;
```
dl1jbe: I think 'curpos' is a little bit of a misnomer. I would not expect a position to be counted backwards from the end of an string. A better name would be good.

src/qtcwin.c
```diff
@@ -0,0 +1,1814 @@
...
+    "",
+    "",
+    "Press CTRL+S to SAVE!"
+};
+
+char * qtccallsign;
+int * qtccount;
+int * qtccurrdiretion;
+
+int qtc_main_panel(int direction) {
```
dl1jbe: That is a very big function. It will make it hard to maintain it in the future. I suggest to split it into parts, e.g. for creating the panel, running the loop and for destroying the panel. What do you think

src/qtcwin.c
```diff
@@ -0,0 +1,1814 @@
...
+    if (*qtccurrdiretion == RECV || *qtccurrdiretion == SEND) {
+	mvwprintw(qtcwin, 1, 19, "%s %2d QTC", qtcdirstring[*qtccurrdiretion], qtc_temp_obj->total);
+    }
+    else {
+	mvwprintw(qtcwin, 1, 19, "Total: %2d QTC (R: %d, S: %d)", qtc_temp_obj->total, qtc_temp_obj->received, qtc_temp_obj->sent);
+    }
+    return 0;
+}
+
+int strip_spaces(char * src, char * tempc) {
```
dl1jbe: The function name is wrong. It does not 'strip spaces' but reduce duplicate spaces to one and repce them by '-' if in DIGIMODE.

Btw, an easier way is to use g_strsplit() and then g_strjoin() wih a given separator.

src/background_process.c
airween: Ok, many thanks for the full description :)

src/qtcwin.c
airween: Adding a second \0 to terminate, but then it returns, the next code could insert a charater to the current position. What do ou propose to handle this inssert function?
It could be better to make a function with arguments field and character, eg "insert_char(field, c)"? What do you think?

src/qtcwin.c
airween: 'Curpos' stores the current position, but not (just) from the and to the begin, in backwards direction. Sometimes it counts as backward, somtimes as forward. Here is a snippet for all cases:
```C
	  case 155:	// left
		    if (DIRCLAUSE) {
			if (curpos < curfieldlen) {	// curpos is a shift, means lenght - position
			    curpos++;
			    showfield(activefield);
			}
		    }
		    break;
	  case 154:	// right
		    if (DIRCLAUSE) {
			if (curpos > 0) {
			    curpos--;
			    showfield(activefield);
			}
		    }
		    break;
```

src/qtcwin.c
airween: I've changed to replace_spaces(). Now I keept the own function instead of the g_str*() functions. Pushed.

src/qtcwin.c
airween: Well, you're absolutely right. I'll reduce the core functions to a new function - but I need a little bit more time to made this. Keep this issue open please.

src/score.c
```diff
@@ -336,6 +356,24 @@ int score()
 	    points = 0;
     }
 
+    /* HA2OS mods */
+    // only continent list allowed
+    if (continentlist_only == 1) {
+	is_cont_mult = continent_found();
+	if (is_cont_mult == 1) {
+	    // if we are on DX continent
+	    if (continent == mycontinent) {
```
dl1jbe: That is no string comparison above. It only tests if the addresses (not the content) of  both strings are the same.

src/parse_logcfg.c
```diff
@@ -1377,12 +1475,300 @@ int parse_logcfg(char *inputbuffer)
...
+		qtcrec_record_command[i][q++] = ' ';
+		qtcrec_record_command[i][q++] = '&';
+	    }
+	    qtcrec_record_command[i][q] = '\0';
+ 	    break;
+    }
+    case 224: {
+	    PARAMETER_NEEDED(teststring);
+	    if (strcmp(fields[1], "CONTINENTLIST")) {
+	        if (continent_multiplier_list == NULL) {
```
dl1jbe: Bad string check

src/parse_logcfg.c
```diff
@@ -1377,12 +1475,300 @@ int parse_logcfg(char *inputbuffer)
...
+		    showmsg
+			("WARNING: you need to set the CONTINENTLIST paramter...");
+		    sleep(5);
+		    exit(1);
+		}
+		exclude_multilist_type = 1;
+	    }
+	    else if (strcmp(fields[1], "COUNTRYLIST")) {
+	        if (mit_multiplier_list == NULL) {
```
dl1jbe: Bad string check - will not work.

src/score.c
airween: Well, I'm really sorry for that really amateur error :). Fixed and pushed.

src/parse_logcfg.c
airween: Fixed, pushed.

src/parse_logcfg.c
airween: Fixed, pushed with prevoius fix.

src/main.c
airween: Done, see the commit: https://github.com/airween/tlf/commit/9d32cee8fcc0452d0237ea6370c8b7cc5f5aad9e

src/qtcwin.c
dl1jbe: Yes. ENTER shoudl work in any field, but only if all three fields contains data. 
So you have to fill in some data, but you can go back edit any value and just press ENTER there without the need to go back to the last field.

src/qtcwin.c
dl1jbe: A 'edit_field' function seems to make sense. But if the actual implementation works - that means the function does not do anything at the moment and can be dropped completely. But that has to bet tested. Please check it.

src/qtcwin.c
dl1jbe: Ok, I will look into it later when I pull it over to master branch.

src/qtcwin.c
dl1jbe: THe name 'position' means to me: "Look there it is. It is the 3rd character in the field". But as you see yourself it is the distance to the end of the string. Yes from that distance you can calculate the position but it is something different. 

A better name with a good meaning seems not easy to find, maybe 'tillEnd' 

But maybe a better concept would be not to rename the variable but to simply count the cursor position from the beginngin of the field. I do not know yet.

src/qtcwin.c
airween: Thanks for the help. I've removed the fake condition, and some forgotten "TODO" lines, which had been fixed early. Now the ENTER works in all fields, but only goes to the next line, if all fields contains the right data. Else, Tlf sends an "AGN" message, or what had been configured to F8 key. Checked and pushed.

src/qtcwin.c
airween: Ok, I've checked out it, and removed all occures of shift_right() function.

dl1jbe:
All changes pulled in via cleaned up branch dl1jbe/waedc-qtc

PR #33, opened 2015-02-22
Updated manual page

N0NB:
I've been studying the recommended format for man pages from the Linux man pages projects.  Hopefully, I have not made too much of a mess!  It looks good to me in the terminal and in a generated PS file from a2ps.

I've tried to incorporate the recommended conventions for writing man pages and I've made some editorial changes to some of the content to (hopefully) clarify some of the language.

73, Nate


dl1jbe:
Thanks Nate for the contribution. I am sure it was a lot of work.
I had a first look - all seems to be well.

I cherry picked the commit as I didi not know where the first of the commits wrt .gitignore changes are coming from. I have no 'compile' file here a t my side. 

73, de Tom

N0NB:
* On 2015 23 Feb 12:39 -0600, Thomas Beierlein wrote:
> Thanks Nate for the contribution. I am sure it was a lot of work.
> I had a first look - all seems to be well.
> 
> I cherry picked the commit as I didi not know where the first of the
> commits wrt .gitignore changes are coming from. I have no 'compile'
> file here a t my side.

It probably comes from me using the Autotools in Debian Sid and then
running 'autoreconf -i' which installs files rather than symbolic
links.  As I sometimes leave a project configured for a while I found
that strange things would happen when some symlinked file was updated
and the generated files already present didn't get updated by a new run
of autoreconf.

Including the .gitignore would be harmless.  Your choice, of course.
:-)

73, Nate

-- 

dl1jbe:
Ah, thanks for clarification.
And sure, you are right. It is harmless. I am just curious :-).

N0NB:
* On 2015 23 Feb 13:28 -0600, Thomas Beierlein wrote:
> Ah, thanks for clarification.
> And sure, you are right. It is harmless. I am just curious :-).

This is why with Hamlib I chose to remove all generated files from Git
control.  I was poking at cwdaemon and even though he has all the
generated files (configure, etc) in the repository, it didn't work so I
had to run autoreconf anyway.  It seems like it should have worked, but
I did not pursue it since I'm familiar with Autotools.

On a related note, I spent the morning rounding up what source I could
find for Rein's winkeydaemon that was in Debian 5.0 and then removed
after that.  As I have a Hamgadgets Master Keyer 1 which does K1EL
emulation, I'm looking at it as a means to use Tlf with it.  As I
couldn't find anything else on it besides an update from Wilbert PE7T to
make it work with his Winkeyer USB Lite, I presume no one is maintaining
it so I've been putting it into Git and studying it.  Also, getting
reacquainted with Perl.  Heh!

73, Nate

-- 

PR #34, opened 2015-05-31
update "repeat number" string

dh5fs:
its better to send "5NN" as prefix before repeating the number than NR, some people react better to that under low S/N conds

### DIFF
--- a/src/prevqso.c
+++ b/src/prevqso.c
@@ -54,7 +54,7 @@ int prev_qso(void)
 	nr_buffer[1] = '\0';
     }
 
-    str = g_strdup_printf("NR %s ", nr_buffer);
+    str = g_strdup_printf("5NN %s ", nr_buffer);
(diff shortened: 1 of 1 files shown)


dl1jbe:
Good Idea Fred. I have just one small comment: I know most of us
use simple 5nn for rst but as tlf allows you to change tlf-sent for
the qso it seems better to be consistent and repeat what you gave in
the first report.

dh5fs:
Thats true. I use mostly shorted numbers.

dl1jbe:
Ok. So it should use the actual rst and also honor the SHORT_SERIAL setting. 
I added it to the todo list.

dl1jbe:
Sorry for being so late. Lot of QRL here. Please test and report back if needed.

dh5fs:
Works!! TU!

PR #44, opened 2015-10-26
Fixed break off auto_cq bug

airween:
When Tlf sends CQ in auto mode, and op press the first letter of callsign, that will be lost, only the second letter will visible and will stored.


dl1jbe:
Good catch. Thanks

PR #46, opened 2015-11-04
New features: BMAUTOADD and BMAUTOGRAB keywords

airween:
New features: with BMAUTOADD and BMAUTOGRAB keywords, Tlf will add to or grab from bandmap the matching callsigns


src/bandmap.c
```diff
@@ -413,9 +418,11 @@ void bandmap_age() {
 	spot *data = list->data;
 	GList *temp = list;
 	list = list->next;
-	if (data->timeout) {
-	    data->timeout--;
-	    if (data->timeout == 0) {
+	if (data->timeout >= 0) {
+	    if (bmautograb == 0 || (bmautograb > 0 && data->timeout > 0)) {
```
dl1jbe: The whole logic here seems weird. If bmautograb is 0 then the timeout will be decremented even if it ist already 0. That makes no sense.

src/bandmap.c
```diff
@@ -802,14 +812,39 @@ void bandmap_show() {
...
 	if (!on_qrg) {
 	    printw ("%7.1f   %s", centerfrequency,  "============");
+	    if (bmautograb != 0 && autograbbed == 1 && cqmode == S_P) {
+		hiscall[0] = '\0';
+		showinfo( getctydata( hiscall ) );
+		searchlog( hiscall );
+		autograbbed = 0;
+	    }
+	    // clear hiscall if no qrg
+	    if (bmautoadd == 1 && bmadded_spot == 1 && strlen(hiscall) > 0) {
```
dl1jbe: I assume that the code below should handle the case that we tuned frequency after the auto_add. Then the added call will be cleared?

I think it is not the right place to do that.  From a logical point of view it should be done in that section which added the call to the bandmap - that would be callinput. What do you think?

dl1jbe:
Any news on this?

airween:
Currently no - I've worked on this:
https://github.com/Tlf/tlf/pull/54
But now I can start to work with this request (46).

PR #47, opened 2015-11-11
Bandmapfixup

airween:
Fix some small bugs in auto_cq mode:
* bandmap didn't refreshed, when Tlf worked in auto_cq mode
* cluster info didn't writed to dbfile (.bmdata.dat file) in every 10secs

### DIFF
--- a/src/autocq.c
+++ b/src/autocq.c
@@ -30,6 +30,7 @@
 #include "stoptx.h"
 #include "cw_utils.h"
 #include "ui_utils.h"
+#include "bandmap.h"
 #include "clusterinfo.h"
 #include <sys/time.h>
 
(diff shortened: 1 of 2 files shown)


src/autocq.c
```diff
@@ -77,8 +84,15 @@ int auto_cq(void)
 		    stoptx();
 		    break;
 		}
+		gettimeofday(&timecnt, NULL);
```
dl1jbe: While correct, I think that is too complex. Just calling time_update() here does the whole job (same in line 105). It has the additional value that other time related work gets also done during auto_cq.

src/bandmap.c
```diff
@@ -826,6 +817,16 @@ void bandmap_show() {
     move(cury, curx);			/* reset cursor */
 
     refreshp();
+
+    gettimeofday(&tv, NULL);
```
dl1jbe: Maybe the bmdata_write_file should also be called from time_update(). Then we have concentrated all time dependent actions in one place. What do you think?

dl1jbe:
Hi Ervin, in meantime I found time to fix the two problems in the master branch:

- aging and display during auto CQ and
- regular writing of the bandmap data 

Please check that it is working as expected. If so you can close the pull request. Thanks again for reporting the problem.

airween:
Hi Thomas,

I've checked these two modifications, and looks like everything is good - thanks for the help.
I'll close this pull request.

Issue #51, opened 2015-12-06
panel.h not found on some distributions.


N0NB:
Per this thread:

http://lists.nongnu.org/archive/html/tlf-devel/2015-12/msg00009.html

David reported an issue with his distribution not havng panel.h in /usr/include.  At this time configure.ac does not have an explicit test for the location of panel.h.  Suspect this can be resolved by use of extra macros available from the Autoconf macro archive:

http://www.gnu.org/software/autoconf-archive/

Tom, you may assign me to this, if you can.

dl1jbe:
Seems I can only assign organisation members. Lets see what can be done here.

dl1jbe:
So, now I could assign you.

N0NB:
I have submitted pull request #62 for this issue.

dl1jbe:
Closing the issue as pull request #56 solved the problem. Thanks Nate.

PR #52, opened 2015-12-08
First crack at finding ncurses header files in ncurses directory

N0NB:
Some systems, OpenSuSE LEAP 42.1, for example, put the ncurses header files under /usr/include/ncurses.  This patch set modifies the build system to look for the ncurses headers in either /usr/include or /usr/include/ncurses and when found sets constants in config.h.  Then use CPP tests to load the found file.

This references issue #51

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -20,9 +20,18 @@ PKG_CHECK_MODULES(GLIB, glib-2.0)
 AC_CHECK_LIB([m], [atan])
 AC_CHECK_LIB([pthread], [pthread_create])
 ACX_PTHREAD
-AC_CHECK_LIB([ncurses], [initscr],,AC_MSG_ERROR([needs ncurses library]))
-AC_CHECK_LIB([panel], [update_panels],,
-		AC_MSG_ERROR([needs ncurses panel library]))
+
+
+dnl Order matters as -lpanel should be before -lncurses and
+dnl AC_SEARCH_LIBS prepends found libraries to @LIBS@ so search
+dnl for the panel library after ncurses to get the correct order.
+AC_SEARCH_LIBS([initscr], [ncurses],,[AC_MSG_ERROR([needs ncurses library])])
+AC_SEARCH_LIBS([update_panels], [panel],,
+	[AC_MSG_ERROR([needs ncurses panel library])
+])
+
+dnl Ncurses libraries have been found, now look for the headers.
+AC_CHECK_HEADERS([ncurses.h curses.h ncurses/ncurses.h ncurses/curses.h panel.h ncurses/panel.h])
 
 # Checks for header files.
 AC_HEADER_STDC
--- a/src/addmult.c
+++ b/src/addmult.c
@@ -29,14 +29,22 @@
 #include <string.h>
 #include <unistd.h>
 
-#include <curses.h>
-
-#include "addmult.h"
-#include "globalvars.h"		// Includes glib.h and tlf.h
-
 #ifdef HAVE_CONFIG_H
 # include <config.h>
 #endif
+
+#ifdef HAVE_NCURSES_NCURSES_H
+# include <ncurses/ncurses.h>
+#elif defined HAVE_NCURSES_CURSES_H
+# include <ncurses/curses.h>
+#elif defined HAVE_NCURSES_H
+# include <ncurses.h>
+#elif defined HAVE_CURSES_H
+# include <curses.h>
+#endif
+
+#include "addmult.h"
+#include "globalvars.h"		// Includes glib.h and tlf.h
 
 #define MULTS_POSSIBLE(n) ((char *)g_ptr_array_index(mults_possible, n))
 
--- a/src/addspot.c
+++ b/src/addspot.c
@@ -31,7 +31,19 @@
 #include <string.h>
 #include <time.h>
 
-#include <curses.h>
+#ifdef HAVE_CONFIG_H
+# include <config.h>
+#endif
+
+#ifdef HAVE_NCURSES_NCURSES_H
+# include <ncurses/ncurses.h>
+#elif defined HAVE_NCURSES_CURSES_H
+# include <ncurses/curses.h>
+#elif defined HAVE_NCURSES_H
+# include <ncurses.h>
+#elif defined HAVE_CURSES_H
+# include <curses.h>
+#endif
 
 #include "get_time.h"
 #include "lancode.h"
--- a/src/audio.c
+++ b/src/audio.c
@@ -32,15 +32,23 @@
 #include <unistd.h>
 #include <sys/ioctl.h>
 
-#include <curses.h>
+#ifdef HAVE_CONFIG_H
+# include <config.h>
+#endif
+
+#ifdef HAVE_NCURSES_NCURSES_H
+# include <ncurses/ncurses.h>
+#elif defined HAVE_NCURSES_CURSES_H
+# include <ncurses/curses.h>
+#elif defined HAVE_NCURSES_H
+# include <ncurses.h>
+#elif defined HAVE_CURSES_H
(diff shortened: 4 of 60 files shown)


airween:
Hi Nate,

just a short question: it would be better to solve the ncurses dependency as an easier way? I mean, you create a new header file, eg. tlfcurses.h, which contains the macros, and includes the necessary curses header file:

```C
#ifdef HAVE_NCURSES_NCURSES_H
# include <ncurses/ncurses.h>
#elif defined HAVE_NCURSES_CURSES_H
# include <ncurses/curses.h>
#elif defined HAVE_NCURSES_H
# include <ncurses.h>
#elif defined HAVE_CURSES_H
# include <curses.h>
#endif
```
and then just need to include this header in affected files? Just an idea... :)

73, Ervin

dl1jbe:
I would like to support Ervins idea.

Nate your original idea introduces too much noise into the files. And to be honest if someone adds a new file to tlf he will forget to add that complicated sequence and just use '#include <ncurses.h>.

Another idea would be to add that inlcude sequence to the already existing ui_utils.h and use that file instead ncurses.h and panel.h. That could led us in the long term to move more of the now scattered display code to a common place (ui_utils.c) and get a cleaner code base.

N0NB:
* On 2015 08 Dec 02:31 -0600, Thomas Beierlein wrote:
> I would like to support Ervins idea.
> 
> Nate your original idea introduces too much noise into the files. And
> to be honest if someone adds a new file to tlf he will forget to add
> that complicated sequence and just use '#include <ncurses.h>.

I understand the points both you and Ervin make.  I came at it from the
angle of making it obvious what each source file depends on.

> Another idea would be to add that inlcude sequence to the already
> existing ui_utils.h and use that file instead ncurses.h and
> panel.h. That could led us in the long term to move more of the now
> scattered display code to a common place (ui_utils.c) and get a
> cleaner code base.

That will work, it just won't be obvious if a given source file depends
on [n]curses.h or panel.h is all.  Perhaps toss in a comment after the
included ui_util.h noting that this is for [n]curses.h or panel.h, etc.

The code will compile either way and I can work up a patch (I should
have had this in a separate branch.  Sigh), but it will not be until
later today as I'll be on the road for work all day.

As you're looking further down the road toward separating the UI code
from the other code as much as possible, this is a better idea.

73, Nate

-- 

N0NB:
As an alternative for both of you to consider.  It is possible to add
boilerplate to config.h during the configure process.  The tests and
includes could be put in that file since it needs to be included anyway
for the HAVE constants for the preprocessor tests.

73, Nate

-- 

airween:
Nate,

you mean the autotool checks the existing (n)curses.h header, and put it to all file?

I prefer the unique header in tlf source tree, which contains the necessary header file(s), but it responsible for only one thing, espacially for curses header. For the panel header (if it's also affected by this bug), we need to prepare an another custom header. The tlf_curses.h and/or tlf_panel.h would be better, I think.

73, Ervin

N0NB:
* On 2015 08 Dec 06:34 -0600, Ervin Hegedus wrote:
> Nate,
> 
> you mean the autotool checks the existing (n)curses.h header, and put
> it to all file?

The configure script does the same thing as now by writing the HAVE_*
constants to config.h.  What I am proposing is moving the code I added
to each source file to the end of config.h when configure is run.  This
way, each affected source only needs to include config.h, which is done
now anyway.

> I prefer the unique header in tlf source tree, which contains the
> necessary header file(s), but it responsible for only one thing,
> espacially for curses header. For the panel header (if it's also
> affected by this bug), we need to prepare an another custom
> header. The tlf_curses.h and/or tlf_panel.h would be better, I think.

It was the panel header location in OpenSuSE 42.1 reported by David that
started this mess.  ;-)  Per the RPM I looked at, they install a symlink
for [n]curses.h in /usr/include but have a fixed path of
/usr/include/ncurses for panel.h.  That is what all of these tests are
about.

73, Nate

-- 

dl1jbe:
There is some kind of an compromise we could adopt:

Let config.h define some TLF_NURSES_H and TLF_PANEL_H macros which resolves to the actual file path. Then we could simple '#include TLF_NCURSES_H' or 'TLF_PANEL_H' in our files. Looks nice, avoids having an additional layer of a 'tlf_panel.h' header file and makes it obvious what we want to include. We just need to have a look for new code that the original files are not included directly.

airween:
Yes, so far this is the best solution :)

73, Ervin

N0NB:
* On 2015 08 Dec 07:56 -0600, Thomas Beierlein wrote:
> There is some kind of an compromise we could adopt:
> 
> Let config.h define some TLF_NURSES_H and TLF_PANEL_H macros which
> resolves to the actual file path. Then we could simple '#include
> TLF_NCURSES_H' or 'TLF_PANEL_H' in our files. Looks nice, avoids
> having an additional layer of a 'tlf_panel.h' header file and makes it
> obvious what we want to include. We just need to have a look for new
> code that the original files are not included directly.

You would still have to #include <config.h> in every file anyway to get
the macro.  My idea is to take the code I added to the files and have it
in config.h.  Then a simple #include <config.h> will give the file the
correct header 'for free'.

Let me work up a new patch set over next couple of days so everyone can
take a look and we'll work from there.

73, Nate

-- 

N0NB:
* On 2015 08 Dec 07:56 -0600, Thomas Beierlein wrote:
> There is some kind of an compromise we could adopt:
> 
> Let config.h define some TLF_NURSES_H and TLF_PANEL_H macros which
> resolves to the actual file path. Then we could simple '#include
> TLF_NCURSES_H' or 'TLF_PANEL_H' in our files. Looks nice, avoids
> having an additional layer of a 'tlf_panel.h' header file and makes it
> obvious what we want to include. We just need to have a look for new
> code that the original files are not included directly.

After digging through the GCC and Autotools docs this morning I can find
no place where the actual path can be captured and assigned to a macro
as nice as that might be.

The configure script relies on the default includes search path of gcc
and any path supplied via the CPPFLAGS variable.   For example, here is
what is found in config.log from the AC_CHECK_HEADERS macro looking for
ncurses.h and ncurses/ncurses.h:

configure:5118: checking ncurses.h usability
configure:5118: gcc -c -g -O2 -I/home/nate/local/include conftest.c >&5
configure:5118: $? = 0
configure:5118: result: yes
configure:5118: checking ncurses.h presence
configure:5118: gcc -E -I/home/nate/local/include conftest.c
configure:5118: $? = 0
configure:5118: result: yes
configure:5118: checking for ncurses.h
configure:5118: result: yes
.
.
.
configure:5118: checking ncurses/ncurses.h usability
configure:5118: gcc -c -g -O2 -I/home/nate/local/include conftest.c >&5
conftest.c:61:29: fatal error: ncurses/ncurses.h: No such file or directory
 #include <ncurses/ncurses.h>
                             ^
compilation terminated.
configure:5118: $? = 1
                              

All that is tested is that ncurses.h was found in one of
/home/nate/local/include (specified by passing
CPPFLAGS="-I$HOME/local/include" to configure) or the default includes
search path built into gcc, but we don't know which one nor does the
configure script.  All that is known is that '#include <ncurses.h>'
succeeded and '#include <ncurses/ncurses.h>' did not.  Nothing more.
The 'ncurses.h' file could have been found in /nate/home/local/include,
/usr/local/include, or /usr/include, but which is unknown.  As a result
the configure script goes on to set the HAVE_NCURSES_H macro which gets
written to config.h for this run of the configure script.

The upshot is that using the macro test boilerplate for ncurses.h and
panel.h is necessary to use the correct #include directive.  In the
other pull request commentary is the question of whether to use two
files or one.

Hope that helps explain a somewhat thorny issue.

73, Nate

-- 

N0NB:
Closing request as not applicable.

PR #53, opened 2015-12-11
Ncurses search

N0NB:
I hope this looks better than my prior attempt.  In this patch set the CPP ifdef boilerplate is copied to the generated config.h file.  Source files needing to include [n]curses.h or panel.h will get the proper file simply by Including config.h.

Also, since the AC_CONFIG_HEADERS macro will generate config.h.in automatically, there is no need to carry it in the Git repository so the last patch removes it and sets the AC_CONFIG_SRCDIR to look for the tlf.h file to assure the build system is in the correct archive.

This patch set is to resolve issue #51.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*~
 Makefile
 Makefile.in
 *.[oa]
--- a/config.h.in
+++ b/config.h.in
@@ -6,6 +6,9 @@
 /* Define to 1 if you have the `bzero' function. */
 #undef HAVE_BZERO
 
+/* Define to 1 if you have the <curses.h> header file. */
+#undef HAVE_CURSES_H
+
 /* Define to 1 if you have the <fcntl.h> header file. */
 #undef HAVE_FCNTL_H
 
@@ -57,11 +60,26 @@
 /* Define to 1 if you have the `mkfifo' function. */
 #undef HAVE_MKFIFO
 
+/* Define to 1 if you have the <ncurses/curses.h> header file. */
+#undef HAVE_NCURSES_CURSES_H
+
+/* Define to 1 if you have the <ncurses.h> header file. */
+#undef HAVE_NCURSES_H
+
+/* Define to 1 if you have the <ncurses/ncurses.h> header file. */
+#undef HAVE_NCURSES_NCURSES_H
+
+/* Define to 1 if you have the <ncurses/panel.h> header file. */
+#undef HAVE_NCURSES_PANEL_H
+
 /* Define to 1 if you have the <netdb.h> header file. */
 #undef HAVE_NETDB_H
 
 /* Define to 1 if you have the <netinet/in.h> header file. */
 #undef HAVE_NETINET_IN_H
+
+/* Define to 1 if you have the <panel.h> header file. */
+#undef HAVE_PANEL_H
 
 /* Define if you have POSIX threads libraries and header files. */
 #undef HAVE_PTHREAD
@@ -198,3 +216,21 @@
 
 /* Define to `int' if <sys/types.h> does not define. */
 #undef ssize_t
+
+
+#ifdef HAVE_NCURSES_NCURSES_H
+# include <ncurses/ncurses.h>
+#elif defined HAVE_NCURSES_CURSES_H
+# include <ncurses/curses.h>
+#elif defined HAVE_NCURSES_H
+# include <ncurses.h>
+#elif defined HAVE_CURSES_H
+# include <curses.h>
+#endif
+
+#ifdef HAVE_NCURSES_PANEL_H
+# include <ncurses/panel.h>
+#elif HAVE_PANEL_H
+# include <panel.h>
+#endif
+
--- a/configure.ac
+++ b/configure.ac
@@ -5,7 +5,7 @@ AC_INIT([Tlf],
 	[tlf],
 	[https://github.com/Tlf/tlf])
 AM_INIT_AUTOMAKE
-AC_CONFIG_SRCDIR([config.h.in])
+AC_CONFIG_SRCDIR([src/tlf.h])
 AC_CONFIG_HEADERS([config.h])
 
 dnl Clean compilation output makes compiler warnings more visible
--- a/configure.ac
+++ b/configure.ac
@@ -20,9 +20,13 @@ PKG_CHECK_MODULES(GLIB, glib-2.0)
 AC_CHECK_LIB([m], [atan])
 AC_CHECK_LIB([pthread], [pthread_create])
 ACX_PTHREAD
-AC_CHECK_LIB([ncurses], [initscr],,AC_MSG_ERROR([needs ncurses library]))
+
+AC_CHECK_LIB([ncurses], [initscr],,[AC_MSG_ERROR([needs ncurses library])])
 AC_CHECK_LIB([panel], [update_panels],,
-		AC_MSG_ERROR([needs ncurses panel library]))
+	[AC_MSG_ERROR([needs ncurses panel library])])
+
+dnl Ncurses libraries have been found, now look for the headers.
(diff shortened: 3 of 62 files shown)


N0NB:
Finally, The Autoconf documentation recommends that config.h be included before any other headers in case constants or something needs to be overridden.  I opted to follow the Autotools recommendation.

dl1jbe:
Sorry for being late.

I just learned something new about autotools /AH_BOTTOM) from your patch. Thanks.

To comment to the proposed patch I must say I feel not quite sure if your current solution is the best we can reach:

- For another programmer who do not well know tlf's source code it is difficult to see that including config.h will also include the ncurses headers. He will have a difficult start to understand where the ncurses headers are coming from. That mental gap seems a little bit wide.

You wrote yourself in comments to your last pull request (see https://github.com/Tlf/tlf/pull/52):
> That will work, it just won't be obvious if a given source file depends on [n]curses.h or panel.h is all. 
> Perhaps toss in a comment after the included ui_util.h noting that this is for [n]curses.h or panel.h, etc. 

And that was a for a file with a name suggesting user interface stuff (ui_utils.h).

- The proposed patch has the downside that every file which include config.h will also include ncurses.h and panel.h - even if not needed. That will result in a longer compile time without need. (Even more as I think that all our code should include the config headers.)


Finally, but in some part unrelated to the above,  I am not sure if we can flatten out the #ifdef HAVE_CONFIG_H. It definitely looks better without it, but I am not quite sure if we drop portability or something else. Do you know more about it?

N0NB:
* On 2015 13 Dec 23:36 -0600, Thomas Beierlein wrote:
> Sorry for being late.

No problem.  :-)

> I just learned something new about autotools /AH_BOTTOM) from your
> patch. Thanks.
> 
> To comment to the proposed patch I must say I feel not quite sure if
> your current solution is the best we can reach:
> 
> - For another programmer who do not well know tlf's source code it is
> difficult to see that including config.h will also include the ncurses
> headers. He will have a difficult start to understand where the
> ncurses headers are coming from. That mental gap seems a little bit
> wide.
> 
> You wrote yourself in comments to your last pull request (see
> https://github.com/Tlf/tlf/pull/52):
> > That will work, it just won't be obvious if a given source file
> > depends on [n]curses.h or panel.h is all.  Perhaps toss in a comment
> > after the included ui_util.h noting that this is for [n]curses.h or
> > panel.h, etc.
> 
> And that was a for a file with a name suggesting user interface stuff
> (ui_utils.h).
> 
> - The proposed patch has the downside that every file which include
> config.h will also include ncurses.h and panel.h - even if not
> needed. That will result in a longer compile time without need. (Even
> more as I think that all our code should include the config headers.)

Good reasons that got me to thinking that what really needs to be done
should be contained in the build system.  I need to dig around in the
Autoconf macros a bit more and see what variables are available.  In
short, what I can try next is to determine if ncurses.h and/or panel.h
is found in /usr/include or /usr/include/ncurses and then set
AM_CPPFLAGS in src/Makefile.am to that path.  That *should* allow
leaving the source files as they are.  I'll tinker with this a bit and
see what I come up with.

> Finally, but in some part unrelated to the above, I am not sure if we
> can flatten out the #ifdef HAVE_CONFIG_H. It definitely looks better
> without it, but I am not quite sure if we drop portability or
> something else. Do you know more about it?

As I see it, The #ifdef HAVE_CONFIG_H is kind of extra noise that just
determines where the point of failure is should config.h actually not
exist (actually, we could name the file anything we like, it doesn't
have to be 'config.h', that's just convention).

Here is the logic as I understand it:

With or without #ifdef HAVE_CONFIG_H if config.h doesn't exist the C
preprocessor will generate an error of not being able to open the
included file config.h.

With #ifdef HAVE_CONFIG_H if something really screwy happens and the
build system doesn't pass -DHAVE_CONFIG_H to gcc, or someone tries to
compile a single file without the build system, then config.h won't be
loaded and gcc will generate an error of implicit declaration or the
like.

Note that I haven't ever tested the above scenarios, this is just the
way I read the logic.

To be really safe one could do the following (untested):

#ifdef HAVE_CONFIG_H
# include <config.h>
#else
# error "Build system error, HAVE_CONFIG_H is undefined."
#endif

It adds a bit more noise to the source files but it points everyone to
where the problem is, namely that -DHAVE_CONFIG_H was not passed to gcc,
either by the build system or by manual invocation.  Ordinarily I would
never expect to see that error message.

HTH,

73, Nate

-- 

N0NB:
After pouring over the Autoconf/Automake and CPP documentation, there
really is no way to implement the idea I thought was possible.  At least
not in a clean or portable way.  As I understand it, when the configure
script performs the tests on header files it is using the paths that are
set temporarily by the Autoconf macros and by any user supplied value in
CPPFLAGS (although I am not certain of this).  I have not seen a way to
append a directory name onto any of those paths so when a test
(generated by AC_CHECK_HEADERS, for example) puts '#include
<ncurses/panel.h>' into the test program, gcc will try all the paths
given with the -I option(s) and then the fixed paths of
/usr/local/include and /usr/include in that order (technically the
manual shows two more directories searched between the two, but they
aren't a part of this discussion).  When gcc successfully compiles the
test program, the configure script will set HAVE_NCURSES_PANEL_H in
config.h in this case.

The common way to deal with this is as the first patch set showed, by
testing defined macros and then including a specific file and having
this in the source file.  The idea of putting that boilerplate into
separate file(s) from config.h is sensible.

I'll propose putting the boilerplate into separate files such as
tlf_curses.h and tlf_panel.h which then contain the macro test
boilerplate as originally suggested by Ervin.  I think this is going to
be the cleanest way to handle this for now.  At some point if the UI
stuff gets more separated we can revisit this by putting the macro test
boiler plate into a relevant header file.

Thoughts?

73, Nate

-- 

airween:
Hi Nate,

of course, I agree my suggestion :)
Anyway, I think I'm the last who had joined to Tlf developer group. I'm not an autotool expert (basically, I'm not a C-expert, like you). I had to learn many-many things, Thomas is a really good, and (it's very important) patient mentor. After I reviewed your comments, I think I understand your idea, but don't understand all of your steps - it doesn't matter, I'll do it soon :).
But as I suggested in my previous comment, the most cleaner solution (for a beginner) is that. I'm happy now, because this indicates, I'm on a right way. Thanks guys!

73, Ervin

N0NB:
* On 2015 14 Dec 13:58 -0600, Ervin Hegedus wrote:
> Hi Nate,
> 
> of course, I agree my suggestion :)
> Anyway, I think I'm the last who had joined to Tlf developer group.

I think you've been contributing to Tlf much longer than me.

> I'm not an autotool expert (basically, I'm not a C-expert, like
> you).

Whoa!  I'm no C expert either.  I know it well enough to satisfy my
passion for doing some hacking.  I have no formal training and this is
all part of my amateur radio hobby.

> I had to learn many-many things, Thomas is a really good, and
> (it's very important) patient mentor.

In a project like this it is one thing to know the language the project
uses,  it is quite another to learn and understand the logic flow of the
program.  Programming is much like writing prose, every one has a
slightly different way of thinking and that is reflected in the code.
The next person that comes along has to adopt the first person's thought
process to gain understanding of the program.  To me this is the more
difficult of the two.  Of course, a person wants to rewrite things to
match their own understanding but that is often a temptation that needs
to be resisted for a while.

> After I reviewed your comments, I think I understand your idea, but
> don't understand all of your steps - it doesn't matter, I'll do it
> soon :).  But as I suggested in my previous comment, the most cleaner
> solution (for a beginner) is that. I'm happy now, because this
> indicates, I'm on a right way. Thanks guys!

You're welcome.  We all learn from each other and in the process try to
make contributions that makes the program more reliable and more useful
for others.

73, Nate

-- 

N0NB:
Closing as no applicable.

PR #54, opened 2015-12-12
Added new feature: calculate the QRG in different modulations, as FSK and LSB/USB

airween:
Hi Thomas,
this is a modified version of my previous fldigixmlrpc code. Now Tlf can be distinguished the modes (FSK, LSB/USB), if RIG can tells the mode (with rig_get_mode()). If RIG works in FSK mode, then Tlf set the Fldigi's carrier to 2210Hz always, if user moves it to an another station, and tune the VFO to the calculated QRG. If RIG is in LSB/USB mode, then Fldigi's carrier stay as is, Tlf calculates the accurate QRG, and also tune the VFO.
I've also attached a short RTTY mode README.

### DIFF
--- a/doc/README.RTTY
+++ b/doc/README.RTTY
@@ -1,0 +1,99 @@
+Tlf RTTY howto
+
+2015, Ervin Hegedus, HA2OS
+
+This is a guide for Tlf, how to use it with Fldigi in RTTY mode,
+especially in FSK, LSB or USB modulations.
+
+To work in RTTY, you need to solve two problems: read and demodulate
+RTTY signals (RX), and send your messages (TX). To demodulate the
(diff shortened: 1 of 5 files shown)


dl1jbe:
The code does not compile when you have --disable-hamlib during configure. The patches are easy. I will add them and then merge the code in. 

Thanks Ervin.

dl1jbe:
In fldigixmlrpc.c there are some unreachable code sequences (following a 'break' from the switch)

airween:
Hi Thomas!

Thanks for your notification, I've fixed up them in my last commit:
https://github.com/airween/tlf/commit/0dd07a7b525fe7b945a8378722c9a2e72afc835b

Hope this helps.

PR #56, opened 2015-12-15
Adjust tests in configure.ac

N0NB:
Reposition several tests to be run before the tests for third party
packages.

Temporarily comment out the AC_PROG_GCC_TRADITIONAL and AC_FUNC_STRFTIME
macros as the Autoconf documentation states they are obsolescent.

Temporarily comment out the AC_HEADER_STDBOOL macro as no source file
loads stdbool.h at this time.

Remove some header files from AC_CHECK_HEADERS as they are already
checked by the AC_HEADER_STDC macro.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,6 +15,25 @@ m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AC_PROG_CC
 AC_PROG_INSTALL
 
+# Checks for header files.
+AC_HEADER_STDC
+AC_CHECK_HEADERS([arpa/inet.h fcntl.h netdb.h netinet/in.h \
+	sys/ioctl.h sys/socket.h sys/time.h syslog.h termios.h])
+
+# Checks for typedefs, structures, and compiler characteristics.
+dnl needed? AC_HEADER_STDBOOL
+AC_TYPE_MODE_T
+AC_TYPE_PID_T
+AC_TYPE_SIZE_T
+AC_TYPE_SSIZE_T
+
+# Checks for library functions.
+dnl obsolescent AC_PROG_GCC_TRADITIONAL
+dnl obsolescent AC_FUNC_STRFTIME
+AC_CHECK_FUNCS([bzero floor ftruncate gethostbyname memset mkfifo putenv \
+	select socket sqrt strcasecmp strchr strcspn strdup strpbrk strspn \
+	strstr])
+
 # Checks for libraries.
 PKG_CHECK_MODULES(GLIB, glib-2.0)
 AC_CHECK_LIB([m], [atan])
@@ -23,10 +42,6 @@ ACX_PTHREAD
 AC_CHECK_LIB([ncurses], [initscr],,AC_MSG_ERROR([needs ncurses library]))
 AC_CHECK_LIB([panel], [update_panels],,
 		AC_MSG_ERROR([needs ncurses panel library]))
-
-# Checks for header files.
-AC_HEADER_STDC
-AC_CHECK_HEADERS([arpa/inet.h fcntl.h memory.h netdb.h netinet/in.h stdlib.h string.h sys/ioctl.h sys/socket.h sys/time.h syslog.h termios.h unistd.h])
 
 dnl Check if we want to link the Hamradio control libraries (hamlib)
 AC_ARG_ENABLE([hamlib],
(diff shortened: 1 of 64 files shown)


N0NB:
Tom,

Look this over (as I'm sure you will).  Do you know of any reason the three commented macros should be used?  As I note, two of them are considered obsolete so may cause problems these days, but I suspect care is taken by the Autotools developers to avoid that.  The STDBOOL macro just seemed like an extra bit of cruft as I think Glib provides a bool.

I think it's important to reposition this code before adding code for ncurses header searching.

N0NB:
I've pushed additional commits to this branch/pull request.  This puts the boilerplate into two separate local header files.  It also uses the AX_CURSES* macros from the [Autoconf Macro Archive](http://www.gnu.org/software/autoconf-archive/).

N0NB:
Tom, hold off on this.  We found a bug in one of the macro files and I am contacting the author.

N0NB:
I just pushed a fix for the macros/ax_with_curses_extra.m4 bug that kept the panel library from being correctly detected.  This set should now be ready to merge.

dl1jbe:
A first  quick check looks good. Congrats for the solution Nate.
Maybe we have still to modify Makefile.am so that the new macro file will be distributed with the rest of the sources. You can check by doing a 'make distcheck' and inspecting the tarball afterwards.

N0NB:
* On 2015 17 Dec 00:21 -0600, Thomas Beierlein wrote:
> A first quick check looks good. Congrats for the solution Nate.  Maybe
> we have still to modify Makefile.am so that the new macro file will be
> distributed with the rest of the sources. You can check by doing a
> 'make distcheck' and inspecting the tarball afterwards.

Hi Tom.

The AC_CONFIG_MACRO_DIR macro sets up those rules automatically for us.
I double checked and the macros directory and .m4 files are in the
source tarball.  If we had other files in the macros directory, then a
Makefile.am would be required for them.

73, Nate

-- 

dl1jbe:
Well, I had not quite understood that 'make dist' picks up all files it needs for distribution.
Thanks for checking. 

I had a second look just now and think it looks good. Glad that we could fix that issue.

N0NB:
* On 2015 17 Dec 14:14 -0600, Thomas Beierlein wrote:
> Well, I had not quite understood that 'make dist' picks up all files
> it needs for distribution.  Thanks for checking.

I had a thought today that something a bit different was being done in
Hamlib.  For whatever reason, there is a Makefile.am in its macros
directory.  I may remove it in the future since it is probably
unnecessary.

I did do a test on the main Tlf Makefile.am by adding the following
lines:

# Install any third party macros into our tree for distribution
ACLOCAL_AMFLAGS = -I macros --install

That doesn't really gain anything extra for Tlf as it copies the pkg.m4
macro to the macros directory but the aclocal.m4 file is still created
with all of the standard Autoconf macros included.  Without that stanza
the contents of pkg.m4 are also copied into the generated aclocal.m4

As I see it, neither way has much of an advantage.  If I were doing a
new package I'd probably include the change as I could find a given 3rd
party macro more easily.  If you want to change it, I can send a pull
request.

> I had a second look just now and think it looks good. Glad that we
> could fix that issue.

I have some other changes to the build system that I want to look at to
update things just a bit more.

73, Nate

-- 

dl1jbe:
Am Thu, 17 Dec 2015 15:45:47 -0800
schrieb Nate Bargmann <notifications@github.com>:

> * On 2015 17 Dec 14:14 -0600, Thomas Beierlein wrote:
> > Well, I had not quite understood that 'make dist' picks up all files
> > it needs for distribution.  Thanks for checking.
> 
> I had a thought today that something a bit different was being done in
> Hamlib.  For whatever reason, there is a Makefile.am in its macros
> directory.  I may remove it in the future since it is probably
> unnecessary.
> 
> I did do a test on the main Tlf Makefile.am by adding the following
> lines:
> 
> # Install any third party macros into our tree for distribution
> ACLOCAL_AMFLAGS = -I macros --install
> 
> That doesn't really gain anything extra for Tlf as it copies the
> pkg.m4 macro to the macros directory but the aclocal.m4 file is still
> created with all of the standard Autoconf macros included.  Without
> that stanza the contents of pkg.m4 are also copied into the generated
> aclocal.m4

Thanks for checking it again. 
> 
> As I see it, neither way has much of an advantage.  If I were doing a
> new package I'd probably include the change as I could find a given
> 3rd party macro more easily.  If you want to change it, I can send a
> pull request.
> 
No, just keep it as it is for the moment. Time will tell if it is good.
If not we can still fix it. At the moment it looks very stable.

> > I had a second look just now and think it looks good. Glad that we
> > could fix that issue.
> 
> I have some other changes to the build system that I want to look at
> to update things just a bit more.
> 
Just go ahead and let the good ideas flowing. Your contributions are
very welcome.

73, de Tom


-- 
"Do what is needful!"
Ursula LeGuin: Earthsea
--

N0NB:
* On 2015 17 Dec 22:41 -0600, Thomas Beierlein wrote:

> No, just keep it as it is for the moment. Time will tell if it is good.
> If not we can still fix it. At the moment it looks very stable.

Very well.

> Just go ahead and let the good ideas flowing. Your contributions are
> very welcome.

Thanks!

73, Nate

-- 

Issue #57, opened 2015-12-18
search for ncurses fails with new ax_with_curses.m4 on some systems


dl1jbe:
After configure I got "SysV or X/Open-compatible header file not found" on some systems.

Further inspections shows:
- ncurses got installed with unicode support so /usr/include/ncurses.h and /usr/include/ncursesw/ncurses.h are installed.
- configure found the ncursesw/ncurses.h which are not supported in out tlf_ncurses.h
- The supplied ax_with_curses.m4 is patched to *not* look for ncursesw but anyhow it got found.

So finally:
- The system has a preinstalled ax_with_curses.m4 which is not patched and gets used instead of our file.

Some possible ideas for getting around:
- patch the supplied ax_with_curses.m4 and rename the helper function to a tlf specific one. (Ugly solution)
- See if we can control the with_ncursesw and with_ncurses variables from outside the m4 file 

Other ideas?

N0NB:
By default that macro will search for ncursew before ncurses.  If this is not a problem, and if we'd like to enable the use of "wide" characters in the future, then the easiest is to probably unset the variables I set in the macro file and let it determine which library is available.

As I understand it, ncursesw is backward compatible to ncurses, but I chose to err on the side of caution by only enabling ncurses and curses.

As for the options, I *think* they can be set in configure.ac and will be honored.  At the very least we should test them so that they're sane, but looking at the documentation I don't see a combination that would result in a situation where some curses library is never chosen.

Probably the easiest (best?) solution is to uncomment the changes I made to ax_with_curses.m4 and go with it.  :-)

dl1jbe:
Downside on simple using the original macro is that anytime ncursesw is selected we get the above error message.

So we either have to accept the wide character support or give a comment, e.g.
'At the moment tlf does not support wide character display, please use --without-ncursesw'

Both cases needs an correction of the tlf_[ncurses|panel].h files.

N0NB:
* On 2015 19 Dec 10:39 -0600, Thomas Beierlein wrote:
> Downside on simple using the original macro is that anytime ncursesw
> is selected we get the above error message.
> 
> So we either have to accept the wide character support or give a
> comment, e.g.  'At the moment tlf does not support wide character
> display, please use --without-ncursesw'

I am guessing the library is backward compatible.  I am also guessing
that since most everything is using UTF-8 these days, just sticking with
the ASCII character set is not a problem.

As I see it, I18n is down the road a ways, probably 2.0 or later.

[...]

PR #59, opened 2015-12-19
Revert local changes to macros/ax_with_curses.m4

N0NB:
Addresses the local changes made to macros/as_with_curses.m4 per issue #57.

This should be merged before the other pull request of this morning.


dl1jbe:
First check here confirms the patch. I will have to add a test on a second machine on which issue #57 got observed.  I have access to that machine tomorrow. So let us just delay the merge until then.

PR #61, opened 2015-12-21
New features: BMAUTOADD and BMAUTOGRAB keywords

airween:
New features: with BMAUTOADD and BMAUTOGRAB keywords, Tlf will add to or grab from bandmap the matching callsigns.


dl1jbe:
Hi Ervin, thanks for the patch. Some comments from my side:

- man page states that BMAUTOADD will add the call if it is at least 3 chars long. That seems to be wrong. Here 4 chars are needed.
- Line 1320 is syntactical wrong: The name of a function IS the pointer to its code. The additional '&' is not needed.
- You should rethink your signal handler here. signal handlers are fragile in UNIX. Please do as less work as possible in it. Not all library functions are reentrant and it may be that the signal handler is called inside one of those functions. Better would be to just set a flag registering that the signal has arrived and handle the flag in the normal main loop.

dl1jbe:
An additional hint: Testing bmautograb (and other variables used as bools) by a 'bmautograb > 0' is not correct.
ANY value not zero counts as true in C. So better use 'bmautograb != 0'. (That goes also as a reminder to myself).

PR #65, opened 2016-01-23
Use memmove() for overlapping strings.

N0NB:
Checking Tlf with Valgrind showed that memcpy() was assigning a string
back onto itself.  The Glibc documentation says doing so will result in
undefined behavior and recommends using memmove() instead.


dl1jbe:
Thanks. Applied

PR #67, opened 2016-05-09
Aligned hamlib's passband argument to the new macro: RIG_PASSBAND_NORMAL

airween:
Hamlib (3.0 and 3.1) has a new feature at rig_* functions, where the rig_passband variable is required: there is a new macro, called RIG_PASSBAND_NORMAL. With this, Tlf keeps the current filter settings when starts, switches mode or band.

### DIFF
--- a/src/gettxinfo.c
+++ b/src/gettxinfo.c
@@ -36,6 +36,11 @@
 
 #ifdef HAVE_LIBHAMLIB
 # include <hamlib/rig.h>
+#ifdef RIG_PASSBAND_NOCHANGE
+#define TLF_DEFAULT_PASSBAND RIG_PASSBAND_NOCHANGE
+#else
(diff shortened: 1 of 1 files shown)


dl1jbe:
I assume you mean the new macro RIG_PASSBAND_NOCHANGE. Other wise it looks good
If I understand your code correct it will use the new macros during compile time if available and otherwise fall back to the old one.

airween:
Hi Thomas,
yes - sorry for the typo. And sure, the compiler checks that the new macro exists, and if not, then it uses the "old" (RIG_PASSBAND_NORMAL).
I've compiled the full source tree (after I merged it with upstream master branch) with old and new hamlib versions, and it works as well.

dl1jbe:
Very well. So I will only check for the new hamlib version with my rig.  That should do it.

dl1jbe:
applied and pushed.

PR #68, opened 2016-05-10
sendbuf.c: reworked ExpandMacro

zcsahok:
Hi,

Trying to get a better readable code and also get rid of buffer overruns.

73!

### DIFF
--- a/src/netkeyer.c
+++ b/src/netkeyer.c
@@ -22,6 +22,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <glib.h>
+
 #include "netkeyer.h"
 #include "tlf.h"
 #include "tlf_curses.h"
@@ -29,9 +31,9 @@
 
 int netkeyer_port = 6789;
 char netkeyer_hostaddress[16] = "127.0.0.1";
-int socket_descriptor;
-struct sockaddr_in address;
-struct hostent *hostbyname;
+
+static int socket_descriptor;
+static struct sockaddr_in address;
 
 int netkeyer_init(void)
 {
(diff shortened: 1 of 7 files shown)


dl1jbe:
Hi Zoltan,
thanks for the patch. To split out a replace() function looks like a very good idea.

I tried to understand your work and got two questions. Maybe you can give some background to me:
- You do some replacements unconditionally (% and [) and sometimes (@ and #) you check before if the string does contain the character to be replaced. What is the idea behind that difference?
- I am not quite sure if we always get 0-terminated strings out of the expansion. The strncpy function has the nasty side effect that it does not guarantee the correct string termination if the buffer gets filled. Can you please check again? (As an idea - there is a safe g_strlcpy function in glib. Maybe we shoudl use it here).
73, Tom DL1JBE

zcsahok:
Hi Tom,

Thanks for your feedback.
The patch has no (intended) functional changes. The handling of @ and # is special as it involves side effects: for @ auto-sending state must be updated and for # the serial number shall be sent over the net if needed. So in these cases a check before starting with replacement is necessary. It was also quite hard to understand for me what was going on there...

I'm aware of strcnpy's working, so to be really sure I'll double check the replacement function and also consider using g_strlcpy.

73,
Zoli
ha5cqz

dl1jbe:
Thanks for the clarification, Zoli.
> 
> I'm aware of strcnpy's working, so to be really sure I'll double check
> the replacement function and also consider using g_strlcpy.
> 
Please give me an information when you are ready. Be aware that I will be away until tuesday.

73, de Tom DL1JBE

zcsahok:
I'll let you when I'm finished. Have a nice WE.

73,
Zoli

zcsahok:
Hi Tom,
Fixed replace_n and improved it to do all the work really in-place.
I have a bunch of test cases checked in in src/test. Its Makefile is not integrated to the main autoconfig mechanism, though.

73,
Zoli

dl1jbe:
Very well Zoltan.
I applied and pushed your changes. 
The tests got integrated into my own internal test suite (just started some month ago, not published yet). I normally use cmocka test suite, but the minunit looks nice also.

Thanks, Tom DL1JBE

dl1jbe:
Let us close the old pull request as the new one (#69) reflects the last changes best.

PR #69, opened 2016-06-30
netkeyer: streamlined + fixed buffer overflow

zcsahok:


dl1jbe:
Hi Zoltan, I picked up the easy changes and simplified a little bit more. Thanks.

For the changes to netkeyer() I am not quite sure that it is the best solution yet. I am thinking about something like the following one:

char *buf;
...
case K_RESET:
  buf = g_strconcat("\e0", NULL);   break;
case K_MESSAGE:
 buf = g_strconcat(cwmessage, NULL); break;
case K_SPEED:
 buf = g_strconcat("\e2", cwmessage, NULL); break;
....
sendto()
g_free(buf);

That show clearly what gets sent and we can drop the awful manipulation of buf[0], buf[1] and so on. What do you think?

73, Tom

dl1jbe:
Ok, lets go the way with adding some synthactic sugar for readability.

dl1jbe:
Please check and reopen if needed. Thanks

zcsahok:
Hi Tom, fine with me. Just added a minor fix: as buf is set to all zeros setting buf[0] and buf[1] is not needed.
Zoli

PR #70, opened 2016-06-30
Overdue commits from earlier this year

N0NB:
Hi Tom.

I am pushing out some overdue commits from this winter as I get back to working with TLF some more.

Mostly this changeset consists of addng support for the Home and End keys work in the call input and exchange fields as well as making sure Ctl-A and Ctl-E properly.  It also includes an extensive rewrite to update the man page to the keyboard handling although much work likely remains to align it 100% with the code.


dl1jbe:
Hi Nate,

many thanks for the contribution. Merged in and pushed.

PR #71, opened 2016-08-15
Enable Hamlib CAT PTT if wanted and available

N0NB:
Introduce a new keyword for logcfg.dat, RIGPTT, which is a directive to
test for a Hamlib radio backend's PTT via CAT capability.  If the
backend reports it is capable, Hamlib CAT PTT will be used in place of
the netkeyer PTT if Tlf is compiled with Hamlib support.


N0NB:
Tom, this is mostly a request for you to take a look at this patch set and see what I missed or what incorrect assumptions I made!

dl1jbe:
Hi Nate, sorry for being very late on this. Let me bring holiday and a quite heavy start afterwards on QRL as excuses. Now to the comments...

The code looks mostly good and should work in principle. (I think you have tested it already).
What should be done is a warning in init_tlf_rig() if RIGPTT is not available in the capabilities.

A second problem arises from requesting PTT_ON and _OFF from callinput in one thread and doing the on/off switching in the second thread. In most cases it needs some synchronisation with an mutex for manipulating the rigptt  variable. I am not quite sure if it is needed here. 

As far as I can see you use that RIGPTT only for SSB. Do you think we need similar for CW (if not working FullBK)  and/or digimodes?

N0NB:
* On 2016 21 Sep 10:40 -0500, Thomas Beierlein wrote:
> Hi Nate, sorry for being very late on this. Let me bring holiday and a
> quite heavy start afterwards on QRL as excuses. Now to the comments...

I've been quite busy of late and am now returning to some development
again.

> The code looks mostly good and should work in principle. (I think you
> have tested it already).  What should be done is a warning in
> init_tlf_rig() if RIGPTT is not available in the capabilities.

That's probably a good idea so that at least the user is alerted when
the rig does not support RIGPTT.

> A second problem arises from requesting PTT_ON and _OFF from callinput
> in one thread and doing the on/off switching in the second thread. In
> most cases it needs some synchronisation with an mutex for
> manipulating the rigptt variable. I am not quite sure if it is needed
> here.

I was following the way mode setting is done by using a variable as a
flag value.  I'm still quite a novice when it comes to threading.

> As far as I can see you use that RIGPTT only for SSB. Do you think we
> need similar for CW (if not working FullBK) and/or digimodes?

I coded this up so I could use the voice macros for my participation in
the Kansas QSO Party in August.  I could have used VOX but thought that
this would be a good exercise for eliminating having to run another line
from the computer to the K3.  I use a Perl script as a daemon that
responds to cwdaemon commands while sending the data to my Master Keyer
1 in K1EL (WinKey) emulation for CW.  I have not yet tried TLF and
digital.

It's possible this could be useful for the other modes, I just don't
know.

73, Nate

-- 

dl1jbe:
Applied and pushed to the master branch. Thanks!
I added a warning if RIGPTT is not available. Please check if all is working as expected and then close the request if it is well.

dl1jbe:
As asll is applied I am closing the pull request on behalf on Nate. Thanks.

PR #74, opened 2017-05-11
added option to use backspace in the keyer window

zcsahok:
This is for the fast but not perfectly typers: it allows erasing the last character in the keyer window using backspace/delete/arrow-left key.
Upcoming libcw (and thus cwdaemon) will support removing queued character using ASCII BS (0x08).
By default this feature is disabled so normal use is not affected.
(note: as tlf has no information if a char has been already sent one can always backspace until the first char)


dl1jbe:
applied and pushed. Thanks Zoltan

PR #75, opened 2017-05-22
fix display corruption when switching to a WARC band on the rig

zcsahok:
Unknown (WARC in contest mode or just out-of-band) frequency is mapped to NBANDS index in gettxinfo. This then resulted in a index out of bounds access to bands[] (defined in main.c) leading to some garbage band text.

In the fix bands[] has been extended with an entry for unknown band ("??") and usage of hard coded constants related to NBANDS (8 and 9) have been revisited.

On the same occassion band switching code in callinput.c have been refactored into a common function.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -84,6 +84,26 @@
 void send_bandswitch(int freq);
 int autosend(void);
 int plain_number(char *str);
+void handle_bandswitch(int direction);
+
+#define UP      +1
+#define DOWN    -1
+
+extern int no_arrows;
+extern char hiscall[];
+extern int bandinx;
+extern char band[NBANDS+1][4];
+extern int contest;
+extern int dxped;
+#ifdef HAVE_LIBHAMLIB
+extern freq_t outfreq;
+#else
+extern int outfreq;
+#endif
+extern float freq;
+extern int trx_control;
+extern float bandfrequency[];
+
 
 /** callsign input loop
  *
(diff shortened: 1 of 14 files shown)


dl1jbe:
Hi Zoltan,
please give me some help with your pr. 
Yes bandinx is set to NBANDS for frequnecies out of band (but not for WARC bands). 
I can not observed the garbage text you mentioned. Can you give me a hint where that happened?

I am not quite sure, your proposed extension of bands[] is the right way to go as there are quite some more arrays defined for size NBANDS.
Maybe we should add a 'fictious' bandindex BANDINDEX_NO_BAND to the list and include it in NBANDS. What do you think?

In meantime I picked up your refactoring for the band switching logic. Thanks.

dl1jbe:
Never mind. Found the code in question: update_line puts the actual band in front of the displayed logline.

zcsahok:
Hi Tom,
In fact the issue is not directly with WARC bands as I stated in the comment but rather with old radios (like my IC-751) that select an out-of-band frequency (e.g. 10.045) after band switching. This is a "feature" of the radio that can't be fixed. So when going from 40m to 20m on the radio one has to go over an apparently OOB frequency and thus garbling the band in front of the log line.

I tried to scan and fix uses of constant 8s and 9s that are linked to NBANDS.
BANDINDEX_NO_BAND is also fine with me, it's more clear.

We shall also consider adding support for 60m and cleaning up band related code would help this.

dl1jbe:
Hi Zoltan,
thanks for clarification.
I am just working on the proposed change. Main work is done - I finally called the define BANDINDEX_OOB. I changed all arrays depending on the bandnumber to the new extended range . there were one or two more than you had found.
I still have to check that that change has no unwanted consequences - should be done in one or two days.

> We shall also consider adding support for 60m and cleaning up band related code would help this.

You are right. Both should be done in near future.

dl1jbe:
Hi Zoltan,

sorry for that very late reply. Real life crept in and stole some weeks.
I just extended your idea of an additional OutOfBand entry in the bands arrays (see commit https://github.com/Tlf/tlf/commit/4ffbe728667c376d38ceb8f8ff1cd712695dcc16) . I had to go through the whole code and fix it also on some other array declarations.
In doing so I tried to find all places with a fixed band number and migrated it to use NBANDS instead.

Please have a look and test the result. If it works as expected, please close the pull request.
Tnx, de Tom DL1JBE

zcsahok:
Hi Tom,

Thanks for the update, all works fine, closing the pull req.

73,
Zoli

PR #76, opened 2017-09-12
fixing segv related to failed cluster connection

zcsahok:
Reconnection after an initially failed cluster connection using the change parameters mode leads to segfault.

Steps to reproduce:
1) configure a non-reachable cluster access
2) start Tlf and enter change parameters menu by pressing colon (:)
3) either try to reconnect (RECON) or leave the menu (Enter)
Both produce segv.

The code tries to use curses objects that were freed after the failed connection.
I've added guards against this in two functions. Other functions are non susceptible as they have implicit checks.

Also removed unused variables. In the second commit the header file is a bit cleaned up and public/private functions are separated.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -866,8 +866,7 @@ void tlf_cleanup()
 	pthread_join(background_thread, NULL);
     }
 
-//    commented out for the moment as it will segfault if called twice
-//    cleanup_telnet();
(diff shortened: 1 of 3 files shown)


dl1jbe:
Thanks for the catch and the provided solution.
Just merged in. Sorry for being late.

PR #77, opened 2017-11-25
Switch back to normal keypad mode in logview

zcsahok:
This fixes loss of cursor key functionality once the log view feature has been triggered.
Keypad mode was not restored after executing external pager program.

Steps to reproduce:
- add at least one log line
- press cursor up until log view is started (try to move above the topmost log line)
- exit `less` by pressing `q`
- upon return to the main screen cursor keys stop working (showing only A-B-C-D chars)

### DIFF
--- a/src/logview.c
+++ b/src/logview.c
@@ -53,6 +53,7 @@ int logview(void)
 
 	refreshp();
 
+        keypad(stdscr, TRUE);
 
 	return(0);
 }


dl1jbe:
I can not reproduce the behaviour here. What terminal and $TERM setting do you use?

zcsahok:
I use `TERM=xterm` in a `xfce4-terminal`.
Just checked and this error is not present when using console with `TERM=linux`

dl1jbe:
I can reproduce it with TERM=xterm for xfce-terminal. I am using TERM=xterm-256color here and the problem does not show up.

Before picking up the patch I would like you to test two other cases where we leave curses mode temporarily. They use a different algorithm and maybe we should use that way. 

First try if cursor keys are working after an '!' which starts a temporary subshell (back to tlf with 'exit').
Second, what happens after a ':help' ?

It seems you have to use cursor keys in all that modes (especially left and right) to provoke the problem. Please try and give me a short report.

zcsahok:
Checking with `TERM=xterm`:
- in ! (subshell, bash) cursor keys **do not** work
- after ! + exit (back to TLF) cursor keys work

- in :help (less) cursor keys work
- after :help cursor keys work

zcsahok:
(in `xterm-256color` the colors are quite different from what one has with `linux` on the console. that's why I use `xterm`)

dl1jbe:
At least left/right will work in shell as soon as you enter some characters. But ok, main information from your side: tlf cursor keys are working after '!' and ':help'. So the code used there is working and we can use it also for logview.

The code should look as follows:

strcat (comstr, logifle);

endwin();   /* That drops off curses mode */
rc = system(comstr);
refreshp() /* switch back to curses mode */

I will check and report back.

P.S.: I used 'xterm' all time until my distro dropped the 'xterm-256color' in.

zcsahok:
commit ac5106b fixes the issue

dl1jbe:
Merged into master. Thanks Zoltan.

PR #78, opened 2017-12-01
spawn $SHELL instead of sh

zcsahok:
Use user's setting instead of plain sh when opening a shell using `!`


dl1jbe:
Applied. Thanks Zoltan!
P.S: Btw similar could be used for choosing the editor to use. Feel free to do it if you like.

PR #80, opened 2017-12-06
Correct documentation on Enter key to match observed behavior

### LINKED ISSUE #79: Documentation disagrees with behavior on behavior of Enter key in S&P mode
Recently I acquired a very inexpensive USB serial CW interface.  I was trying to debug why **cwdaemon** turns my call sign from `W6` into `M6`, and `CQ` into `KQ` (i.e. drops the first dah or dit).  As a workaround I asked **tlf** via rules files to send some extra characters to **cwdaemon** to make it "wake up" and send complete characters, e.g.:
```
F6=+- %
```
The [manual page](/Tlf/tlf/blame/35320fcc86b73cf22ef967ba5ee7f1714c7e1902/tlf.1.in#L532) claims that *Enter* sends the `MY` message as if I had hit *F6*.  My ears tell me otherwise; if it were doing that my call would not have been clipped to sound like `M6`.

Other bits of documentation, had I encountered them earlier, would have given me the obvious solution:
```
S&P_CALL_MSG=+- %
```
But the description of what the *Enter* key does in S&P mode is still a bit misleading.

I plan on submitting a pull request with documentation changes to match the observed behavior of the code.

tnalpgge:
Fix #79

### DIFF
--- a/tlf.1.in
+++ b/tlf.1.in
@@ -509,35 +509,59 @@ S&P mode is for tuning up or down the band and answering the calls of other
 stations.
 .
 .IP
-.B CQ Mode:
+.B In CQ Mode:
 .RS 7
 .IP \(bu 2
-With the call input field empty sends the F12 message (Auto CQ).
+With the call input field empty,
+\*(lqEnter\*(rq
+sends the F12 message (Auto CQ).
 .IP \(bu 2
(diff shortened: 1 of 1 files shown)


dl1jbe:
Merged in. Thanks & 73

PR #83, opened 2017-12-26
BMAUTOGRAB only in S&P mode

zcsahok:
Autograbbing should work in S&P mode only. In CQ mode it grabs own call or other unwanted nearby calls cluttering the input field.


dl1jbe:
Merged in. Thanks Zoltan.

PR #84, opened 2017-12-26
In case of SPRINTMODE the TU_MESSAGES was cutted

airween:
Today was the DARC XMAS sprint contest. I realized that Tlf didn't sent the TU_MESSAGE (S&P and CQ), but any other messages was sent. I tunred off the SPRINTMODE, and TU_MESSAGES was sent.

I reviewed the code, and found the work logic in SPRINTMODE:

- logit() logs the QSO
- logit() calls the keyer() and change_mode()
- keyer() starts to send the message
- change_mode() calls the cleanup()
- keyer() (in background) checks the data_ready variable
- meantime cleanup() calls the keyer_flush() which set data_ready to 0
- keyer() found the data_ready as 0, and doesn't start to send the message.

I've take an argument to cleanup(), that describes a mask: which function and assignment should excluded.

This patch solves the problem.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -215,10 +215,10 @@ char callinput(void)
 		}
 	    }
 
-	    /* if BMAUTOGRAB is active and input field is empty and a spot has
+	    /* if BMAUTOGRAB is active in S&P mode and input field is empty and a spot has
 	     * not already been grabbed here check if a spot is on freq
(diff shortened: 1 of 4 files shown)


dl1jbe:
Cleaned up and applied. Thanks Ervin.

PR #85, opened 2018-01-15
limit freq polling to 200 ms; reuse freq2band func

zcsahok:
In the current code the background process delays 10 ms between frequency read-outs (if no other action like lan check has to be done). This means a very frequent (10-50 times/sec) polling that has no real use as the display is updated once per second. Normally a poll each 200-500 ms should be sufficient to get a reasonably live frequency display.  Too frequent polling can have negative impact on other actions (e.g. sending cw via hamlib).

gettxinfo was changed to skip polling if the previous was done within 200 ms.  The code was restructured and some code duplicates removed. The function freq2band is reused as it provides the needed mapping. This serves also to reduce code duplication.

note: outfreq needs to be mutexed (r/w access from two threads), I'll fix that later.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -36,14 +36,7 @@
 #include "tlf.h"
 #include "tlf_curses.h"
 #include "ui_utils.h"
-
-#ifdef HAVE_CONFIG_H
-# include <config.h>
-#endif
-
-#ifdef HAVE_LIBHAMLIB
-# include <hamlib/rig.h>
-#endif
+#include "gettxinfo.h"
 
 #ifdef __OpenBSD__
 # include <soundcard.h>
@@ -394,13 +387,6 @@ int drawSmeter(int xpos, int ypos, int yheight, float testvalue)
 
 int panscan(void)
 {
-
-#ifdef HAVE_LIBHAMLIB
-    extern freq_t outfreq;
-#else
-    extern int outfreq;
-#endif
-
     int rc, j, key = 0;
     float testvalue;
     float FromFrequency = 0.0;
@@ -459,16 +445,12 @@ int panscan(void)
 	    int i;
 
 	    for (i = 0; i < 100; i++) {
-		if (outfreq == 0)
+		if (get_outfreq() == 0)
 		    break;
-		usleep(10000);
+		usleep(10 * 1000);
 	    }
 
-#ifdef HAVE_LIBHAMLIB
-	    outfreq = (freq_t) (frequencies[j] * 1000);
-#else
-	    outfreq = (int) (frequencies[j] * 1000);
-#endif
+	    set_outfreq(frequencies[j] * 1000);
 
 	    usleep(50 * 1000);
 	    testvalue = get_audio_sample();
@@ -496,12 +478,6 @@ int panscan(void)
 
 int nbscan(void)
 {
-#ifdef HAVE_LIBHAMLIB
-    extern freq_t outfreq;
-#else
-    extern int outfreq;
-#endif
-
     int rc, j, key = 0;
     float testvalue;
     float FromFrequency = 0.0;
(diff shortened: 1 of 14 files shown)


dl1jbe:
Hi Zoltan,

thanks for the PR.

As I see it there are 3 different points in your work:

Let us take it step by step:

I) 
> In the current code the background process delays 10 ms between frequency read-outs (if no other 
> action like lan check has to be done). This means a very frequent (10-50 times/sec) polling that has 
> no real use as the display is updated once per second. Normally a poll each 200-500 ms should be 
> sufficient to get a reasonably live frequency display. Too frequent polling can have negative impact 
> on other actions (e.g. sending cw via hamlib).
>gettxinfo was changed to skip polling if the previous was done within 200 ms.

That is not quite true. The polled frequency is used for the AUTOGRAB feature (see start of callinput() function) and is updated more often.
I think you are right, that 50 polls/s are too much, but I would like to experiment a little bit with regards to the responsiveness of the AUTOGRAB feature when you are tuning around the band.  (I will use the contest on weekend to have a look).

II)
> The code (gettxinfo()) was restructured and some code duplicates removed. The function freq2band 
> is reused as it provides the needed mapping. This serves also to reduce code duplication.

Good work. I will pick it up shortly.

III)

> note: outfreq needs to be mutexed (r/w access from two threads), I'll fix that later.

I know about it. There are some more variables which needs such a protection. But go ahead please.

dl1jbe:
Something weird is going on with your pull request:

- Switching to use of freq2band in gettxinfo works without problems.
- Polling only every 200ms seems to work also, even for the above mentioned AUTOGRAB function.

- But there are problems with the restructuring of gettxinfo itself (switch of prototype and refactoring of band switching).
 
When I incorporate that changes the band switching via left/right key does not always work. Sometimes I have to press the key more than once. 

We should look into it more closely.

zcsahok:
I've experienced the issue with the band switching and the same effect appeared when using grabbing.
The root cause of both is the race condition in accessing outfreq. This was made more pronounced by my code change as a reset was added to the end of gettxinfo().

Until now this must have been a sporadic issue that could remain unnoticed. One example use case: grabbing calls grabspot() that sets outfreq in two steps (due to fldigi offset) and then back in callinput outfreq is read into grab.spotfreq. The actual frequency value is either lost or not, depending whether gettxinfo() was executed in the meantime.

I've pushed an update that addresses the locking issue by defining atomic operations for outfreq. The fix is partial, I was only focusing on band switching and grabbing functions. My tests show that both work now correctly, no repeat is needed.

More will follow as it requires quite a number of changes. A secondary goal of these changes is to move all rig (hamlib) related functions into an own file and thus remove the `#ifdef HAVE_HAMLIB` directives now present in almost all files.

dl1jbe:
> I've experienced the issue with the band switching and the same effect appeared when using 
> grabbing. The root cause of both is the race condition in accessing outfreq. This was made more 
> pronounced by my code change as a reset was added to the end of gettxinfo().
>
It seems you are right. And I must confess that the handling of getting and setting rig frequency is a very dark corner we inherited from former maintainer. But it seems it is time to fix it now and forever.

> I've pushed an update that addresses the locking issue by defining atomic operations for outfreq. 
> The fix is partial, I was only focusing on band switching and grabbing functions. My tests show that 
> both work now correctly, no repeat is needed.

I can confirm that the fix is working as expected. Good work.

> More will follow as it requires quite a number of changes. 

So I will wait a little bit longer before I pick up the results. 

> A secondary goal of these changes is to move all rig (hamlib) related functions into an own file and t
> thus remove the #ifdef HAVE_HAMLIB directives now present in almost all files.

That is also a good idea. Maybe you can make it separate pull requests.

dl1jbe:
By the way there is a fix related to the band switching logic you want to cherry pick for your branch. See commit 227af82b0803f11e278fe26ef964c58f69ff30eb227af82b0803f11e278fe26ef964c58f69ff30eb
on master. Now we can again wrap around from 10 to 160 m and vice versa.

zcsahok:
Pushed part 2 incl. the band switching fix. As far as I could test all works.
I couldn't resist to simplify sendqrg and changefreq. Latter has a design issue (rig and displayed freq  can get out of sync), to be checked later.

dl1jbe:
Applied and pushed. Had to resort things a little bit, so no direct merge.

Thanks for sorting that out Zoli.

PR #86, opened 2018-02-06
fix long CQ bug

zcsahok:
This PR fixes the "long CQ bug": when sending a longer series of unanswered CQ's the cursor  sometimes jumps to the exchange field. This causes F1 to send your call instead of CQ, so the op has to press either Esc or Tab to get back to CQ'ing. Quite annoying, but keeps op awake. :-)

An Easter egg version of this "feature" is:
* start tlf
* press F3 14 times
* CW keyer window pops up.

The bug is a result of two issues.
* In `callinput()` key codes returned as `char` although they fit in an `int` only. E.g. F1 is 0x109 and F3 is 0x10B.  On return only the lower 8 bits are kept and thus F1 is transformed into a Tab/Ctrl-I (0x09) and F3 into Ctrl-K (0x0B).
* `callinput()` is supposed to return to `logit()` only when some non-local keypress is to be handled (e.g. Enter).  Instead it returns when its internal counter `i` reaches 14: 
```
for (i = strlen(hiscall); i <= 13; i++) {
```

The idea may have been an optimization to avoid excessive calls to `strlen` and track call length in `i`.
But this doesn't bring anything as this check is done only once per keypress and in the inner loop [`while (x < 1)`] `strlen` and other stuff is happening 100 times a second anyway.

As a side effect I have changed the signature of `logit` from the mystical
`void *logit(void *ptr)`
to a plain
`void logit(void)`
It neither expects nor returns anything.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -105,7 +105,7 @@ extern float bandfrequency[];
 /** callsign input loop
  *
  * \return code of last typed character */
-char callinput(void)
+int callinput(void)
 {
     extern int itumult;
     extern int wazmult;
@@ -169,7 +169,7 @@ char callinput(void)
 
 
     int cury, curx;
-    int i, j, ii, rc, t, x = 0;
+    int j, ii, rc, t, x = 0;
     char instring[2] = { '\0', '\0' };
     static int lastwindow;
 
@@ -179,7 +179,7 @@ char callinput(void)
     printcall();	/* print call input field */
     searchlog(hiscall);
 
-    for (i = strlen(hiscall); i <= 13; i++) {
+    while (strlen(hiscall) <= 13) {
 
 	show_zones(bandinx);
 	printcall();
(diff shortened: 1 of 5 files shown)


zcsahok:
It seems that Travis has been enabled for Tlf. We have to add `.travis.yml` or builds will fail.

dl1jbe:
Right. I added the travis check but got distracted before adding the .travis.yml file. Just added it some minutes ago.

I will pull in the migrated tests from dl1jbe/tlf (I hope to do it tomorrow) before picking up your PR.

dl1jbe:
I fear you have to rebase your PR onto the actual state so that the .travis.yml file is in the branch.

zcsahok:
No problem, will fix if needed.

dl1jbe:
> As a side effect I have changed the signature of logit from the mystical
> void *logit(void *ptr) to a plain void logit(void).
> It neither expects nor returns anything.

That is a remnant from former code  where logit() was started as a separate thread in main which needs the 'mystical' signature. Good catch anyway.

dl1jbe:
Applied and pushed. I did split the commit into two and rebased it on top of master.
Please close your branch.

PR #87, opened 2018-02-19
introduced README.md

zcsahok:
Renamed README and added CI status image. Just to make it look nicer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ Makefile.in
 *.gcno
 tlf
 tlf.1
+README
 autom4te.cache
 aclocal.m4
 compile
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,3 +5,6 @@ SUBDIRS = doc rules src share scripts test
 CLEANFILES = *~
 
 man_MANS = tlf.1
+
+README: README.md
+	grep -v "Build Status" $< > $@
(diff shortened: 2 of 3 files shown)


dl1jbe:
Thinking a little bit longer I think it is ok to go your route for now. We should discuss the problem and make a decision later.

zcsahok:
The main motivation was to add CI status to the front page. It looks more professional. This README is mainly for github, there is another (alas quite outdated) one in `docs`.

I figured out that Automake requires a README file, so the easiest way was to generate it from MD by stripping the CI link. It should fit for the purpose as I didn't really touch the contents and the formatting is also readable as text. 

There are other markups also supported
https://github.com/github/markup#markups, but MD is the de-facto standard.

ADOC looks also nice. We could use it for the real docs, once someone gets to that point...
Another topic for 1.4...

dl1jbe:
> I figured out that Automake requires a README file, so the easiest way was to generate it from MD 
> by stripping the CI link. It should fit for the purpose as I didn't really touch the contents and the 
> formatting is also readable as text.

Yes, that was a good idea.

> ADOC looks also nice. We could use it for the real docs, once someone gets to that point...
> Another topic for 1.4...

For a a start you can have a look at https://git.hub.com/Tlf/tlf.github.com/oldmanual.
I already  converted Reins old Tlf help page to asciidoc. Maybe we can use it as a starter.

PR #88, opened 2018-02-21
Test initial exchange + fix handling DOS line endings

zcsahok:
Added tests for `initial_exchange.c ` to show the issue with DOS text files.
As initial exchange files originate from external sources they can have DOS format also. I ran into this during the RAEM contest.
(there could be other affected places, to be checked) 

The test failure is seen in https://travis-ci.org/zcsahok/tlf/builds/344438125

```
[ RUN      ] test_ok_dos
"�@" != "__NOT_SET__"
test_initial_exchange.c:65: error: Failure!
[  FAILED  ] test_ok_dos
```

The solution is mainly based on `g_strstrip`, it removes all whitespace including CR/LF. Also added a feature to ignore additional columns in the file.

Somewhat restructured `make_ie_list` function and removed unused test method. Made `free_ie_list` private. Applied `astyle` formatting.

There are 2 minor commits also: wrap of `sleep` (I was wondering why `test_addmult` takes 5 secs) and avoiding a warning.


dl1jbe:
> Made free_ie_list private. 

Only downside of making it private is that you can not use it for tests.

zcsahok:
Tests do no free the allocated list, that's true. But I think that causes no issue.

Looked for other places where external file is read: `addmult`, `parse_logcfg`, `readcabrillo`, `rules` and `searchlog` are to be checked.

Will make a minor improvement in `addmult` so that comment lines with leading spaces are also recognized. Apart from that it looks OK (it has `g_strstrip` already).

PR #89, opened 2018-02-23
Reformat code with astyle

dl1jbe:
Added tools/.astylerc as discused on mailing list. Changed only one rule to indent 'case' blocks relative to the surrounding 'switch' statement.

Reformatted most of the code according to the adopted style. There are some minor things left (mostly comments) which do not match the style.  That needs further work.


zcsahok:
Big bang. :-)  Looks good.

Maybe it's better to remove the leading dot from `.astylerc` so that it's clearly visible after cloning on a Unix system. One has to move it anyway to home dir or do something else with it.

dl1jbe:
Am Fri, 23 Feb 2018 19:49:34 +0000 (UTC)
schrieb Zoltan Csahok <notifications@github.com>:

> Big bang. :-)  Looks good.
> 
> Maybe it's better to remove the leading dot from `.astylerc` so that
> it's clearly visible after cloning on a Unix system. One has to move
> it anyway to home dir or do something else with it.
> 
Hmm. I will think about it. Maybe adding a README to the directory
could be of some use too.

73, Tom



-- 
"Do what is needful!"
Ursula LeGuin: Earthsea
--

PR #90, opened 2018-02-23
minor improvement in addmult

zcsahok:
Test DOS text and leading space. Test fails as comment lines with leading space are not recognized.

Will add fix shortly.


zcsahok:
Moved `g_strstrip`  rigth after reading. Taking early exit if no file was found.

PR #91, opened 2018-03-01
Added test for searchlog

zcsahok:
Covering mostly `load_callmaster`. Other functions need curses mocking, and more work...

Allowing leading spaces in callmaster file, just in case. Made reading loop more compact.

To do: opening file from package dir should be done in a common function.


test/test_searchlog.c
```diff
@@ -0,0 +1,152 @@
...
+
+int teardown_default(void **state) {
+    remove_callmaster();
+    return 0;
+}
+
+void test_callmaster_no_file(void **state) {
+    remove_callmaster();
+    int n = load_callmaster();
+    assert_int_equal(n, 0);
```
dl1jbe: That test fails as soon as TLF is already installed. 'load_callmaster()' is also looking in the PACKAGE_DATA_DIR location where a callmaster file is present in that case.

zcsahok:
I see. I was testing on a machine without TLF installed.
As currently reading from data dir is hardcoded I have disabled this test.
After the release I'll add a function to open data file from current dir or data dir and that can be mocked in the tests.

PR #94, opened 2018-03-31
Update man page for more consistency

N0NB:
New keywords, commands, and configuration parameters are now in bold while
value descriptions and files and pathnames are in italics.  Typed strings are
enclosed in double quotes while typed key names are enclosed in single quotes.

Tried to ensure that text is more clear and from an objective point of
discussion.

Added a sample Xresources file for configuring Xterm with Linux console
compatible color scheme for use with Tlf.


dl1jbe:
Applied. Thanks Nate.

PR #96, opened 2018-08-08
Add DIGI_RIG_MODE config option

W8BSD:
This option can be one of RTTY, RTTYR, USB, or LSB.  If set, the
rig mode is changed to it any time digi mode is activated.

The mode is also set on startup.

This commit also cleans up the commads array so that MAX_COMMANDS
doesn't need to be updated every time a new command is added.

### DIFF
--- a/src/sendqrg.c
+++ b/src/sendqrg.c
@@ -26,6 +26,7 @@
 #include "startmsg.h"
 #include "gettxinfo.h"
 #include "bandmap.h"
+#include "globalvars.h"
 
 
 void send_bandswitch(int trxqrg);
@@ -214,6 +215,18 @@ int init_tlf_rig(void) {
 	sleep(10);
 
     }				// end debug
+
+    switch (trxmode) {
+	case SSBMODE:
+	    set_outfreq(SETSSBMODE);
+	    break;
+	case DIGIMODE:
+	    set_outfreq(SETDIGIMODE);
+	    break;
+	case CWMODE:
+	    set_outfreq(SETCWMODE);
+	    break;
+    }
 
     sleep(1);
     return (0);
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -441,6 +441,7 @@ int changepars(void) {
 	}
 	case 32: {		/* DIGIMODE  */
 	    trxmode = DIGIMODE;
+	    set_outfreq(SETDIGIMODE);
 	    break;
 	}
 	case 33: {		/* PACKET  */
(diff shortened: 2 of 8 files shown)


src/sendqrg.c
```diff
@@ -215,6 +216,18 @@ int init_tlf_rig(void) {
 
     }				// end debug
 
+    switch (trxmode) {
+	case SSBMODE:
+	    set_outfreq(SETSSBMODE);
```
airween: You will set the freq to -1, -2 or -4 Hz? Why? :)

K6BSD:
On 08/08/18 06:30, Ervin Hegedus wrote:
>
> *@airween* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In src/sendqrg.c 
> <https://github.com/Tlf/tlf/pull/96#discussion_r208532097>:
>
> > @@ -215,6 +216,18 @@ int init_tlf_rig(void) {
>   
>       }				// end debug
>   
> +    switch (trxmode) {
> +	case SSBMODE:
> +	    set_outfreq(SETSSBMODE);
>
> You will set the freq to -1, -2 or -4 Hz? Why? :)
>

This is the same way it's done in changepars()
https://github.com/Tlf/tlf/blob/master/src/changepars.c#L439

The negative values don't show up on the screen at all.

I'm not sure of the history on why SETSSBMODE and SETCWMODE are done 
this way, but making the same call there works and having it set the 
mode at startup is useful.  I agree that setting the output frequency to 
a negative number to change the mode is odd, but that's what tlf 
currently does in other places.

¯\_(ツ)_/¯

airween:
Hi Stephen,

> This is the same way it's done in changepars()
> https://github.com/Tlf/tlf/blob/master/src/changepars.c#L439

I've saw meantime, that's why I removed my comment - sorry for
that.
 
> The negative values don't show up on the screen at all.

sure,
 
> I'm not sure of the history on why SETSSBMODE and SETCWMODE are done 
> this way, but making the same call there works and having it set the 
> mode at startup is useful.  I agree that setting the output frequency to 
> a negative number to change the mode is odd, but that's what tlf 
> currently does in other places.
> 
> ¯\_(ツ)_/¯
> 

yes, you're right - sorry for the disturb, and thanks for the
patch.


73, Ervin

dl1jbe:
Thanks for all the pull requests. As I am only 24h away from a holiday trip I can pick up the changes only as soon as I am back in early september (just to let you know).

Just keep the patches coming. All help is appreciated.

73, Tom

dl1jbe:
Apllied & pushed. Thanks.

PR #97, opened 2018-08-08
Fix text.get_rx.  Was using start:end instead of start:length

W8BSD:


dl1jbe:
Looks good. Thanks. Applied and pushed.

PR #98, opened 2018-08-08
malloc() is defined in stdlib.h.

W8BSD:
stdlib.h has been the correct header for malloc() since C89.


dl1jbe:
Merged and applied. Thanks

PR #100, opened 2018-08-08
POSIX doesn't define the -w option

W8BSD:
It seems to be a GNU extension, and doesn't seem to be needed in
this use case.  The FreeBSD uniq utility doesn't support it.


dl1jbe:
>It seems to be a GNU extension, 

Not quite sure, but reading the POSIX manual confirms that it is not common.

> and doesn't seem to be needed in this use case. 

Yes we should check the whole line including the trailing QRG

Applied and pushed. Thanks.

PR #101, opened 2018-08-08
Support CT9 cty.dat files

W8BSD:
This adds three new overrides, lat/lon, continent, and time offset.
showinfo() now needs a prefix, so create add a getctydata_pfx()
function which returns the prefix rather than the country.

If a callsign in cty.dat starts with an '=', it will only use
exact matching, not prefix matching.  To allow backward compatability,
Tlf remembers if it saw any '=' prefixes and if so, find_full_match()
will only match those.  If there were no '=' prefixes, the behaviour
is as before, and it will match any prefix.

### DIFF
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -110,7 +110,7 @@ void calledit(void) {
 		hiscall[j] = hiscall[j + 1];	/* move to left incl. \0 */
 	    }
 
-	    showinfo(getctydata(hiscall));
+	    showinfo(getctydata_pfx(hiscall));
 
 	    if (cnt > 1)
 		searchlog(hiscall);
@@ -128,7 +128,7 @@ void calledit(void) {
 		    hiscall[j] = hiscall[j + 1];
 		}
 
-		showinfo(getctydata(hiscall));
+		showinfo(getctydata_pfx(hiscall));
 
 		if (cnt > 1)
 		    searchlog(hiscall);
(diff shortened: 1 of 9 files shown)


dl1jbe:
Thanks for the good contribution. The ability to use the new CT9 syntax was on my ToDo list for quite some time. I would like to pick it up shortly.

I have only one concern about the use of TRUE and FALSE here. Until now we have avoided the coding of boolean values with that words for consistency and used 0 and 1.  Furthermore the two definitions are inherited from ncurses.h. Better would be to define them by ourself.

I had planned to switch 'all' boolean values to real boolean variables by use of stdbool.h in next releases. That should be supported by most but old compilers.
I think we should go that way also for your case here. 

Any thoughts?

W8BSD:
My feeling is that using a literal 0 and 1 is the worst possible choice, and using stdbool the best.  The main reason I didn't want to introduce stdbool initially is that adding -std=c99 (or std=gnu99) to CFLAGS seemed like too much work, and I'm not sure just how strict you're trying to be.

I've updated this pull request to use stdbool (but not use std=).

dl1jbe:
Very well, Stephen. Yes, the question if we should force -std=c99 has been around for quite some time here. I had a look into the default std-values for the last gcc versions recently. On one hand it seems that stdbool.h is supported from gnu90 on, on the other hand gnu90 or c99 (or newer) seems to be the default configuration for actual compilers.
So I think we should use stdbool.h here as some test case and see if there are still problems out there. If so, we can correct them in next version, if not we can switch more code to the use of the bool data type.

dl1jbe:
Applied and pushed. Added some minor modifications to make use of bool data consistent.

PR #102, opened 2018-08-09
Fix fldigi keying.

W8BSD:
Use of the keyer with fldigi was causing PTT flapping and partial
sends.  Instead of stopping TX every sent string, track tx state,
and only send "^r" when there's nothing left to send.

### DIFF
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -67,7 +67,6 @@ extern char call[];
 extern int trxmode;
 extern int digikeyer;
 extern int trx_control;
-extern int use_fldigi;
 
 int cw_simulator(void);
 
@@ -118,7 +117,7 @@ void *background_process(void *ptr) {
 	 *   fldigi_get_log_call() reads the callsign, if user clicks to a string in Fldigi's RX window
 	 *   fldigi_get_log_serial_number() reads the exchange
 	 */
-	if (digikeyer == FLDIGI && use_fldigi == 1
+	if (digikeyer == FLDIGI && fldigi_get()
 		&& trx_control == 1) {
 	    if (fldigi_rpc_cnt == 0) {
 		fldigi_xmlrpc_get_carrier();
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -100,7 +100,6 @@ int changepars(void) {
     extern char sc_volume[];
     extern int cwstart;
     extern int digikeyer;
-    extern int use_fldigi;
 
     char parameterstring[20];
     char parameters[52][19];
@@ -699,13 +698,11 @@ int changepars(void) {
 	}
         case 51: {              /* FLDIGI - turn on/off */
 	    if (digikeyer == FLDIGI) {
-		if (use_fldigi == 0) {
-		    use_fldigi = 1;
+		if (fldigi_toggle()) {
 		    fldigi_clear_connerr();
 		    mvprintw(13, 29, "FLDIGI ON");
 	        }
 	        else {
-		    use_fldigi = 0;
 		    mvprintw(13, 29, "FLDIGI OFF");
 		}
 		refreshp();
--- a/src/fldigixmlrpc.c
+++ b/src/fldigixmlrpc.c
@@ -67,9 +67,11 @@ extern int use_fldigi;
 
 int fldigi_var_carrier = 0;
 int fldigi_var_shift_freq = 0;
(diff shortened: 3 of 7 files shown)


airween:
Could you describe, what does it mean "causing PTT flapping and partial sends"? I usually use RTTY mode with Fldigi, but didn't get any similar issue.

Some notes about your patch:
- src/fldigixmlrpc.c:322 - this removed part looks the RX/TX state of Fldigi. Not just Tlf can send message to Fldigi (eg. Fldigi can send from itself with own macros), and Tlf can "overwrite" the sent message (eg. AUTO CQ with wrong timing). I think this is an important part, would be keep in code
- src/write_keyer.c:120 (in new version) - this part always executed when background_process() calls write_keyer() (and the buffer is empty), and fldigi_stop_send() decides that sending of ^r is required or not. I think the previous version is more simple: after we sent the message to Fldigi, sent a ^r sequence.

I'ld like to understand your issue, and of course, would like to help to fix that.

W8BSD:
So if I open the keyer with CTRL-K and quickly type "This is a test of the keyer", fldigi gets multiple short strings ending with ^r, and some of them are aborted while sending.  Not only does sending ^r after each poll from the background thread cause issues, but stopping the TX while it's sending from the last time the background thread sent causes delays and restarts.

Essentially, what was happening is that the background process would wake up, notice there's text to be sent right after the first key was entered.  It would then send "T^r" to fldigi.  Fldigi would key the rig and start sending 'T'.  While 'T' is being sent, the background thread wakes up and notices that "his i" is in the input box, discover fldigi is in TX mode, switch to RX, sleep two seconds, then send "his i^r".  fldigi would then start sending "his i", which would again be aborted, and "s a test of the keyer^r" send.  So what ends up happening is this:

<Rig in TX>
"T"
<Rig in RX>
<2 second delay>
<Rig in TX>
"h"
<Rig in RX>
<2 second delay>
<Rig in TX>
"s a test of the keyer^r"

Basically, the CTRL-K keyer is not usable with fldigi, which means that instead of being able to send from tlf, you have to switch to fldigi to send arbitrary text.

If you're sending from fldigi and tlf at the same time, it's not any better to have tlf replace what you're sending with an auto CQ than it is to have the auto CQ inserted into what you're typing without clearing what you've typed or the macro you're sending.

airween:
> So if I open the keyer with CTRL-K and quickly type "This is a test of the keyer",

Well, I've never used Tlf own keyer :), so perhaps that's why I didn't understood your issue. :)

Now when I open the keyer with CTRL+K, I can type the message, but I can't send it to Fldigi. The ENTER only makes spaces... How can I send the message? I've tried in two terminal types (MATE and Gnome terminal). And - of course - the typed message also not sending to Fldigi (but the macros works as well).

Anyway, if the error triggered in KEYER window, may be it would be good to handle at there, I mean, the CTRL+K opens a window with title "CW Keyer", but that's _not_ CW keyer, it's DIGI Keyer now. Then we could avoid to send the messages by characters/words, and could send it as "message" to Fldigi/other modem.

> If you're sending from fldigi and tlf at the same time, it's not any better to have tlf replace what you're sending

sure, but as is in Highlander: "There can be only one" :), and I had to decide, which is the more important. We can replace it, that if Fldigi in TX state, then Tlf doesn't send the message...

So, I prefer to ignore this modification, and make a DIGI Keyer.

zcsahok:
I agree @W8BSD, the Ctrl-K keyer is not usable with Fldigi.

Current message sending was made with the predefined texts of Fx keys in mind. There ^r does make sense and is actually necessary. [well, sometimes it's wrong: one can't send (enqueue) two messages, e.g. own call when calling back]

This, however, doesn't fit the keyboard typing use case. Even adding a bit delayed ^r as suggested would not help as from "A^rB^r" typed in Fldigi only A is sent.

The difference to a CW keying is that in this case TLF has to explicitly take care of T/R switching. Taking this a bit further we could say that when using Ctrl-K then the user has to control T/R switching. The are provisions for this for the (I guess mostly historical) MFJ1278 keyer:  `{` and `}` keys switch to TX and RX, respectively, and there is no automatic switching. Just like when operating Fldigi.

IMO it would make sense to consider this approach since then Ctrl-K could effectively replace direct Fldigi manipulation.

dl1jbe:
Fixed in PR #171- Closing.

PR #105, opened 2018-08-09
Fix warnings.

W8BSD:
While many of these return codes should be checked, none of them
actually are, so having the return codes stored in variables is
pointless.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -378,7 +378,7 @@ int drawSmeter(int xpos, int ypos, int yheight, float testvalue) {
 
 int panscan(void) {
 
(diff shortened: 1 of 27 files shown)


W8BSD:
More work needed... update pending.

W8BSD:
Ok, all warnings seem to be fixed on my system, and in Travis now.

dl1jbe:
Looks like an semicolon too much crept into the IGNORE macro. Dropped it. 
Otherwise thanks. 
I did further sort and merge the single commits a little bit to reduce noise. 
Applied & pushed.

PR #106, opened 2018-08-10
Separate messages for RTTY

W8BSD:
This one includes #96, and really needs a discussion since it's not as straight-forward as the rest.

RTTY contests use a space to terminate a message, not a linefeed, so the existing CW macros which all end in a linefeed aren't optimal for RTTY contests.  As a work-around, this changes the training space to a newline when you enter digi mode, and changes it back when you enter cw mode.

It may make more sense to have completely separate macros for CW/DIGI, especially since RTTY macros should actually start with a CR or CRLF.

### DIFF
--- a/src/autocq.c
+++ b/src/autocq.c
@@ -28,6 +28,7 @@
 
 #include "clear_display.h"
 #include "cw_utils.h"
+#include "globalvars.h"
 #include "printcall.h"
 #include "sendbuf.h"
 #include "stoptx.h"
@@ -43,7 +44,6 @@ int play_file(char *audiofile);
 int auto_cq(void) {
     extern char mode[];
     extern char message[][80];
-    extern char ph_message[14][80];
     extern int cqdelay;
     extern int cqmode;
     extern int trxmode;
@@ -58,10 +58,7 @@ int auto_cq(void) {
     strcpy(mode, "AUTO_CQ ");
     clear_display();
     while (delayval == 0) {
-	if (trxmode == CWMODE || trxmode == DIGIMODE) {
-	    sendmessage(message[11]);
-	} else
-	    play_file(ph_message[11]);
+	send_standard_message(11);
 
 	mvprintw(12, 29 + strlen(hiscall), "");
 
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -19,6 +19,7 @@
  */
 
 
+#include <assert.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
@@ -43,7 +44,6 @@
 #include "qtcvars.h"
 
 
-extern int stop_backgrnd_process;
 extern int this_second;
 extern int cluster;
 extern int packetinterface;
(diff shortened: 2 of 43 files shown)


airween:
> RTTY contests use a space to terminate a message, not a linefeed

Could you show me/us some information about this? Now I've checked it, but didn't found any relevant information.
Anyway, your last modification totally breaks the sending of QTC's on WAE-RTTY, where the \n is as is: separates the lines in QTC block.

Some more details: in most cases (I think the only one exception is the QTC's) all RTTY messages are in one line. Tlf sends the message to the modem (usually the Fldigi), and doesn't care, what is the end of the line, the terminate sequence is ^r (for the Fldigi). Now I don't see, if we replace the \n to space, what changes.

airween:
Ah, thanks - the sent link showed me the essence, now I understand. And sorry, I didn't realize that you just replace characters in the messages (macros), not all sent messages.

Now everything is clear, thanks for it :).

The separated digi messages are good thing - I think it would be good to discuss on mailing list. I prefer that do not always have to re-read the configuration when I switch the mode, and edit the messages before that. But may be some user don't like it...

W8BSD:
Ok, updated with separated digi messages.  Just realized I forgot to update the man page as well... should have that done in a few.

W8BSD:
Closing to trigger a new Travis build...

W8BSD:
Restarting for Travis again... it's having problems fetching hamlib.

dl1jbe:
As far as I can see, there seems to be some problems in the way you allocate the digi_messages.

In many places you access the content of the message with something like

  strcpy(buffer, digi_message[i], strlen(digi_message[i]));

strlen() requires that the pointer it uses as start of string is *not NULL* or result in a segfault. How do we guarantee that (E.g. in the case that read_rules can not allocate a buffer with asprintf())?

A further problem I see in the start of parse_logcfg() (line 2066) where you try to free the allocated fields from last run. What happens in the first run when the digi_message field is uninitialized and maybe contain garbage (I know most C startup zeroes out uninitialized data, but that is not guaranteed)? Maybe we should initialize the field at startup to NULL pointers?
After parsing the logconf.dat read_rules() fill in all fields without

dl1jbe:
> > How do we guarantee that (E.g. in the case that read_rules can not allocate a buffer with asprintf())?
> 
> An error is returned if asprintf() fails. No matter what is done after that error, the program will behave incorrectly when the macro is accessed... ie: if someone continues after the error and sends that macro, the program cannot possibly behave correctly. Attempting another minimal allocation for a zero-length string would almost certainly fail as well, so it's not a reasonable band-aid.
> 
> Since the user explicitly accepted that there was an error and decided to continue anyway, the reason for the crash should be obvious. Adding explicit error handling everywhere just in case a user ignores failures just to have a "less bad" unrecoverable failure seems like a terrible design constraint.
> 
Right, but how about pointing to an preallocated static string of zero length?

char *dummy = "";
if (asprintf(..) == -1) {
  digi_message[i] = dummy;
  ...

That avoids surprises for our users. 

> > What happens in the first run when the digi_message field is uninitialized and maybe contain garbage (I know most C startup zeroes out uninitialized data, but that is not guaranteed)? Maybe we should initialize the field at startup to NULL pointers?
> 
> Global variables _are_ implicitly zeroed... that's guaranteed by all versions of C. It can't contain garbage.
>
Not quite true. I have some background in embedded programming. Here the startup code (crt0) is part of your program and often adapted. One often used possibility is to drop initialization of BSS for minimizing boot up time.

But ok, thanks for the pointer to the C standard document.  I will think about it tomorrow.
 
> From the C standard:
> 
> > If an object that has static storage duration is not initialized explicitly, it is initialized implicitly as if every member that has arithmetic type were assigned 0 and every member that has pointer type were assigned a null pointer constant
> 
> Note that "static storage duration" doesn't just mean variables declared static, it includes global variables as well.

W8BSD:
> Thinking some minutes longer, there is still the problem that we can not free that later. So the code should be more like
> 
> char *dummy = "";
> if (asprintf(..) == -1) {
> digi_message[i] = strdup(dummy);

Yeah, that's the minimal allocation I mentioned.  If an 83 (or whatever) character asprintf() fails, a 1 byte strdup() is almost guaranteed to fail as well.

dl1jbe:
> > Thinking some minutes longer, there is still the problem that we can not free that later. So the code should be more like
> > char *dummy = "";
> > if (asprintf(..) == -1) {
> > digi_message[i] = strdup(dummy);
> 
> Yeah, that's the minimal allocation I mentioned. If an 83 (or whatever) character asprintf() fails, a 1 byte strdup() is almost guaranteed to fail as well.

Silly me. I should not talk about code in the middle of the night. 
You are quite right. It can not work this way.

So in that case we have an 'out of memory' which shortly after will lead to more major problems. I think in this case we should report that to the user and exit the program. The returned status and the following request for confirmation to continue is thought for wrong configuration in logcfg.dat and rules file which will change the behaviour of the program but not led to a later crash. I will change your pull request accordingly.

dl1jbe:
Did some reordering in the sequence of commits and made allocation error fatal. No changes otherwise.
Applied & pushed. Many thanks Stephen.

PR #107, opened 2018-08-10
Clean up various race conditions in fldigi stuff, and most importantly,

W8BSD:
have fldigi_xmlrpc_query() return -1 when use_fldigi is false.

Returning success resulted in uninitialized data being used by callers,
making it unsafe to ever turn off fldigi from tlf.

### DIFF
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -67,7 +67,6 @@ extern char call[];
 extern int trxmode;
 extern int digikeyer;
 extern int trx_control;
-extern int use_fldigi;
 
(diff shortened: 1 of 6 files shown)


W8BSD:
I initially pushed this into #102, but since you don't plan on accepting that, I pulled it out separately here.

dl1jbe:
Applied and pushed. Thanks.

Fixed missing release of lock around line 148 and renamed fldigi_get() to fldigi_isenabled() for a more clean understanding.

PR #108, opened 2018-10-11
add lan port config directive

etamme:
This is a WORK IN PROGRESS, I am having trouble getting the lan port to parse from parse_logcfg.c

I added an extern declaration of lan_port within the parse function, as well as declaring it globally in lancode.c, and then again as extern within resolveService().

I double checked the command array, and ALT_DK10 was def. index 262, not 263, so i modified the switch statement and added the LAN_PORT config directive to both the commands array, and the case, as 263.  

In lancode.c I modified resolveService() to utilize the parsed directive via the extern declaration of lan_port.  I left the code in place to auto resolve based on service entry if the lan_port was not set.

Any help / pointers on getting this working would be appreciated.  

The end goal is to be able to run multiple networked instances of TLF on the same server.

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -65,6 +65,8 @@ int send_error_limit[MAXNODES];
 //--------------------------------------
 /* default port to listen for incomming packets and to send packet to */
 char default_lan_service[16] = "6788";
+/* lan port parsed from config */
+int lan_port = 0; 
 
 int lan_active = 0;
 int send_error[MAXNODES];
@@ -91,13 +93,22 @@ int resolveService(const char *service) {
     struct servent *service_ent;
     service_ent = getservbyname(service, "udp");
     int port = 0;
-    if (service_ent != NULL) {
-	port = service_ent->s_port;
-    } else if (strlen(service) > 0) {
-	port = atoi(service);
-    }
-    if (port == 0) {
-	port = atoi(default_lan_service);
+    extern int lan_port;
+    /* if we have no lan port from parse config */
+    if(lan_port == 0 ) {
+      /* lookup port based on services db - does .. this ever work? */
+      if (service_ent != NULL) {
+  	port = service_ent->s_port;
+      } else if (strlen(service) > 0) {
+  	port = atoi(service);
+      }
+      /* if no port was returned from service entry, use the default port */
+      if (port == 0) {
+        /* any reason not to store default port as int instead of char arr?*/
+        port = atoi(default_lan_service);
+      }
+    } else {
+	port = lan_port;
     }
     return port;
 }
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -66,7 +66,7 @@ int send_error_limit[MAXNODES];
 /* default port to listen for incomming packets and to send packet to */
 char default_lan_service[16] = "6788";
 /* lan port parsed from config */
-int lan_port = 0; 
+int lan_port = 0;
 
 int lan_active = 0;
 int send_error[MAXNODES];
@@ -98,9 +98,9 @@ int resolveService(const char *service) {
     if(lan_port == 0 ) {
       /* lookup port based on services db - does .. this ever work? */
       if (service_ent != NULL) {
-  	port = service_ent->s_port;
+	    port = service_ent->s_port;
       } else if (strlen(service) > 0) {
-  	port = atoi(service);
+	    port = atoi(service);
       }
       /* if no port was returned from service entry, use the default port */
       if (port == 0) {
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -571,7 +571,7 @@ int parse_logcfg(char *inputbuffer) {
 	"DKF12",
 	"DKCQM",			/* 250 */
 	"DKSPM",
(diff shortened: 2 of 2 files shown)


airween:
Hi Eric,

looks like there was a missed comma in parse_logcfg.c, and the line number was invalid: you assumed that was 263, but the line 254 at the end of "DKSPC" keyword, the comma missed.

Here is the fix:
https://github.com/airween/tlf/commit/902dd1d8afe41929e41054f7d41395913bb693d0

73, Ervin

(sorry, I've modified some other lines with correct indentation :))

etamme:
Hey Ervin,

Thanks for noticing that error, and for the whitespace fixes - sorry I need to switch tabs I gather.  

it looks like setting "lan_active=1;" got dropped.

I cherry picked your commit, and added lan_active=1 back to the 263 case and all appears to be good now.  Thank you!.

etamme:
So .. actually now that I have added that back - I guess there is no reason to activate the lan unless you have other nodes, which I guess is the point.  So I will remove it except for my testing.

airween:
Eric,

I think the lan_active=1 doesn't need here. What happen, when you put the LAN_PORT keyword, but you don't set up any other node (with ADDNODE/THISNODE)?

I've checked your patch without lan_active=1 in block of LAN_PORT section, and it works as well.

etamme:
I agree, better to leave lan_active only when other nodes have been added.SUCCESS# netstat -lnp | grep tlfudp 0 0 0.0.0.0:9992 0.0.0.0:* 1666/tlfudp 0 0 0.0.0.0:50620 0.0.0.0:* 1666/tlfudp 0 0 0.0.0.0:45311 0.0.0.0:* 1671/tlfudp 0 0 0.0.0.0:9991 0.0.0.0:* 1671/tlf-Eric / KK0ECTOn October 11, 2018 at 4:02 PM Ervin Hegedus <notifications@github.com> wrote:  Eric,I think the lan_active=1 doesn't need here. What happen, when you put the LAN_PORT keyword, but you don't set up any other node (with ADDNODE/THISNODE)?I've checked your patch without lan_active=1 in block of LAN_PORT section, and it works as well.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.

etamme:
So I have two directories setup, logA and logB each with its own logcfg.dat and the nodes are configured on ports 9991 and 9992 as node A and node B respective.  

I start tlf in each directory and pass the -f flag to specify the config.  

No matter what, TLF complains of a node conflict, however I can see that the nodes are sending traffic back and forth to each other with the proper node ID's etc.

See ngrep of node communication here https://gist.github.com/etamme/dbb20ee3b144e702eb13e2ed65f7e230 


Im digging now as to why it thinks there is a conflict

airween:
Guess your modification in resolveService() is wrong, because in case of both directions (send, receive) init function(s) use it. But in that function, you overwrite the passed argument with the value of lan_port, if it set.

It'ld be better to roll back the modification of resolveService(), add the lan_port value (after the string convert) to default_lan_service variable. Now the default will be the new explicit value, but when Tlf calls the lan_send_init(), all configured nodes will have its own port, what you set in ADDNODE keyword.

Hope this helps - let me know if you have any question.

etamme:
Thanks again for that info Ervin.  I didn't realize resolveService was called on RX as well as on the lan init.  I will revisit the code paths, as well as looking at your patch.

Thanks again for all your help.

etamme:
Now that I have commented above - I am actually not sure that resolveService is wrong - if you look at the trace, tlf is sending to and from the appropriate ports - for some reason its confused about "thisnode" e.g. A , B, C etc.  That is the only way the message gets generated

etamme:
I dug into this further.Even with Ervin's modificationshttps://github.com/airween/tlf/commit/c6565ff95d8c223f2f3f25e828e6dd77174e9a31The issue persists.  I determined that the nodes are sending to themselves.  For example - node A, configured to listen on port 9991, with an "ADDNODE=192.168.1.192:9992", ngrep shows the following when after I hit enter to log a QSO.U 2018/10/12 08:04:52.961570 192.168.1.192:39929 -> 192.168.1.192:9991A1 40CW 12-Oct-18 14:04 0004 KG7FF 599 599 05 0 .The send_init is calling resolveService and passing the port from the configured node from the ADDNODE directive .. but some how the port is being set back to the lan_port of 9991, which is why the node is saying "CONFLICT" because the first char of the message it has received matches its node ID.I am digging further, any pointers are appreciated.-EricOn October 11, 2018 at 6:06 PM Ervin Hegedus <notifications@github.com> wrote: Guess your modification in resolveService() is wrong, because in case of both directions (send, receive) init function(s) use it. But in that function, you overwrite the passed argument with the value of lan_port, if it set.It'ld be better to roll back the modification of resolveService(), add the lan_port value (after the string convert) to default_lan_service variable. Now the default will be the new explicit value, but when Tlf calls the lan_send_init(), all configured nodes will have its own port, what you set in ADDNODE keyword.Hope this helps - let me know if you have any question.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or mute the thread.

PR #110, opened 2018-10-18
Make Hamlib mandatory 

zcsahok:
As is makes not much sense to use Tlf without rig control and Hamlib is readily available it is made mandatory. Rig control still has to be explicity enable using the RADIO_CONTROL config.
This saves us a number of #if/#endif's cluttering the code.

On mid term we shall consider using Hamlib's freq_t (double, in Hz units) to store frequencies instead of float in kHz. Fractional values in float can lead to loss of precision.

I've also removed guards from config.h includes as this file should be always present.
It has been kept only where needed (e.g. files using HAVE_LIBXMLRPC).

Also added a guard to avoid compilation warning about redefining _GNU_SOURCE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - rm -rf cmocka-1.0.1 cmocka-1.0.1.tar.xz
   
 script: 
-  - autoreconf -i && ./configure --enable-hamlib && make   
-  - make clean && ./configure --enable-hamlib --enable-fldigi-xmlrpc && make
+  - autoreconf -i && ./configure && make
+  - make clean && ./configure --enable-fldigi-xmlrpc && make
   - make check; rc=$?; cat test/run_*.log; exit $rc
 
--- a/INSTALL
+++ b/INSTALL
@@ -47,13 +47,6 @@ autoreconf --install
 before the above commands.
 
 
-If support for Hamlib (>=1.2.8) is wanted, use 
-
-----
-./configure --enable-hamlib
-----
-
-
 By default, tlf will install into /usr/local/bin, together with the
 shell-scripts from the scripts directory. Data files will install into
 /usr/local/share/tlf.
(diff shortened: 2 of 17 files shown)


dl1jbe:
Thanks Zoli. Applied and pushed. 




Am Thu, 18 Oct 2018 13:31:35 -0700
schrieb Zoltan Csahok <notifications@github.com>:

> As is makes not much sense to use Tlf without rig control and Hamlib
> is readily available it is made mandatory. Rig control still has to
> be explicity enable using the RADIO_CONTROL config. This saves us a
> number of #if/#endif's cluttering the code.
> 
> On mid term we shall consider using Hamlib's freq_t (double, in Hz
> units) to store frequencies instead of float in kHz. Fractional
> values in float can lead to loss of precision.
> 
> I've also removed guards from config.h includes as this file should
> be always present. It has been kept only where needed (e.g. files
> using HAVE_LIBXMLRPC).
> 
> Also added a guard to avoid compilation warning about redefining
> _GNU_SOURCE. You can view, comment on, or merge this pull request
> online at:
> 
>   https://github.com/Tlf/tlf/pull/110
> 
> -- Commit Summary --
> 
>   * make hamlib mandatory
>   * avoid redefining _GNU_SOURCE
>   * fix config.h usage
> 
> -- File Changes --
> 
>     M .travis.yml (4)
>     M INSTALL (7)
>     M configure.ac (51)
>     M src/callinput.c (20)
>     M src/changepars.c (8)
>     M src/fldigixmlrpc.c (13)
>     M src/getsummary.c (6)
>     M src/gettxinfo.c (42)
>     M src/globalvars.h (6)
>     M src/logit.c (2)
>     M src/main.c (36)
>     M src/parse_logcfg.c (18)
>     M src/rules.c (3)
>     M src/sendqrg.c (2)
>     M src/sendqrg.h (8)
>     M src/tlf_curses.h (4)
>     M src/tlf_panel.h (4)
> 
> -- Patch Links --
> 
> https://github.com/Tlf/tlf/pull/110.patch
> https://github.com/Tlf/tlf/pull/110.diff
> 



-- 
"Do what is needful!"
Ursula LeGuin: Earthsea
--

PR #112, opened 2018-11-01
fix SegV on missing RIGPORT

zcsahok:
Missing RIGPORT caused SegV due to calling _strlen(NULL)_.

Also fixed some messages texts (e.g. "rig num" -> "rig model") and reduced function complexity by introducing _parse_rigconf_ and _debug_tlf_rig_.


dl1jbe:
Thanks Zoli. Good catch. applied and fixed.

PR #115, opened 2018-12-04
Added test_clusterinfo.c

zcsahok:
In preparation of refactoring clusterinfo.c
XPlanet stuff is intentionally not covered.


zcsahok:
Re C99: shouldn't we switch a more recent Ubuntu version (xenial) that has gcc 5.4.0 which can cope with C99 out of the box?

dl1jbe:
Just picked up the switch to C99 out of the commit chain as a start.
If you could prepare another pull request with only the new test code we do not need to clutter the history so much. Thanks

zcsahok:
OK, will create a new branch+PR

PR #116, opened 2018-12-05
added test_clusterinfo.c

zcsahok:
added clear_mvprintw_history() and renamed the history array


dl1jbe:
Thanks Zoli. Applied and pushed

PR #122, opened 2019-03-24
Switch to consistent Hz internal frequency unit

zcsahok:
Using  hamlib's freq_t instead of float.

Some more clean-up is still needed as freq-band-etc. conversions happen at many places.

Tests shall be added to freq sensitive parts, e.g. Cabrillo/ADIF generation.


dl1jbe:
Thats a big pull request. Thanks for preparing it. 
On a first check only the definition of bandfrequency in test/data.c needs still to be converted.

zcsahok:
Good point, fixed.

dl1jbe:
Applied and pushed. Thanks for sorting things out.

PR #124, opened 2019-04-03
show only valid DX info

zcsahok:
suppressed displaying partial and incorrect info like 

`Not Specified   -- 25   5564 km/206 deg`

also fixing a minor typo on line 113 of `showinfo.c`

renamed `qrb_` and caller now uses its return value. unused int return dropped.


dl1jbe:
Applied and pushed. Thanks Zoli.

PR #126, opened 2019-04-30
switch back to CQ after working a grabbed station

zcsahok:
This is my best effort for "automatically switch back to 'Run' mode from "grab spot" after QSO".
While designing it I realized that it's impossible to implement this behavior as one has to wait at least until the end of the SP_TU message before switching back to CQing. It is actually even worse, as there could be still repeats happenning after this message (either exchange or callsign not received correctly). So one can switch to CQ only on operation action.
The solution is that after a completed grabbed S&P QSO pressing Esc reverts to CQ frequency. It is not too much effort from the op and gives the freedom of finishing the QSO correctly. Esc is normally next to F1, so no big movement is involved.
In addition, now `#`  restores the operating mode also (CQ/S&P). I think this makes sense.

Changes:
- main.c : `mem_cqmode` and `mem_grab` hold info on the mode at the time of storing mem and if it was due to a grab.
- logit.c: `change_mode()` moved to callinput.c; a bit restructured the logit() mega-function, reducing the indent depth for readability.
- callinput.c: added helper functions for mode change and VFO-MEM operations. They still mix logic with display code, this needs to be fixed when UI is refactored. update_mode() should be called instead of setting mode string in other places.
- grabspot.c: store cqmode and grab info into MEM along with the freq.

A note regarding VFO-MEM: now it's in fact a stack with `#` pushing (VFO->MEM) or popping (MEM->VFO) depending whether the stack if empty or not. It may also make sense to add an exchange operation VFO<->MEM. This would allow switching back and forth between CQ and S&P freqs without losing them. We could use `$` key for it, as it's just next to `#`.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -82,6 +82,7 @@ void send_bandswitch(freq_t freq);
 int autosend(void);
 int plain_number(char *str);
 void handle_bandswitch(int direction);
+void revert_grab();
 
 extern int no_arrows;
 extern char hiscall[];
@@ -92,6 +93,13 @@ extern int dxped;
 extern freq_t freq;
 extern int trx_control;
 extern freq_t bandfrequency[];
+extern int cqmode;
+extern char mode[];
+extern freq_t freq;
+extern freq_t mem;
+extern int mem_cqmode;
+extern int mem_grab;
+
 
 
 /** callsign input loop
@@ -106,9 +114,7 @@ int callinput(void) {
     extern int early_started;
     extern char hiscall_sent[];
     extern char comment[];
-    extern int cqmode;
     extern int trxmode;
-    extern char mode[];
     extern char lastcall[];
     extern int cqdelay;
     extern char his_rst[];
@@ -116,7 +122,6 @@ int callinput(void) {
     extern int cluster;
     extern int announcefilter;
     extern char ph_message[14][80];
-    extern freq_t mem;
     extern SCREEN *mainscreen;
     extern int simulator;
     extern int simulator_mode;
@@ -253,23 +258,9 @@ int callinput(void) {
 
 	/* special handling of some keycodes if call field is empty */
 	if (*hiscall == '\0') {
-	    if ((x == '+') && (*hiscall == '\0') && (ctcomp == 0)) {
+	    if (x == '+' && *hiscall == '\0' && !ctcomp) {
 		/* switch to other mode */
-		if (cqmode == CQ) {
-		    cqmode = S_P;
-		} else
-		    cqmode = CQ;
-
-		/* and show new mode */
-		attron(COLOR_PAIR(C_HEADER) | A_STANDOUT);
-
-		if (cqmode == CQ) {
-		    mvprintw(0, 2, "Log     ");
-		    strcpy(mode, "Log     ");
-		} else {
-		    mvprintw(0, 2, "S&P     ");
-		    strcpy(mode, "S&P     ");
-		}
+                change_mode();
 		cleanup();
 
 	    }
@@ -515,9 +506,8 @@ int callinput(void) {
 
 	    // <Insert>, send exchange in CT mode
 	    case KEY_IC: {
-		if (ctcomp != 0) {
+		if (ctcomp) {
 		    send_standard_message(1);		// F2
-
 		}
 		break;
 	    }
@@ -544,20 +534,7 @@ int callinput(void) {
 
 	    // Hash, save xcvr freq to mem or restore mem to xcvr.
 	    case '#': {
-		if (mem == 0.0) {
-		    mem = freq;
-		    mvprintw(14, 68, "MEM: %7.1f", mem/1000.);
-		} else {
-		    freq = mem;
-
-		    set_outfreq(mem);
-
-		    mem = 0.0;
-		    mvprintw(14, 68, "            ");
-		}
-		mvprintw(29, 12, " ");
-		mvprintw(29, 12, "");
-		refreshp();
+                swap_mem();
 		break;
 	    }
 
@@ -836,6 +813,7 @@ int callinput(void) {
 		if (early_started == 0) {
 		    /* if CW not started early drop call and start anew */
 		    cleanup();
+                    revert_grab();
 		    clear_display();
 		} else {
 		    /* otherwise just stop sending */
@@ -934,7 +912,7 @@ int callinput(void) {
 
 	    // Alt-g (M-g), grab first spot matching call field chars.
 	    case 231: {
-		double f = grabspot();
+		freq_t f = grabspot();
 		if (f > 0.0) {
 		    grab.state = IN_PROGRESS;
 		    grab.spotfreq = f;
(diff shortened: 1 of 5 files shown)


airween:
This was a missing feature, thanks! :)

dl1jbe:
> This is my best effort for "automatically switch back to 'Run' mode from "grab spot" after QSO".

Good work. Thanks for bringing it forward.

> While designing it I realized that it's impossible to implement this behavior as one has to wait at least until the end of the SP_TU message before switching back to CQing. It is actually even worse, as there could be still repeats happenning after this message (either exchange or callsign not received correctly). So one can switch to CQ only on operation action.

Right. An automatic switch back does not really work. We need to do it by hand.

> The solution is that after a completed grabbed S&P QSO pressing Esc reverts to CQ frequency. It is not too much effort from the op and gives the freedom of finishing the QSO correctly. Esc is normally next to F1, so no big movement is involved.

Hmm, just tested it. Works as expected but it seems ESC is not really needed. With your modifications you can also switch back by '#'.
 
Some additional thoughts from my side are:

>     * main.c : `mem_cqmode` and `mem_grab` hold info on the mode at the time of storing mem and if it was due to a grab.

How about presenting  the related information in a struct? 

>     * callinput.c: added helper functions for mode change and VFO-MEM operations. They still mix logic with display code, this needs to be fixed when UI is refactored. update_mode() should be called instead of setting mode string in other places.

Fine. As a side comment swap_mem() may not need the refreshp(). That is always called at the end of the key input loop. So it may be unnecessary here (Btw, same arguments holds for the original version) .
> 
>     * grabspot.c: store cqmode and grab info into MEM along with the freq.
>
I am not sure if we need some way to nil out the 'grabbed' state by hand. I can think of an operator who did a grab but decided to stay on S&P afterwards. A first idea may be to drop the 'grabbed' state every time you switch cqmode by hand ('+'). What do you think?  

There is one further question I would like to discuss but I think it goes out of range of the discussion here. so please look for a private mail in next hours.

zcsahok:
Thanks for the feedback. I have to still work a bit on this as using `Esc` to implicitly switch back could cause problems. Need to test it.
(the advantage of `Esc` is that it's easier to press than `#`, no shift is involved)
Also need to review a bit the possible use cases.

zcsahok:
Here are some thoughts about the MEM topic. Let me know what you think.

General
----------
MEM is a 1-level stack that stores mode (CQ/S&P) along the frequency.
The idea is to add operations that help restoring a CQ frequency and
also ones to make MEM manipulation complete.

Keys and operations
----------
 (* = new/changed)
[#] - push/pop MEM: MEM->VFO and clear MEM (pop, if MEM is not empty) or VFO->MEM (push, else)
*[\`] - "swap": MEM <-> VFO (if MEM is not empty) or VFO->MEM (else)
     _rationale_: unlike [#] this one keeps MEM
*[\~] - forget MEM (swap+drop in Forth terms) (Shift-\`)
        _rationale_: there is no other way to clear MEM without
            losing current frequency information.
*[F1] - "auto revert": in S&P mode if call field is empty and MEM has a CQ freq
            (not neccessarily due to a grab) then revert to it (i.e. back to CQ mode). Executes a swap ([\`]) to store current S&P freq.
        _rationale_: in S&P mode call field is empty after a QSO and
            pressing F1 can signal in this particular case that one is
            ready to revert to CQ. In S&P there is no reason to press F1
            if callsign hasn't been entered yet.
*[@] - "forced push": overwrite MEM, shortcut for [~]+[#]

Use cases
----------
**In CQ mode:**

"switch to S&P without grab"

    - optional: store freq in MEM [#]/[@] to have return option
    - [+]
    - tune around and work stations
    - optionally return to CQ [`]/[#] or forget it [~]

"switch to S&P where left before"

    - previous S&P freq is in MEM
    - swap [`]
    - tune around and work stations
    - return to CQ [`]/[F1], MEM keeps current S&P freq

"grab and return"

    - grab [Alt-G] ([Ctrl-G] makes no sense)
    - MEM has the CQ freq - grab executes [@] (forced push)
    - work station
    - optionally "grab station/next"
    - return with [F1]/[`] or [#]

"grab and leave CQ freq"

    - grab [Alt-G]
    - MEM has the CQ freq  - grab executes [@] (forced push)
    - work station
    - look for other stations/CQ freq
    - forget old CQ freq [~]

"switch to S&P"

    - optionally store freq [#]/[@]
    - [+]

**In S&P mode:**

"grab station/next"

    - grab [Alt-G/Ctrl-G] - grab doesn't touch MEM in S&P mode
    - MEM not changed

"store freq"

    - option1: if MEM not empty then clean it with [~]
    - push/pop [#]
    - option2: use [@] instead of [#]
    - tune/grab around
    - return with [`]/[#] (or forget with [~])

"switch to CQ"

    - optionally store freq [#]/[@]
    - [+]

dl1jbe:
I need to think a little bit more about the use cases.

But some things to think about as a starter:

- There are different keyboard layouts (especially for different languages). What is easy to press on one board may be different on others (e.h. '#' is quite easy on german layout - just beside Enter).

- I am not quite sure if [`] is a good choice. There are 3 nearly common ones [`], [´] and ['] - I will surely mix them all the time (and others will do also).

- We should concentrate on making the most used cases simple and easy (one keypress). Other seldom used cases may require more than one key press.

So, now to think about what are the main cases....

dl1jbe:
I think the most used and needed features for memory manipulation are [#] push/pop and [@] forced push.
With that you can empty the memory with [@][#] (save QRG and immediately return back to that point). 

With these two keys (and F1 for return) we could handle most (but one) of your use cases. The only feature missing would be a 'swap'. 

As a side idea: If only one memory may not be enough we could precede [#] and [@] by a single number (0..9) in the otherwise empty call input field. The main problem would be to get info which QRG is in which memory.

Looking into the use cases I think the ones most used are the followings:

When running contest in CQ mode you would like to:
* Grab one/more station/s, work them in S&P  and return to CQ and your old frequency.
* Remember your running frequency and try a different one. Come back if not successful.
* Switch to S&P with/without remembering QRG

When hunting stations in S&P mode you would like to:
* Remember a QRG for a station with a lot of pile up at the moment. Come back later to try again. Could also be done with spotting to bandmap (Ctrl-a). More than one memory could be useful.

dh5fs:
I support the idea of connecting the "#" frequency with RUN. In most cases this was the RUN frequency and you want to continue CQ after grabbing a Stn in S&P. This saves pushing "+".
I support also the idea of "@" changing to RUN (=MEM frequency) changing to CQ mode and start F1 right away.
73 Fred

zcsahok:
This seemed to be an easy task but at the end we have identified a number of issues.

1. Current MEM storage needs to enhanced with swap (VFO<->MEM) and forced store (VFO->MEM) functions. These solve the issue of inability to overwrite MEM and to check MEM freq without losing VFO setting. Note that the new functions are just the well-established A-B VFO operations A/B and A=B.

2. In addition to the working frequency MEM should also store (at least)
- CQ/S&P mode
- call sign input field
The latter is to able to quickly check a S&P without the need of retyping the call.

3. Switch back to CQ shall be added, either as an extended `F1` functionality or a separate key. (or potentially both)

4. Key remapping shall be supported. This is to allow optimizing for a specific keyboard layout and user.
E.g. `+` and `#` are "easy" keys on German keyboard (near Enter, no shift), whereas on US keyboard they're shifted.

My plan is to drop this PR and rework the code to add 1) and 2). For keys my suggestion is `$` for VFO->MEM and `%` for VFO<->MEM. (one can remap them once 4) is done anyway)
It this works then I'll do 3).
During this period we can think about 4) and come up with an design.

zcsahok:
closing, replaced by #155

PR #127, opened 2019-05-14
Vertical resize

dl1jbe:
Enables vertical resizing of TLFs display.

You can resize the terminal before the start of TLF and also during it is working. Resize should work in the main panel and in the packet window. Resize during display of multis, propagation infos and during scroll in packet window results simple in a stop of that display and return to the normal main or packet display.

Vertical resize should be working down to 23 lines without problems. Making it smaller will result in a distorted display. 

The code could need some testing before merge. Please report back. Thanks.

### DIFF
--- a/src/bandmap.c
+++ b/src/bandmap.c
@@ -48,6 +48,14 @@
 #define DISTANCE(x, y) \
     ( x < y ? y - x : x -y )
 
+extern int ymax;
+#define TOPLINE 14
+#define LASTLINE (ymax - 2)
+
+#define LINELENGTH 80
+#define COLUMNS ((LINELENGTH - 14) / SPOT_COLUMN_WIDTH)
+#define NR_SPOTS ((LASTLINE - TOPLINE + 1) * COLUMNS)
+
 pthread_mutex_t bm_mutex = PTHREAD_MUTEX_INITIALIZER;
 
 /** \brief sorted list of all recent DX spots
@@ -69,6 +77,7 @@ bm_config_t bm_config = {
 };
 short	bm_initialized = 0;
 
+extern int ymax;
 extern freq_t freq;
 extern int trx_control;
 extern int bandinx;
@@ -488,7 +497,7 @@ void bm_show_info() {
 
     /* show info field on the right */
     attrset(COLOR_PAIR(CB_DUPE) | A_BOLD);
-    move(14, 66);
+    move(TOPLINE, 66);
     vline(ACS_VLINE, 10);
 
     mvprintw(18, 68, "bands: %s", bm_config.allband ? "all" : "own");
@@ -600,8 +609,8 @@ void show_spot_on_qrg(spot *data) {
  */
 void next_spot_position(int *y, int *x) {
     *y += 1;
-    if (*y == 24) {
-	*y = 14;
+    if (*y == LASTLINE + 1) {
+	*y = TOPLINE;
 	*x += SPOT_COLUMN_WIDTH;
     }
 }
@@ -727,8 +736,8 @@ void bandmap_show() {
 
     getyx(stdscr, cury, curx);		/* remember cursor */
 
-    /* start in line 14, column 0 */
-    bm_y = 14;
+    /* start in TOPLINE, column 0 */
+    bm_y = TOPLINE;
     bm_x = 0;
 
     /* clear space for bandmap */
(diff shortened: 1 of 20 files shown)


dl1jbe:
Be aware that Github can not rebase the pull request. Nevertheless it can be merged or checked out separately for test. 

I will rework the pull request to fix it in next days.

zcsahok:
Got a Seg Fault upon resizing. Not sure what the root cause is (using xfce4-terminal from Debian buster).
Backtrace:

```
#0  0x00007ffff7cd659a in wtouchln ()
   from /lib/x86_64-linux-gnu/libncursesw.so.6
#1  0x00007ffff7cf5e3c in replace_panel ()
   from /usr/lib/x86_64-linux-gnu/libpanelw.so.6
#2  0x00005555555896c3 in refresh_splitlayout () at splitscreen.c:573
#3  0x000055555558bd1f in resize_layout () at ui_utils.c:339      
#4  0x000055555558bdc8 in getkey (wait=wait@entry=0) at ui_utils.c:149
#5  0x000055555558c017 in key_poll () at ui_utils.c:132
#6  0x0000555555561ba5 in callinput () at callinput.c:250
```

dl1jbe:
Thanks for the report. Can you give just some info to the situation it happened?

- Are you in packet window or main window?
- Did it happen on first try to resize or sporadically?

If you were in main window, you can try to comment out the call to refresh_splitlayout() in ui_utils.c:573.
Than resize should work for all but the packet window.

I think I know what happened but do not understand why yet. I will try to prepare some test code and send it to you by mail.

zcsahok:
It is on main screen, the only user action is the resize.
Figured it out: it couldn't connect to the cluster server and then `main.c` it has released all structs

```
        if (init_packet() == 0)
            packet();
        else
            cleanup_telnet();
```

This resulted in a bogus internal _maxy leading to a segv.

So `refresh_splitlayout` shall just check `initialized`.

zcsahok:
When not using rig control then the place of TRX/MEM info is not cleared and can contain garbage after extreme resizing. A solution is to add clearing that region in `show_freq` in the else branch:

```
if (trx_control == 1) {
...
    } else {
        attron(modify_attr(COLOR_PAIR(C_LOG)));
        mvprintw(13, 67, "             ");
        mvprintw(14, 67, "             ");
    }
```

(`attron` can be pulled in front of `if`)

dl1jbe:
> It is on main screen, the only user action is the resize.
> Figured it out: it couldn't connect to the cluster server and then `main.c` has released all structs

Right. That is the problem. I just missed to test this case. 

> This resulted in a bogus internal _maxy leading to a segv.

No quite. The replace_panel() tries a redraw of the old window to be replaced (packet_window), but that got freed during 'cleanup_telnet().

> So `refresh_splitlayout` shall just check `initialized`.

There will be a fixup commit shortly.

I just tested the case here in meantime. Interestingly I get no segfault. So I can not test the fix.

Good to have different people with different system configuration for tests.

dl1jbe:
> Could the three variables related to vertical terminal size (`ymax`, `LINES`, `YPOS`) be handled somehow unified? I know that `LINES` comes from curses, but will it be updated automatically? All caps suggests a constant.
> 
Yes, that part needs to be cleaned up.
LINES will be set automatically during resize by ncurses themself. So we will used that.

The all caps can not be changed as LINES got introduced by old curses many years ago and still stick around.

> Shouldn't `getkey()` return `0` in case of `KEY_RESIZE`?

I though about that but decided to return the KEY_RESIZE to the application in case some code would need to act accordingly. It is already useful for display of mults or propagation window as they wait for a keypress to vanish. So we need not implement a resize for that displays. They just go away during resize and will be made anew in proper size during next call.

dl1jbe:
> When not using rig control then the place of TRX/MEM info is not cleared and can contain garbage after extreme resizing. 

Also a good catch which needs to be fixed. I wiped out remnants from resizing but only for the case when you have rig control.

zcsahok:
Segfault is gone now, all OK. 
I like the bigger spot display.

`refresh_splitlayout` seems to work also with `wresize`:

```
     wresize(packet_win, LINES, COLS);
     replace_panel(packet_panel, packet_win);
```

this way no `delwin(oldwin)` is needed.

dl1jbe:
> Segfault is gone now, all OK.
> I like the bigger spot display.

Fine :-)

> `refresh_splitlayout` seems to work also with `wresize`:
> ```
>      wresize(packet_win, LINES, COLS);
>      replace_panel(packet_panel, packet_win);
> ```
> 
> this way no `delwin(oldwin)` is needed.

That will need some more tests. Problem is the handling of the subwindows. I tried it with the way you suggests already but got problems in some cases. So I switched to the newwin/delwin mechanism we have now to be sure.

dl1jbe:
I prepared a reworked set of commits in branch 'vertical_resize' with the following changes:

- drop use of ymax and YPOS for positioning and relying only on LINES.
- resort and squash related commits into single ones 

I will prepare a new pull request to replace these one.

PR #129, opened 2019-06-14
Do not highlight exch field on CQ<->S&P

zcsahok:
Upon CQ<->S&P switch (`+` key) exchange field gets unnecessarily highlighted. 
Before switch:
Screenshot - 06142019 - 10:27:34 PM
After switch:
Screenshot - 06142019 - 10:28:21 PM

This fix removes exch highlighting.

_A side note:_ CQ<->S&P switch is only possible if callsign field is empty. This can cause a problem if one forgets to switch to CQ/S&P and realizes it only after having entered a call. Then one has to remember the already typed call, delete it, do the switch and re-type (correctly!) the call.
I see no reason why a switch couldn't be allowed even if callsign field is non-empty.

### DIFF
--- a/src/cleanup.c
+++ b/src/cleanup.c
@@ -48,8 +48,10 @@ int cleanup(void) {
     attron(modify_attr(COLOR_PAIR(NORMCOLOR)));
 
     mvprintw(12, 29, "            ");
-    mvprintw(12, 54, "                        ");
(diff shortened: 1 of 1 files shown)


dl1jbe:
Applied. Thanks for it.
I added your side note  as Issue #130 to not loose it out of focus.

PR #131, opened 2019-06-17
Allow CQ<->S&P switch even if call input field is not empty

### LINKED ISSUE #130: Allow CQ<->S&P switch even if call input is not empty
A side note: CQ<->S&P switch is only possible if callsign field is empty. This can cause a problem if one forgets to switch to CQ/S&P and realizes it only after having entered a call. Then one has to remember the already typed call, delete it, do the switch and re-type (correctly!) the call.
I see no reason why a switch couldn't be allowed even if callsign field is non-empty.

Reported by ZCsahok in PR#129

zcsahok:
fixes #130

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -253,27 +253,6 @@ int callinput(void) {
 
 	/* special handling of some keycodes if call field is empty */
 	if (*hiscall == '\0') {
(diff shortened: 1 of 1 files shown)


dl1jbe:
Thanks. Applied and pushed

PR #132, opened 2019-06-23
Fix frequency adjustment function (Ctrl-F)

zcsahok:
The issue with Ctrl-F is that as it waits for a key press between the updates the 3  frequency values (shown in big digits, the normal TRX display, and the actual one on the rig) can get quite out of sync. I was able to produce a state when all 3 were different by pressing `Up` twice in short succession. This can be rather annoying when one is operating remotely, without actually seeing the rig. Of course after leaving frequency adjustment things sync up but the resulting frequency can be different from the one "set" in Ctrl-F.

The main fix is to replace `key_get()` with `key_poll()`. In order to avoid flickering the order of `freq_display()` and `time_update()` was reversed.
Minor code cleanup was done in `freq_display.c`.

Ctrl-F is now completely live in the sense that it also takes over freq changes from the rig. The cluster display below Ctrl-F is also updated and the displayed values are all in sync. For TRX there is a delay of 1 sec but that's tolerable as it's the normal behavior.

### DIFF
--- a/src/changefreq.c
+++ b/src/changefreq.c
@@ -33,7 +33,6 @@ void change_freq(void) {
     extern int trx_control;
 
     int brkflg = 0;
-    int x;
 
     if (trx_control == 0)
 	return;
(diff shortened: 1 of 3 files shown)


dl1jbe:
Thanks. Applied and pushed.

PR #133, opened 2019-07-06
Improve command line argument handling

zcsahok:
Switched to **Argp** as it provides a number of features (usage, version, error messages, long arguments, ...) out of the box.

The original issue was that argument option had to be written without space (`-fmyfile.dat`). Now both this, the "normal' format (`-f myfile.dat`), and the long format (`--config=myfile.dat`) are also supported. 

I took over all existing options, even though I'm not sure if `-s` is actually functional. This is to be checked.

As there were no established long argument names I made them up.

Also rationalized the welcome text generation.

The formatting changes were done by `astyle`.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <termios.h>
 #include <unistd.h>
+#include <argp.h>
 
 #include "addmult.h"
 #include "background_process.h"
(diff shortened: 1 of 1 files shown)


zcsahok:
(argp also does tilde substitution)

dl1jbe:
> (argp also does tilde substitution)
> (well, maybe it's the shell, not sure)

Yes, tilde expansion is done by the shell.

Applied and pushed. Thanks.

Btw, I got some warnings from pc-lint about mixing storage class identifiers (static) inside of type identifiers (const int). See also  https://stackoverflow.com/questions/12005959/static-const-versus-const-static

PR #134, opened 2019-07-07
More readable long option names

zcsahok:
Hamlib (e.g. `rigctl`) and other tools use this style. I find it more readable.

Also changed `static`.


dl1jbe:
Thanks. Applied and pushed

Should be no criticism. Just stumbled about pc-lints report and wondered where it came from.

PR #135, opened 2019-09-15
New startup

dl1jbe:
New startup code for better readability by user.

*  'verbose' startup mode does no longer slow down display of messages but requires a
  key press before clearing the screen. That gives a better chance to read all the messages 
  and sort out possible problems.

* If started in empty directory (new contest or first start by new user) it switches 
  automatically to verbose mode.

### DIFF
--- a/src/checkparameters.c
+++ b/src/checkparameters.c
@@ -29,17 +29,11 @@
 
 
 int checkparameters(void) {
-
-    extern int emptydir;
-
     FILE *fp;
 
     if ((fp = fopen(".paras", "r")) == NULL) {
 	writeparas_file();
-	emptydir = 1;
-
     } else {
-
 	fclose(fp);
     }
 
--- a/src/getmessages.c
+++ b/src/getmessages.c
@@ -37,6 +37,7 @@
 #include "ignore_unused.h"
 #include "tlf_curses.h"
 
+
 /* get countrynumber, QTH, CQ zone and continent for myself */
 void getstationinfo() {
     extern char call[];
@@ -57,7 +58,8 @@ void getstationinfo() {
     QTH_Long = mydx->lon;
 }
 
-int getmessages(void) {
+
+void getmessages(void) {
 
     extern char call[];
     extern char mycqzone[];
@@ -66,7 +68,6 @@ int getmessages(void) {
     extern int qsonum;
     extern char qsonrstr[];
     extern char backgrnd_str[];
-    extern int emptydir;
 
     FILE *fp;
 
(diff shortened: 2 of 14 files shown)


zcsahok:
Looks good. Maybe capitalization could be unified (reading / Opening).
The "Phone message" is still at a fixed place. If there are multiple such messages only the last would be visible -> makes not too much sense. It's probably better to put them on a new line each and show only in verbose mode. (what is "A:"?? it's in the code but I have no idea for what it stands)

```
        Welcome to tlf-1.4~git by PA0R!!

reading country data
reading configuration data
Opening config file logcfg.dat
reading contest rules file: rules/ha5cqz
Using contest rules file:  rules/ha5cqz
reading callmaster data
Rignumber is 2
Rig speed is 2400
Trying to start rig ctrl
rig_open: error 
Continue without rig control Y/(N)?


     A: Phone message #0 is cq.wav
```

dl1jbe:
> Looks good. Maybe capitalization could be unified (reading / Opening).

Makes sense. See new commit.

> The "Phone message" is still at a fixed place. 

Forgot it. Thanks for reminding. There were some more similar problems for QTC related files.

> If there are multiple such messages only the last would be visible -> makes not too much sense. It's probably better to put them on a new line each and show only in verbose mode. 

Done.

> (what is "A:"?? it's in the code but I have no idea for what it stands)

Good question. Maybe left over from Reins old code. Dropped it.

zcsahok:
Nice!
Looking at the messages I was wondering why it says Reading and then Using. Checking the code shows that Using means simply the end of parsing. IMO that message could be dropped as it is anyway clear, especially if there were no parsing errors.
The while loops reading the configs (2 in rules.c, 1 in parse_logcfg.c) could be also be put into a common function. They must be the result of previous copy-pasting.
I'd also change Opening to Reading for consistency.

```
Opening config file logcfg.dat
  Phone message #0 is cq.wav
Reading contest rules file: rules/ha5cqz
Using contest rules file:  rules/ha5cqz
```

dl1jbe:
> Nice!
> Looking at the messages I was wondering why it says Reading and then Using. Checking the code shows that Using means simply the end of parsing. IMO that message could be dropped as it is anyway clear, especially if there were no parsing errors.
> I'd also change Opening to Reading for consistency.

Done. Good to have a second set of eyes to look for such inconsistencies.

> The while loops reading the configs (2 in rules.c, 1 in parse_logcfg.c) could be also be put into a common function. They must be the result of previous copy-pasting.

Good idea. Done.

PR #136, opened 2019-09-18
fix sideband selection

zcsahok:
An unnoticed constant caused wrong sideband selection when switching to SSB.
Moved SSB and CW settings to functions.


dl1jbe:
Applied and pushed. Thanks.

PR #139, opened 2019-11-18
Fix keyer clearing exchange field

zcsahok:
It started as a quick fix for keyer clearing exchange field which could be quite annoying when one needs that info for e.g. asking confirmation.
Example:
image
after pressing Ctrl-K 123 is gone. At least it is restored once keyer window is closed.
image

The solution seemed to be simply removing `clear_display()` from `keyer.c`. But it caused further issues so at the end the update of the header row was separated into `update_header_line()`. This allowed getting rid of `mode` and its settings scattered all around the code. Introduced `cqmode_t` for mode values.

Bonus: search window doesn't overlap keyer widow after Fx keys or PgUp/PgDn/Backspace.

As already touched, `autocq.c` has been refactored.

Introduced `keyer_append_char()` to avoid locally fiddling with 2-byte char arrays.

A minor fix in `write_keyer()`: if called with an empty string and using `digikeyer == GMFSK` then an array access at index -1 would happen. (both conditions are quite unlikely...)   Added an early exit if there is nothing to send.

Changed come "cryptic" constants as in `case 125:` to characters and switched to `toupper()` instead of the unreadable `if`.

Astyle formatting changed some comments. Perhaps the comments should be also improved.

### DIFF
--- a/src/autocq.c
+++ b/src/autocq.c
@@ -23,8 +23,8 @@
  *--------------------------------------------------------------*/
 
 
-#include <string.h>
 #include <unistd.h>
+#include <ctype.h>
 
 #include "clear_display.h"
 #include "cw_utils.h"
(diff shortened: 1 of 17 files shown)


zcsahok:
An unrelated fix: when compiling with gcc 8.3.0 one gets a pile of warnings like 
```
sendbuf.c: In function ‘ExpandMacro’:
sendbuf.c:205:5: warning: ‘strncpy’ specified bound 4 equals destination size [-Wstringop-truncation]
     strncpy(rst_out, his_rst, 4);
```

IMO we shall fix them (also the non-strncat related ones), as an example I cured this one: in this case it's easier to copy the 3 bytes than call strncpy.

dl1jbe:
Good work. Applied and pushed. Thanks.

Wrt gcc warnings: There are even more in GCC-9. I am in the process of fixing now.

PR #140, opened 2019-11-21
Fix warnings

dl1jbe:
Quell warnings due to stricter check with GCC-8 or GCC-9 compiler.

Most problems were related to buffers to small or size of operands not correctly specified.
Most are easy ones but some are a little bit tricky.

@airween please test the  changes in the WAE related code (genqtclist and qtwin). It is quite complex and I am not sure if I did not break something. So please test the expected behavior especially for the places I did change.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -26,6 +26,7 @@
 
 #include <dirent.h>
 #include <errno.h>
+#include <glib.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
@@ -214,7 +215,7 @@ void record(void) {
 				i = 10;
 				j++;
 			    }
-			    strncpy(printname, soundfilename->d_name, 6);
+			    g_strlcpy(printname, soundfilename->d_name, 7);
 			    mvprintw(j, i, "%s", printname);
 			    refreshp();
 
--- a/src/clusterinfo.c
+++ b/src/clusterinfo.c
@@ -189,10 +189,10 @@ int loadbandmap(void) {
     char spotcall[20];
     char spottime[6];
     char spotline[38];
-    char callcopy[81];
+    char callcopy[20];
     FILE *fp;
     char marker_out[60];
-    char color[20];
+    char color[sizeof("Magenta")];
     int lon;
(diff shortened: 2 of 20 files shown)


airween:
@dl1jbe right - I need few days now to review.

dl1jbe:
Thanks. qsonr_to_str() is quite nice now.

> or are there prefixes longer than 3 chars?

Yes there are some of xy/z..  type. At the moment we show only the first 3 characters and an space as separator (Try cqww and Alt-M to show mults).

But maybe `sprintf(prefix, "%-3s ", dxcc_by_index(i)->pfx);` will do.

zcsahok:
this should work (added precision specifier)
```
sprintf(prefix, "%-3.3s ", dxcc_by_index(i)->pfx);
```

dl1jbe:
> But maybe `sprintf(prefix, "%-3s ", dxcc_by_index(i)->pfx);` will do.

No, it does not. man page states that not more than n char, are written by %ns but practice shows otherwise. I will keep the old code here.

airween:
I've compiled and checked the master branch. Additional info's:

```
gcc version 9.2.1 20191109 (Debian 9.2.1-19) 
```

```
Tlf Version 1.5~git configuration:

 Prefix         /usr/local
 Preprocessor   gcc -E 
 C Compiler     gcc -g -O2

 Package features:

 With XML RPC	true
```

Made two directories to emulate the WAE-EU and WAE-DX contests. Logged few QSO's, everything worked. Received and sent few QTC's, exits the program and restart it again in both cases. The QTC list had remained.

There were some warnings until the copile, if I'll have more time, I'll check them.

Hope this helps.

dl1jbe:
Thanks, Ervin, for checking. Will merge the PR now.

airween:
> Thanks, Ervin, for checking. Will merge the PR now.

Sorry, I've tested the master branch yesterday, but now I merged your modification to my local repository, and re-ran the tests - everything works as well.

dl1jbe:
Applied and pushed in commit 5aa04a237.

PR #141, opened 2019-11-27
New editor

dl1jbe:
* Allow use of arbitrary editor commands 
* If not defined fall back to editor specified in EDITOR environment varaible
* Be aware that you may have to 'rescore' after editing with an GUI based editor.


dl1jbe:
Thanks for both ideas. Pushed in the new commit.

PR #142, opened 2019-11-30
Fix partial display overflow and call selection

zcsahok:
This fix targets issues with the display of partial calls and the USEPARTIALS functionality.

- when calls from the log alone fill the partial display then it still takes one match from callmaster and overflows into the next line.
image

- if there are no matches from the log, USEPARTIALS takes the first one from callmaster, even if it's not unique. Here I entered "OO" and got a 2E call:
image
Note also that while it shows in the Worked panel Belgium correctly, no Belgian call is displayed.

- calls are shown multiple times if they are present multiple times (e.g. different band, mode, or other)
image

First I commited the updated tests that cover these situations.


zcsahok:
The common partial diplay functionality has been moved to `show_partial()`. It filters out duplicate calls, updates screen position and collect the suggestion information. It uses constants for the display size and location to ease eventual resizing later on.

The actual search loops has been simplified, attr-setting code and other loop invariants moved outside.

Callmaster search makes 2 runs
- first it looks for calls starting with the entered one
- then searches the ones containing it (this was the original functionality).

This way preference is given to the prefix search, e.g. the same search for "OO" looks like
image
Makes more sense, IMO.

Suggested partial is used only if a unique match was found.

`displaySearchResults` was just formatted by astyle, but it points to a code complexity issue and also a functional one that I didn't notice: unlike on line 377 here DIGIMODE is not checked.

I'll also add a bit more normalizing to callmaster file reading: upcasing and filtering duplicates. This w
could help in case one edits the file manually.

zcsahok:
(strange, there used to be a Travis build check for the PR. now it's missing)

src/searchlog.c
```diff
@@ -150,23 +148,63 @@ void displayCallInfo(dxcc_data *dx, int z, char *pxstr) {
...
+//
+static int show_partial(int *row, int *col, char *call,
+			GHashTable *callset,
+			int *nr_suggested, char *suggested_call) {
+
+    if (!g_hash_table_add(callset, call)) {
+	return 0;   // already shown
+    }
+
+    mvprintw(PARTIALS_X0 + *row, PARTIALS_Y0 + *col, "%s ", call);
```
dl1jbe: Please recheck. X normally is in column direction, Y are row positions.

src/searchlog.c
```diff
@@ -182,75 +220,78 @@ int displayPartials(char *suggested_call) {
...
-	/* and now check callmaster database */
-	for (k = 0; k < callmaster->len; k++) {
+    }
 
-	    if (strstr(CALLMASTERARRAY(k), hiscall) != NULL) {
+    attron(COLOR_PAIR(C_LOG) | A_STANDOUT);
+    attroff(A_BOLD);
 
-		attron(modify_attr(COLOR_PAIR(C_LOG) | A_STANDOUT));
+    // no callmaster check if exact match was found in the log
```
dl1jbe: What happens if call found in log is a substring of one in callmaster (or log), e.g. K3A and K3AL?

src/searchlog.c
zcsahok: corrected

src/searchlog.c
zcsahok: I tried to keep search functionality unchanged, so let's see.
Is this the setup?
- in log: K3A
- in callmaster: K3AL
- entered: K3A

In this case current code gives only K3A and as dupe. Same with the updated code.

Do you mean it should also list K3AL as an option?

dl1jbe: Yes, but it should work in any combination of log and callmaster. Maybe you already worked K3A. As you use the partial function to look up calls which may fit to that partial string K3AL from callmaster is a valid possibility.

It seems that the old implementation is wrong here.

zcsahok: In the current code this line prevented further search
```
    if (strcmp(hiscall, printres) != 0) {

        /* and now check callmaster database */
```

We could simply ignore this check and do all searches even if call was in the log already.
Makes sense. The display would be then like this (K3AL comes first as I added it extra, so my callmaster file is not sorted)
image


In this case USEPARTIALS wouldn't trigger, but that's OK as the call has been entered anyway.

I'll change the display code a bit to do a tighter checking that would allow K3ATO fit on the first line and thus showing a bit more results.

dl1jbe:
Looks really good now. Much more useful than before.

Any more changes? Otherwise I will pick up the PR.

zcsahok:
Still working on display and callmaster loading improvement. Trying to finish until tomorrow evening.

zcsahok:
Changed position update to check whether the actual call would still fit. This way a wider display is possible. Also changed trailing space to leading so that it's not shown for the last entry on the line.

Here are the previous examples for OK and OO:
image
image
Note that now 3 more calls are shown. (OO7R could be still fit to row 1)

Had to use 6 char calls in test to fill the display...

zcsahok:
This check looks quite nonsense to me: equality of the current and previous mode is required in **mixed** mode only?? (the comment is well-founded)
```
      if (((qsos[qso_index][3] == 'C' && trxmode == CWMODE) ||
               (qsos[qso_index][3] == 'S' && trxmode == SSBMODE) ||
               (qsos[qso_index][3] == 'D' && trxmode == DIGIMODE)) ||
               mixedmode == 0) {
           // ist letzterTest korrekt?
```

`mixedmode` is normally false so the check typically doesn't do anything.

Question: shall we drop it altogether (i.e. no mode checking) or fix it to select only QSOs with same mode?

dl1jbe:
As I read that sequence it should 
a) filter out stations we worked in just our actual mode or
b) show stations in any mode.

Only b) should happen in mixedmode. So the last line 'mixedmode == 0' seems to be wrong and must be inverted.

zcsahok:
Pushed mixmode update incl. tests.

Could we remove `cwkeyer.h` as it's not used anymore?

dl1jbe:
> Could we remove `cwkeyer.h` as it's not used anymore?

Just had a look. I think it can go.

zcsahok:
That was it.

dl1jbe:
Applied and pushed. Thanks.

PR #145, opened 2019-12-03
Fix ctmode

N0NB:
Hi Tom.

Here is a set of fixes for CT Compatible mode.


dl1jbe:
Thanks Nate. Applied and pushed.

PR #146, opened 2019-12-05
Enhance ctmode

N0NB:
Hi Tom.

These are a pair of keystroke enhancements for the CT mode.  I have other ideas but I'll bring them to the list first.

73, Nate


src/callinput.c
```diff
@@ -520,15 +520,30 @@ int callinput(void) {
...
+		/* When in ESM mode, semicolon opens the note dialog.
+		 * In all modes, Alt-N opens the note dialog.
+		 */
+		else if ((ctcomp == 0 && (x == ';' || x == 238))
+		          || ((ctcomp != 0) && (x == 238))) {
+		    include_note();
+		    x = -1;
 		}
 		break;
 	    }
```
dl1jbe: That makes the code much more complex and much less clear. Maintaining this woven set of conditions will become difficult.
I would recommend not to fold all that functions together into one case and then inside it split it again to different behavior. E.g. the Alt-N could stay out and have a case for itself even if that means to have two places which calls 'include_note()'. Maybe the same idea should be applied to the ';' key.

zcsahok: an alternative (IMO still readable):
```
#define ALT_N  238  // to avoid "magic" numbers

   if (x == ALT_N || (ctcomp && x == ';')) {
      ....
}
```

but splitting per-keycode is surely clearer

N0NB: Fair enough, Tom.  I'll revisit this, possibly after this coming weekend's ARRL 160m 'test, but maybe sooner as time permits.

The idea of not using magic numbers should have been resolved long ago as there are many, always preceded by comments, however, but something that should be done regardless.

N0NB:
Will refactor.

PR #147, opened 2019-12-15
New effort to use symbolic names for keystrokes

N0NB:
Many "magic numbers" exist for various keystrokes handled in the code,
mostly callinput.c and getexchange.c but elsewhere.  Even though these
are often commented, symbolic names are more readable.


N0NB:
Hi Tom.

This patch set converts the magic numbers for the control keys and a couple of others that I found.  Some may remain, of course.  Please look this over and I'll start working on the Alt keys next.

N0NB:
This completes my initial work of replacing magic numbers with symbolic name constants.  I'm sure more remain and I only changed a couple of character strings with constants.  There are places where the numbers make more sense than the names so I left those alone.  Anyway, this will hopefully improve readability of the code.

dl1jbe:
Looks good and cleans up the code quite a bit. Thanks for your effort. 

Applied and pushed.

PR #149, opened 2019-12-15
Rescore after delete of QSO 

dl1jbe:
Until now after deleting a QSO we try to correct the counters for qso numbers and multipliers. That works only partially. Not all changes can be undone that way.

The proposed patch does always a complete rescore of the log file after delete. So all counters are correct.

There are some further changes to simplify the code and sort out weird sequences (especially the handling of an empty log at the end of readcalls()). 

Additionally a new file with helper functions for decoding parts of log lines got added.

### DIFF
--- a/src/log_utils.c
+++ b/src/log_utils.c
@@ -1,0 +1,35 @@
+/*
+ * Tlf - contest logging program for amateur radio operators
+ * Copyright (C) 2019 Thomas Beierlein <dl1jbe@darc.de>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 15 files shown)


src/log_utils.c
```diff
@@ -0,0 +1,35 @@
...
+*    util functions to work with lines from log
+*
+---------------------------------------------------------------------------*/
+#include <string.h>
+#include <stdbool.h>
+
+/** check if logline is only a comment */
+bool log_is_comment(char *buffer) {
+
+    if (buffer[0] != ';' && strlen(buffer) > 60) /** \todo better check */
```
zcsahok: what could be the reason for the ">60"?  it was in the code earlier, just don't get the idea behind it.

dl1jbe: > what could be the reason for the ">60"? it was in the code earlier, just don't get the idea behind it.

Good question. I assume in earlier versions there were problems with short log lines sometimes. Had a look into old versions: It was already there in 0.9.10.

Until now I modified code very conservative - especially if it was not clear what the code was for.

My plan was to isolate all the small manipulations of logline into one file, write tests for it and afterwards clean up.

zcsahok: OK.  :+1: 
There is still a lot to improve, step by step.

dl1jbe:
Rearange the code,  rebase it onto actual master and pushed. Closing the PR.

PR #151, opened 2019-12-26
Added variable space generation

zcsahok:
While preparing the switch back feature I ran across a bunch of hard coded strings of spaces.
Added a `spaces(int)` function in order to get them more under control and to facilitate eventual horizontal resizing.

The spaces are served from a pre-allocated string of 1000 spaces. The returned value must be used for reading only.

Also voidified some functions (`addspot`, `log_to_disk`, `show_rtty`, `showscore`). Latter got an early exit to reduce somewhat its complexity.

`getmessages`: removed yet another log reading. This function has nothing to do neither with messages nor with `.paras` file as some comments suggest. That file is in fact write-only and used solely for first run recognition. getmessages deserves a renaming in the near future.

Found some functions that don't use their arguments (e.g. `int country_found(char prefix[])`), they'll be checked in separately.

### DIFF
--- a/src/addspot.c
+++ b/src/addspot.c
@@ -36,6 +36,7 @@
 #include "splitscreen.h"
 #include "tlf.h"
 #include "tlf_curses.h"
+#include "ui_utils.h"
 
 
 /** add call to list of spots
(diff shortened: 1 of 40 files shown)


dl1jbe:
Looks good. Thanks for the cleanup.

Most functions which do not use the arguments access some global variables instead.
We should check if it would be better to make them check the arguments and pass the global data via the argument. That would help in decoupling the code.

For country_found() it really looks like it can be dropped. Passed arguments (prefix) and checked data (hiscall) has nothing in common.

PR #152, opened 2019-12-28
Fix warnings and improve

zcsahok:
Fixed some warning produced by a compilation using `-Wextra`. Some parts of the the were improved (message_change) when already at it.

`_XOPEN_SOURCE` is guraded agains redefinition.

> clear_display.c: In function clear_display:
clear_display.c:83:17: warning: unused variable hiscall [-Wunused-variable]
     extern char hiscall[];

> messagechange.c: In function message_change:
messagechange.c:37:24: warning: unused parameter x [-Wunused-parameter]
 int message_change(int x) {

Here the code has been revisited. The two internal branches were doing essentially the same. Added option to escape without changes, both at message selection and later at message entry (`Esc` then `Enter`).  Dropped adding trailing space or NL: this should be done in the respective message sender (if needed at all). We should also remove trailing NL when reading messages from logcfg, but that's a longer story.

> searchlog.c:713:22: warning: unused parameter searchstring [-Wunused-parameter]
 void searchlog(char *searchstring) {

Again an example of function using global vars instead of argument. I agree that the real fix is to rewrite the function to do what was meant originally, but for the time being it's less effort to correct the signature.

> splitscreen.c: In function sanitize:
splitscreen.c:548:14: warning: comparison is always false due to limited range of data type [-Wtype-limits]
  else if (*s < 0)

`char` can be unsigned (e.g. on ARM), the check was fixed.

One thing that was *not* fixed: 

> score.c: In function country_found:
score.c:42:24: warning: unused parameter prefix [-Wunused-parameter]
 int country_found(char prefix[]) {

This guy gets called with various arguments from `exist_in_country_list`. As it doesn't actually use its arg I suspect here a functional issue. The code is somewhat complex, I couldn't come up with a test case for this.

### DIFF
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -26,7 +26,9 @@
 #define _GNU_SOURCE
 #endif
 
+#ifndef _XOPEN_SOURCE
 #define _XOPEN_SOURCE
+#endif
 
 #include <stdlib.h>
 #include <string.h>
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -113,7 +113,7 @@ void calledit(void) {
 	    showinfo(getctydata_pfx(hiscall));
 
 	    if (cnt > 1)
-		searchlog(hiscall);
+		searchlog();
 
 	    // <Backspace>
 	} else if (i == KEY_BACKSPACE) {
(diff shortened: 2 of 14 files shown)


dl1jbe:
Just a quick note. I will be away from tomorrow until Jan 2. Not sure if I have internet access there. If time permits I will have a look today evening at the PR

dl1jbe:
Hi Zoli. Looks good so far. Had to look for the handling of line endings, but seems ok to me.Will apply and push.

PR #153, opened 2019-12-29
Xresources

N0NB:
Hi Tom.

I've updated the doc/Xresources file to provide color palettes for xterm and urxvt that closely match the Linux console.  I've also updated the manual page to better reflect the available color settings.

Next I plan to work on a single set of default colors for the above terminals and the console.  I think doing so will give us a more consistent UI when a Linux console palette is in use.


N0NB:
I'm still not sure I have the TLFCOLORSx keyword figured out.  I'll get back to this and close this request.

PR #154, opened 2020-01-05
Rearrange header file ordering for Hamlib 4.0

N0NB:
Tom, I would consider this a critical patch until we can figure out why this broke.  BTW, I've not tested this change with Hamlib 3.3 as I have 4.0~git installed.

A recent change to Hamlib 4.0's rig.h file that defines local sleep()
and usleep() macros caused a compile error of the form:

```
  CC       readcabrillo.o
In file included from ../../tlf/src/globalvars.h:7,
                 from ../../tlf/src/readcabrillo.c:24:
/home/nate/local/include/hamlib/rig.h:108:3: error: expected identifier or ‘(’ before ‘do’
   do {\
   ^~
/home/nate/local/include/hamlib/rig.h:113:5: error: expected identifier or ‘(’ before ‘while’
   } while(0)
     ^~~~~
/home/nate/local/include/hamlib/rig.h:99:3: error: expected identifier or ‘(’ before ‘do’
   do {\
   ^~
/home/nate/local/include/hamlib/rig.h:106:5: error: expected identifier or ‘(’ before ‘while’
   } while(0)
     ^~~~~
```

Reordering the local includes to follow the system includes allows the
build to proceed.  I don't know why...

### DIFF
--- a/src/readcabrillo.c
+++ b/src/readcabrillo.c
@@ -20,8 +20,6 @@
 #ifndef _XOPEN_SOURCE
 #define _XOPEN_SOURCE
 #endif
-#include "readcabrillo.h"
-#include "globalvars.h"
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
@@ -35,8 +33,10 @@
 #include "cabrillo_utils.h"
 #include "cleanup.h"
 #include "getexchange.h"
+#include "globalvars.h"
 #include "makelogline.h"
 #include "qtc_log.h"
+#include "readcabrillo.h"
 #include "startmsg.h"
 #include "store_qso.h"
 


N0NB:
Mike, W9MDB, (author of the Hamlib patch) and I discussed this patch set a bit this morning:  https://github.com/Hamlib/Hamlib/commit/24eaeecfdfa1eacd11b64e927db75376ea9ecee8#diff-c754a75eca1c475767e238a0f8d9e583

dl1jbe:
Good catch Nate. Tested your patch with hamlib 3.3. Compiles cleanly and first test looks good. 

After reading the discussion with Mike I think he got the reasoning right. Will pick up the patch in the evening.

N0NB:
Thanks, Tom.

73, Nate

-- 

Web: https://www.n0nb.us

zcsahok:
@N0NB Nate, is this part of the public hamlib API?

If yes, then IMO it should be renamed accordingly. It's not the best practice to override core macros/functions. Other projects may run into issues if they rely on the some details of the original functionality.

If not, then it shall be moved into a private header file.

dl1jbe:
@zcsahok 
> If not, then it shall be moved into a private header file.
It is done already in Michaels repo (see https://github.com/mdblack98/Hamlib/commit/70115f7efd2e90e36ca4c20ce96f22649f323776). Just saw it some minutes before.

zcsahok:
OK, perfect, thanks.

N0NB:
My apologies.  Mike has been doing quite a bit of work that this
particular patch slipped my notice.  I caught it after merging and
installing 4.0~git locally and then having the build failure of Tlf.  I
had my suspicions that this would hit others and Walter Fey reported
several more packages.  Mike has moved it into a separate file that only
Hamlib will use and everything else should use standard sleep and usleep
functions.

If it wasn't a hobby, it wouldn't be fun!

73, Nate

-- 

Web: https://www.n0nb.us

dl1jbe:
Vera well. A quick check showed that there are other files which needs to sort out the include files.

So I will pick up your changes here and add the other files. So we can make a single commit out of it. 

Thanks anyway for the care Nate.

dl1jbe:
Fyi, just pushed a new branch 'fix_includes' to collect the needed changes in include file order. 
Picked up your changes Nate, so the PR here could be closed.

dl1jbe:
Closing.

PR #155, opened 2020-01-12
Extended memory functions (was: Grab switch back)

zcsahok:
Based on the discussions in PR #126 I added the following improved memory functionality to make it actually usable:
- in addition to the frequency memory now holds the mode (CQ/SP) and the entered call (if any)
- `%` swaps TRX and MEM (swap)
- `$` moves MEM to TRX (pop)
- in SP mode `Shift-F1` makes a swap if memory holds a previous CQ frequency and then executes `F1` (calls CQ).

These operations (and also the existing `#`) wait until the frequency is actually updated. Normally this happens quite fast with no noticeable delay.
Mode and entered call are also restored.

After the "switch back" using `Shift-F1` the previous frequency and entered call are not lost, pressing `%` restores the state, so one can continue SP and also move around or grab new stations.

Non-functional changes:
- memory handling moved to a separate file. The functions do no UI actions.
- refactored frequency/memory display in `time_update`. MEM displays scattered in the code were removed, now all is handled by `time_update`. For this a forced update mode was added: this avoids the up to one second delay before the update. Immediate update now happens also when the frequency is changed on the rig. This way also the band map gets more responsive. Band map aging is not affected by this change.

Minor stuff:
- `fabsf` replaced with `fabs`, we are not using floats any more.
- restructured `time_update` 
- added `freq` to `globalvars.h`: IMO we shall add all such variables (mostly defined in `main.c`) to this header file to get rid of ad-hoc `extern` definitions. (i.e. externs should be used only in headers)

### DIFF
--- a/share/logcfg.dat
+++ b/share/logcfg.dat
@@ -129,8 +129,6 @@ CWTONE=800
 #
 #RIT_CLEAR
 #
-#SHOW_FREQUENCY
-#
 #################################
 #                               #
 #  INFORMATION WINDOWS          #
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -30,6 +30,7 @@ tlf_SOURCES = \
 	setparameters.c show_help.c showinfo.c showpxmap.c \
 	showscore.c showzones.c sockserv.c speedupndown.c   \
 	stoptx.c store_qso.c sunup.c splitscreen.c startmsg.c\
+	trx_memory.c \
 	rtty.c time_update.c ui_utils.c \
 	write_keyer.c writecabrillo.c writeparas.c \
 	zone_nr.c
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -62,12 +62,13 @@ noinst_HEADERS = \
 	paccdx.h parse_logcfg.h prevqso.h printcall.h \
 	qrb.h qsonr_to_str.h qtc_log.h qtcvars.h qtcwin.h qtcutil.h \
 	readcalls.h readqtccalls.h readctydata.h recall_exchange.h \
-	rules.h readcabrillo.h \
+	rules.h readcabrillo.h rtty.h \
 	score.h scroll_log.h searchcallarray.h searchlog.h sendbuf.h \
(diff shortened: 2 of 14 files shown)


dl1jbe:
Looks very well on a first check. Great work.
trx_memory.h needs also to be added to src/Makefile.am.
When done we should also document the changes in the man page.

zcsahok:
Added trx_memory.h to Makefile.am

The man page and other doc update I'd do later when the functionality is stable. Let's add it to the TODO list.

There are two points I forgot to mention:
- this change removes the undocumented SHOW_FREQUENCY functionality. I think the idea was to show the frequency in big digits as for `Ctrl-F`, but I couldn't get it working. Probably it was broken for some time. If we want it back, then we shall decide what it shall do. If not, then further cleaning can be done.

- the other thing that has changed is the "feature" that on memory store/recall (`#` key) the MEM field "blinked": it was first drawn with a blue background as the call input field and then repainted by `time_update` in normal colors. I guess this was not intentional. If this was on purpose or we find that it's a nice thing then I can add it somehow.

dl1jbe:
> The man page and other doc update I'd do later when the functionality is stable. Let's add it to the TODO list.

Right my comment should only be a reminder for later. 

> There are two points I forgot to mention:
> 
>     * this change removes the undocumented SHOW_FREQUENCY functionality. I think the idea was to show the frequency in big digits as for `Ctrl-F`, but I couldn't get it working. Probably it was broken for some time. If we want it back, then we shall decide what it shall do. If not, then further cleaning can be done.

I saw it. Yes it was broken as bandmap update did overwrite the frequency display. I think it can go. 

>     * the other thing that has changed is the "feature" that on memory store/recall (`#` key) the MEM field "blinked": it was first drawn with a blue background as the call input field and then repainted by `time_update` in normal colors. I guess this was not intentional. If this was on purpose or we find that it's a nice thing then I can add it somehow.

No, that was no feature, just happened 'by accident'.

zcsahok:
OK, removed references to SHOW_FREQUENCY.

dl1jbe:
Applied and merged. Thanks.

PR #158, opened 2020-01-30
Fix ARRLDX scoring for W/VE stations

dl1jbe:
QSOs between W/VE stations should not count as multi and also give no
points.

Add missing initialization of w_cty and ve_cty variables and fixes logic to count worked countries.

Fix also evaluation of COUNTRYLIST if EXCLUDE_MULTILIST is in use. Old code changed the 'countrynr' variable which was used later.

### DIFF
--- a/src/readcalls.c
+++ b/src/readcalls.c
@@ -411,7 +411,7 @@ int readcalls(void) {
 	    int ci = 0;
 	    int countrynr_tocheck = countrynr;
 	    while (strlen(countrylist[ci]) != 0) {
-		if (getctydata(countrylist[ci]) == countrynr_tocheck) {
+		if (getctynr(countrylist[ci]) == countrynr_tocheck) {
 		    excl_add_veto = 1;
 		    break;
 		}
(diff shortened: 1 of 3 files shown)


N0NB:
I can build it and play with a dummy log inputting various calls, if you'd like.

dl1jbe:
Nate, it should basically work. But sorry, I am not firm with all the exceptions (KH6, KL7, DOC, ..) mentioned on their web side. 

So someone from W looking into it would be nice. But take your time, seems it is not urgent. The code is broken at least since TLf-0.9.10 (long time before me).

N0NB:
DC Looks to simply not count as a mult since it is not a state but will count as QSO point credit for non-W/VE stations, as I understand it.  KH6/KL7 are treated as DX entities and so are not worked by DXCC entities outside of W/VE.  KH6/KL7 are only available to W/VE stations.

http://www.arrl.org/arrl-dx

Scoring is any W/VE is zero points for W/VE (W defined as USA contiguous 48 states.   VE defined as all Canadian provinces/territories except for DXCC entities St. Paul and Sable Islands).

DXCC entities only score W/VE.

HTH

N0NB:
if the exchange validator is used, then W/VE stations will often get reports like "599 400" or "599 k", the latter being the abbreviation for "kilowatt", of course.  The rules note that such an abbreviation is acceptable.

dl1jbe:
> DC Looks to simply not count as a mult since it is not a state but will count as QSO point credit for non-W/VE stations, as I understand it. 

OK, needs to check it.

> KH6/KL7 are treated as DX entities and so are not worked by DXCC entities outside of W/VE. KH6/KL7 are only available to W/VE stations.

That seems to be ok, as far as I have tested here.

> http://www.arrl.org/arrl-dx
> 
> Scoring is any W/VE is zero points for W/VE (W defined as USA contiguous 48 states. VE defined as all Canadian provinces/territories except for DXCC entities St. Paul and Sable Islands).

Maybe the exceptions are a problem.

Do you know of any simple scheme to get province/territory from calls sign? I remember seeing one for VE stations but am not sure if such a relation exist in for W.

N0NB:
The latest CTY.DAT file gives the prefixes of CY9 and CY0 to St. Paul and Sable respectively.  How Industry Canada assigns those prefixes or if any stations operating from there just sign CY0/VE1XYZ, I don't know.

As for other provinces, it seems as though Canadians do honor them much more than we in the US do, e.g., VE2 is QC, VE3 is ON, VE7 is BC, etc.

Don't let perfect be the enemy of good here.  The main thing is to be an aid to the operator but not dictate what will or will not be logged.  At the end of the day the Cabrillo file doesn't contain any of our guesses of points and mults but should be faithful to what the operator entered during the 'test.

N0NB:
Playing around for just a short bit yesterday I entered these:

```
 40CW  30-Jan-20 20:59 0001  DL1JBE         599  599  300           DL       3   7229.3
 40CW  30-Jan-20 21:00 0002  W1AW           599  599  CT                     0   7229.3
 40CW  30-Jan-20 21:18 0003  KL7AJ          599  599  20            KL       3   7229.3
 40CW  30-Jan-20 21:19 0004  VE3KAO         599  599  ONS                    0   7229.3
 40CW  30-Jan-20 21:19 0005  K9FD           599  599  500           KH6      3   7229.3
```

Of particular interest was `K9FD` a well-known station located in KH6.  Tlf showed immediately that `K9FD` was correctly identified as being in KH6, obviously pulled from `CTY.DAT`.  Tlf also handled the mult and score correctly here in W0.

N0NB:
Tom, I think you can merge this.  Perhaps make a bit of an announcement for US ops planning on working ARRL DX or even a point release.

dl1jbe:
> The latest CTY.DAT file gives the prefixes of CY9 and CY0 to St. Paul and Sable respectively. How Industry Canada assigns those prefixes or if any stations operating from there just sign CY0/VE1XYZ, I don't know.
> 
All stations with prefixes not belonging to W and VE in CTY.DAT should be well counted as non US/Canadian. That works already.

> Don't let perfect be the enemy of good here. The main thing is to be an aid to the operator but not dictate what will or will not be logged. At the end of the day the Cabrillo file doesn't contain any of our guesses of points and mults but should be faithful to what the operator entered during the 'test.

And I think that was a wise decision to migrate to the cabrillo format. We should only try to give a good guess about the result in advance.

dl1jbe:
> Tom, I think you can merge this. Perhaps make a bit of an announcement for US ops planning on working ARRL DX or even a point release.

Oh, yes. The CW contest is just in two weeks from now. Thanks for reminding.

I already played with the idea of an point release, but am not sure, that we can manage it in the next two weeks. So an announcement should be appropriate.

And @zoli thanks for the 'go ahead'.

PR #160, opened 2020-02-15
improve WWV handling

zcsahok:
WWV messages are parsed by TLF, but they were shown in the footer just until the next screen rewrite. The parsed data is displayed in the MUF screen (`Ctrl-P`) starting at position 40, thereby potentially losing half of the string.  The parsing was also a bit sloppy.

There are a number of improvements/fixes:

- fixed parsing and status (footer) line generation
- footer shows the last WWV info for 3 minutes after its receipt
- MUF screen shows the parsed and also the original info. Both are split on multiple lines if needed.
- no MUF calculation if destination is not known
- MUF calc uses same Earth radius as `qrb.h` (ARC_IN_KM), now the distance matches on MUF screen and on the footer after search
- `parse_logcfg.c` calls setter functions instead of creating a fake WWV string
- on call input line date is updated not every 60s but 2s, i.e. one doesn't have to wait a minute to get the right date after 0:00. The performance impact is minimal.
- MUF's global variables got better names than `r` and `m`.
- added reference and attribution to `muf.c` source
- `cleanup` is called after startup
- found some places where `spaces()` could be used
- in `globalvars.h` moved `tlf.h` after the system includes
- added test for WWV parsing (it includes an original result with some issues)

Some `mvprintw` were changed to `mvaddstr`. The potential issue with using `mvprintw` for "just" showing a string is that it interprets the argument as a format string and in case it contains `%` the the result will not be as expected.

Also started using `move` instead of `mvprintw` for cursor positioning.

### DIFF
--- a/src/cleanup.c
+++ b/src/cleanup.c
@@ -23,40 +23,31 @@
  *--------------------------------------------------------------*/
 
 
-#include "tlf.h"
+#include "globalvars.h"
 #include "tlf_curses.h"
 #include "ui_utils.h"
 #include "write_keyer.h"
 
 void cleanup_qso(void) {
-    extern char hiscall[];
-    extern char comment[];
-    extern char my_rst[];
-    extern char his_rst[];
-
     hiscall[0] = '\0';	    /* reset hiscall and comment */
     comment[0] = '\0';
     his_rst[1] = '9';	    /* reset to 599 */
     my_rst[1] = '9';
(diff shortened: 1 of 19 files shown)


dl1jbe:
Only one minor warnings about different prototype for wwv-add(). 
We can fix that in next commits.

dl1jbe:
Thanks. Applied and pushed.

PR #161, opened 2020-02-20
Refactor readcalls

dl1jbe:
Main goal of the proposed change is a serious refactoring of the readcall() function.
As new features got added in last years it also grows to a tangled net which was nearly unmaintainable.

The patch contains mostly the following parts
- factor out initialization 
- factor out reading and decoding values from recorded log lines and handling of multis
- drop redundant local variables and use existing ones if they exist (e.g. nr_worked)
- switch to g_strlcpy where appropriate
- simplify the code by dropping redundant tests
- factor out lookup if station is in country  or continent list and use that functions throughout TLF
- factor out scoring logic which counts worked countries and zones 

There were some other changes made:
- drop 'other_flag' variable which was superseded by wysiwyg_multiband
- use test for country and continent list in score.c
- move some variables and prototypes to bool type
- add some tests for factored out functions

As there were nearly no tests in place the patch was made as a series of small steps to enable to follow the made transformations.
The code got some moderate testing in comparing results for known contests before and after modification but could use a second look from outside.

So it would be good if some people could have a look at it and come back with comments, suggestions and/or error reports.

### DIFF
--- a/ToDo
+++ b/ToDo
@@ -17,7 +17,7 @@ Planned for later versions
 
 [x] automatically switch back to 'Run' mode from "grab spot" after QSO (tnx
     Zoli)
-[ ] document new behaviour in man page aftre tests
+[ ] document new behaviour in man page after tests
 [x] Add contest name to Header of ADIF file (suggested by SP3RXO)
 [ ] Refactor write_adif function. Use an add_adif_field() helper and also 
     already existing get_next_record() function.
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -88,7 +88,7 @@ int addcall(void) {
     extern t_pfxnummulti pfxnummulti[MAXPFXNUMMULT];
     extern int pfxnummultinr;
     extern int addcallarea;
-    extern int continentlist_only;
+    extern bool continentlist_only;
     extern char continent[];
     extern int exclude_multilist_type;
     extern int trxmode;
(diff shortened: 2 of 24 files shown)


zcsahok:
Fine with me.

There are still places where we can switch to boolean flags and use boolean conditionals like

```
if (continentlist_only) {
..

if (!continentlist_only && ... ) {
```

``country_found()`` in ``score.c`` still doesn't use its argument (prefix), instead it works with globals. No problem for the moment, but deserves a fix later on.

``bandnr2index`` could be somewhat reduced as

```
int bandnr2index(int nr) {
    switch (nr) {

	case 160:
	    return BANDINDEX_160;

	case 80:
	    return BANDINDEX_80;
...
       default:
            return BANDINDEX_OOB;
```

dl1jbe:
Thnaks for the check.

> There are still places where we can switch to boolean flags and use boolean conditionals 

Done.  

> `bandnr2index` could be somewhat reduced as
> 
> ```
> int bandnr2index(int nr) {
>     switch (nr) {
> 
> 	case 160:
> 	    return BANDINDEX_160;
> 
Done. See last two commits.
 
> `country_found()` in `score.c` still doesn't use its argument (prefix), instead it works with globals. No problem for the moment, but deserves a fix later on.

True. That is still to do. As it is used globally on quite some places we need to make sure to not break other code with that change.

PR #164, opened 2020-03-20
Giving the initial serial value

airween:
In some cases, we need to start the serial number other than 0001 - eg. a contest with more periods with continuous serial (like SCWC - sorry, didn't find any online documentation). This patch can helps to split the periods.

### DIFF
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -172,6 +172,17 @@ int getidxbybandstr(char *confband) {
     return -1;
 }
 
+void setnext_qsonumber(char * field) {
+    if (field != NULL) {
+	if (atoi(field) > 0) {
+	    highqsonr = atoi(field);
+	}
+	else {
+	    showmsg("WARNING: serial could be a positive integer...\n");
+	    exit(1);
+	}
+    }
+}
 
 static int confirmation_needed;
 
@@ -928,6 +939,7 @@ int parse_logcfg(char *inputbuffer) {
 	    break;
 	}
 	case 52: {
+	    setnext_qsonumber(fields[1]);
 	    exchange_serial = 1;
 	    break;
 	}
@@ -1104,6 +1116,7 @@ int parse_logcfg(char *inputbuffer) {
 	    break;
 	}
 	case 88: {
+	    setnext_qsonumber(fields[1]);
 	    serial_section_mult = 1;
 	    break;
 	}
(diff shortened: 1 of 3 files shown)


dl1jbe:
Sorry for late reply. CV19 taking its toll.

Thanks for the pull request. I am not sure if using highqsonr in scroll_log.c to set the start of the serial number is a good idea. highqsonr is used for synchronizing the QSO number in case you have more than one station running TLF on a LAN network.
It would be simpler (and less error prone) to set qsonum direct or just after running parse_logcfg(). 

What do you think?

airween:
> Sorry for late reply. CV19 taking its toll.

No problem - hope you're right.

Thanks for the idea, I'll check that soon. Would you like to see a separated PR, or can I add the commits into this one?

dl1jbe:
Should that work only when starting a new empty logfile?
Or do you want to quit TLF and later continue with a somewhat higher qso number effectively creating a gap in the log file?

airween:
First, sorry for the delay, I had no time to finish yet.

> Should that work only when starting a new empty logfile?

If Tlf founds (or creates) a new and empty logfile, then it will starts to count from the number what you given. If you break the contest, and continue later, then just continues the serial from that point.

If there is a logfile already, and the first number is greater than in config, Tlf warnings you and exits.

I'm used this feature in SCWC contest about 3-4 weeks ago, which has 4 period, and I made 4 separated directory. Before I started the period, I just step into the next dir, and set the next serial.

But meantime HA5CQZ Zoli @zcsahok show me the solution for my original issue: the MINITEST keyword allows an optional argument, which sets the period time in seconds. So I don't know, what can we use this feature. :)

dl1jbe:
> If Tlf founds (or creates) a new and empty logfile, then it will starts to count from the number what you given. If you break the contest, and continue later, then just continues the serial from that point.
> 
OK, that is what I expected. I see.

> But meantime HA5CQZ Zoli @zcsahok show me the solution for my original issue: the MINITEST keyword allows an optional argument, which sets the period time in seconds. 

I am just wondering as that MINITEST feature was introduced by yourself. 

> So I don't know, what can we use this feature. :)

Ok, as I understand we have no real need for theses feature at the moment.
Than we should put it back for now.

Please decide if we keep the PR open or close it for now.

airween:
Hi Thomas,

> I am just wondering as that MINITEST feature was introduced by yourself.

yeah, I totally forgot that - but I don't remember that I implemented the argument too.
Nevermind.
 
> Ok, as I understand we have no real need for theses feature at the moment.
> Than we should put it back for now.
> 
> Please decide if we keep the PR open or close it for now.

You're right, I close this issue.

Thanks.

PR #167, opened 2020-04-06
cty.dat and callmaster updated

sm0rux:


dl1jbe:
Many thanks for the Pull request. I am a little late, sorry for that.

Just fyi, as both files change quite often we normally update them just before a new release. Anyone needing actual files can download them like described in the FAQ.

PR #168, opened 2020-04-14
Check minitest setting for dupe evaluation on bandmap

zcsahok:
In a minitest stations can be worked again once the current period is over. This was correctly handled by the dupe checking in `searchlog`.  The same check was added to `bandmap` so that stations show up again when using dupe filtering mode.

Boolified and made private `bm_ismulti` and `bm_isdupe`.

Formatting changes are due to `astyle`.


dl1jbe:
Thanks for the PR. Will Apply and push now.

PR #169, opened 2020-04-25
Updated man page + help.txt

zcsahok:
Described the recently added MEM operations (`$` , `%`, `Shift-F1`).
Somewhat normalised the key naming and capitalisation.
Fixed CQD keys and the command line help option.

The second commit is not related to the docs, just didn't want to make a separate branch for it. It fixes sending the unnecessary trailing `NUL` char to the keyer. Cwdaemon 0.10.2 works OK without it (and also any other well-behaved keyer should).

### DIFF
--- a/share/help.txt
+++ b/share/help.txt
@@ -1,93 +1,94 @@
 Keys (shortcuts) to be used in TLF
 
-If you want to see some of the start up parameters try: 'tlf -h'.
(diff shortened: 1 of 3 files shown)


dl1jbe:
Thanks. Your PR comes in time for preparation of tlf-1.4.1

The trailing NUL is also not needed for winkeydaemon (just tested).

PR #170, opened 2020-04-26
Fix frequency selection when working with Fldigi

zcsahok:
When using Fldigi the operating frequency is corrected by the pitch of audio frequency used as carrier. The value shown by Tlf is thus the actual carrier frequency.
This was not done in the reverse direction: rig's VFO was set directly, resulting in an offset. The only place considering the carrier offset was in `grabspot`.
For RTTY Tlf even corrects the frequency by half of FSK shift (85 Hz) to show the mark frequency. This causes issues when Fldigi is not in RTTY but say PSK. It took me some time to figure out why the freqs displayed by the two programs plus the rig do not match...

The main change is moving `fldigi_get_carrier` correction from `grabspot` to the central place in `gettxinfo`. This way the freq value handled by Tlf is always the carrier frequency.

`fldigi_xmlrpc_get_carrier()` now checks Fldigi modem type and applies further correction only it's RTTY.

Minor changes:
* flattened somewhat `fldigi_xmlrpc_get_carrier()` by taking early exit
* changed conditional to `#ifdef HAVE_LIBXMLRPC` as in other functions
* application name is taken from `PACKAGE_NAME` defined in `config.h`
* `int fldigi_shift_freq` doesn't have to be static in `gettxinfo()`

(next PR will be a bigger one: rework of time handling. I also have config file parsing improvement in the queue.)

### DIFF
--- a/src/fldigixmlrpc.c
+++ b/src/fldigixmlrpc.c
@@ -47,7 +47,6 @@
 #include "tlf_curses.h"
 #include "ui_utils.h"
 
-#define NAME "Tlf"
 #define XMLRPCVERSION "1.0"
 
 int fldigi_set_callfield = 0;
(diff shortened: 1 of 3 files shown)


dl1jbe:
Thanks for the PR Zoli. As I am just preparing a 1.4.1 minor release (mainly to bring out the GCC-10 bug fix) do you see your as urgent. Otherwise I would be wait till the 1.4.1 is out (hopefully next on or two days).

Furthermore while you are on fldigi we should discuss PR #102 to bring that one forward. 

(My next working field will be the discussed changing of RST).

zcsahok:
This one is not urgent, probably not many people use Tlf with Fldigi. I'm operating [HG20AY](https://www.qrz.com/db/HG20AY) in digimode, also partially to have some live testing...

Will also check #102.

dl1jbe:
Applied & Pushed. Thanks.

Issue #137, opened 2019-11-03
Page UP / Page DOWN to change reports 


M0ORY:
This should really be configurable as not all keyboards these days have PgUp and PgDown keys. Macs in particular as the OS has no concept of PgUp / Down and so even plugging in an IBM keyboard these keys are not interpreted as intended meaning it's impossible to change signal reports on a mac either via ssh through the terminal or via a virtual machine console. 
Ideally I would use the up down arrows to change reports or just have the option to re-key them as a normal exchange field. 

I've been looking at the code on and off for the last year trying to figure this out for myself however it seems to be slightly beyond my skill level.

Cheers
Ben

zcsahok:
Up arrow would clash with log editing. We can check however if shift-up/down could be an option. The keys are currently hardcoded.
Thanks for the report.

zcsahok:
BTW did your try `Fn`+`Up`/`Down`?

https://apple.stackexchange.com/questions/26930/how-can-i-achieve-page-up-and-page-down-in-os-x

M0ORY:
Yes I've tried that in multiple ways. Via SSH from the Mac Terminal, via Virtual Box into an Xubuntu terminal and via VirtualBox into a console. None appear to work however when reading through the source last night I noticed that in CW mode pgup pgdn also changes the CW keyer speed and this does appear to be changing suggesting that the key combo does work. To test this I also tried it in Vim to see if it worked at all and Vim does interpret that key combination as pgup pgdn from a functional perspective .

Also looking at line 446 and down of callinput.c it looks like this is supposed to change CW speed if there is nothing in the callsign field but if the callsign is populated it should change the RST provided the change_rst variable = 1. A little digging later and I see the man page states that CHANGE_RST should be set (doesn't say where or how) I tried putting CHANGE_RST in logcfg.dat and now it works through the mac terminal but not through a VBox console or Xubuntu terminal within virtual box. 

It would probably be useful to have this (and any other such options) as a commented out option in the default logcfg.dat as it would be much more obvious to the user especially those of us with sponky keyboards requiring work arounds which may or may not work given certain scenarios. There's also nothing in the help file or the old documentation stating that this needs to be set in order for it to work. 

[...]

PR #171, opened 2020-05-09
Fix manual keying using Fldigi

zcsahok:
This fix addresses the issue described in #102. Currently keying sends Fldigi messages (both Fn keys and individual keystrokes) using this logic:
 * abort running TX if any
 * append message + "^r" to TX widget
 * start TX

This is fine for Fn messages but unusable for interactive keying (Ctrl-K).

This PR doesn't solve all the problems related to digi keying but at least makes interactive keying usable by supporting manual TX/RX control.
There are two main areas of change:
*  `keyer.c` handles `{` and `}` as _start-tx_ and _switch-to-rx_ also for Fldigi keying. There were already provisions for this for MFJ1278.	Similarly `|` key inserts a line break.
* `fldigi_send_text()` in `fldigixmlrpc.c` checks if interactive mode is used and in that case it doesn't abort TX and makes TX/RX handling based in user input.
	TX/RX control is done using Ctrl-T and Ctrl-R characters as for MFJ1278.
	
The keyer window got also some minor changes:
* now Ctrl-K or Alt-K closes keying mode gracefully, without aborting a running TX.
	This allows switching back to the main screen for call editing, entering serial etc. while TX is still active.
	Currently user has to wait until end of TX before closing keyer. Escape still exits with stopping TX.
 * backspace char is sent even when keyer input field is empty.

The latter is due to the issue that Tlf keyer input and the actual Fldigi TX widget content can get out of sync if `}`/`|` and backspace is used as the former two characters are not represented on the keyer input line.
A use case: one prepares a text for transmission by typing `ABCX}`. Before starting TX with `{` user want correct `X` to `D`.
Keyer input shows `ABCX` so pressing backspace removes `X`, but on Fldigi side it removes the final `r`.
To fix typing user still has to type backspace-backspace-D-}. This makes keyer input field quite mangled, but at least Fldigi shows the correct text.

Now, if in a new keying session one wants to change `E}` into `R}` then it wouldn't be possible without allowing backspacing past the beginning of the input field.
Another option would be to actually insert ^r and | into the input line. 
However this requires a bit more changes in keyer.c and since Fldigi UI is anyway visible the user can use it to check the typing.

There is also an issue with Fldigi itself: it doesn't handle backspacing correctly while transmitting. Fldigi stucks in TX mode and although this works from Fldigi UI, it doesn't via XML RPC. Will file a bug report once I figure out how...

There are multiple ways to use interactive digi keying, but given the quirks above the recommended procedure is
* type the text, correct it if needed
* press `{` and `}`, they're normally located on adjecent keys, a quick tap is enough and order is irrelevant
* exit keyer mode with Ctrl-K

Boolified flags in `fldigixmlrpc` and flattened a bit `fldigi_xmlrpc_query()`.

Notes:
* even though comment in `fldigi_xmlrpc_query()` says that Tlf reconnects to Fldigi automatically, this is not the case.
* Fn messages contain leading (or trailing?) new lines although not specified explicitly in config file. This is partly useful, partly not, depending on the use case. I'll check this later.

### DIFF
--- a/src/fldigixmlrpc.c
+++ b/src/fldigixmlrpc.c
@@ -39,7 +39,9 @@
 
 #include "err_utils.h"
 #include "fldigixmlrpc.h"
+#include "globalvars.h"
 #include "getctydata.h"
+#include "keystroke_names.h"
 #include "logit.h"
 #include "printcall.h"
 #include "searchlog.h"
@@ -63,14 +65,14 @@ typedef struct xmlrpc_res_s {
 				   it back to RIG carrier */
 
 extern char fldigi_url[50];
-static int use_fldigi;
+static bool use_fldigi = false;
 
 int fldigi_var_carrier = 0;
 int fldigi_var_shift_freq = 0;
 #ifdef HAVE_LIBXMLRPC
-static int initialized = 0;
+static bool initialized = false;
 #endif
-static int connerr = 0;
+static bool connerr = false;
 
 char thiscall[20] = "";
 char tcomment[20] = "";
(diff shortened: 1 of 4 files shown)


dl1jbe:
Thanks for sorting things out, Zoli. Does that mean, we can close issue #102 now?

zcsahok:
yes, please close #102 
I'd also close #137 as there was no activity on it recently and it is covered by #172.

PR #174, opened 2020-05-17
Fix time handling and CQWW simulator

zcsahok:
(sorry for the Big Bang-like commit, but time related code is scattered all around, not easy to make separate fixes.)

The motivation for the fix is that after DST switch I started sometimes see a time on call input line that was 1 hour off. Logging worked with the correct time, just the input line was showing sporadically wrong time. Even though I was not able to reproduce it but I do have possible suspects.

Number 1 is the ` gtime()` function that stores its result into a shared static structure. It's clearly not thread safe. Number 2 is `mktime()`: it updates the TZ and DST flag in the shared structure.

Consider this snippet:
``` 
    time_t now = time(0);

    struct tm * time_ptr = gmtime(&now);

    char buf[128];
    strftime(buf, sizeof(buf), "date1: %a, %d %b %Y %H:%M:%S %Z", time_ptr);
    puts(buf);

    time_t t1 = mktime(time_ptr);

    strftime(buf, sizeof(buf), "date2: %a, %d %b %Y %H:%M:%S %Z", time_ptr);
    puts(buf);

    time_t t2 = mktime(time_ptr);
    printf("now=%d  t1=%d   t2=%d\n", now, t1, t2);
```
One would expect the same date/time output twice with now=t1=t2.
What actually happens is something like
```
date1: Wed, 13 May 2020 06:36:39 GMT
date2: Wed, 13 May 2020 07:36:39 CEST
now=1589351799  t1=1589348199   t2=1589348199
```
This was executed on Wed 13 May 2020 08:36:39 AM CEST. `date2` has been messed up by 
`mktime`. I believe this was the reason for the wrong time I was seeing.

All time related operations (getting, formatting, parsing) were moved to `get_time.c` and are done in a thread safe way. Internally time is stored strictly in UTC epoch (seconds since 1970.01.01) instead of mktime's output, Tests were extended to check this.

(a good read on time in C: http://www.catb.org/esr/time-programming/)

As a byproduct of the time-related fixes I had to touch the simulator and figured out the it doesn't fully work as intended. The major issue was that repeating was triggered by looking for a `?` on the call line, but this we don't have since who knows when as `?` is caught already in `callinput.c` so `logit.c` won't get it.
The code was separated from `background_process` and moved into `cqww_simulator` as this is what it actually does. I kept the funny time-based "random" generators. The countries come in a  barely random order if callmaster file is sorted. It's no problem, one just has to shuffle the file for a different pattern. As a new feature I added a progressive slow down for repeat.
A note on repeat: it's triggered by any message (F-key) but upon completing the logging the state is overwritten to FINAL. I.e. there is no explicit check for F5/F8/F9/F10 that one would typically use for this purpose.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -8,7 +8,8 @@ tlf_SOURCES = \
 	background_process.c bandmap.c bands.c \
 	cabrillo_utils.c calledit.c callinput.c changefreq.c changepars.c \
 	change_rst.c checklogfile.c checkqtclogfile.c \
-	checkparameters.c cleanup.c clear_display.c clusterinfo.c cw_utils.c \
+	checkparameters.c cleanup.c clear_display.c clusterinfo.c \
+        cqww_simulator.c cw_utils.c \
 	dxcc.c \
 	deleteqso.c displayit.c \
 	edit_last.c editlog.c err_utils.c \
@@ -45,7 +46,7 @@ noinst_HEADERS = \
 	cabrillo_utils.h calledit.h callinput.h changefreq.h changepars.h \
 	change_rst.h checklogfile.h checkqtclogfile.h \
 	checkparameters.h cleanup.h clear_display.h clusterinfo.h \
-	cw_utils.h \
+	cqww_simulator.h cw_utils.h \
 	dxcc.h \
 	deleteqso.h displayit.h \
 	edit_last.h editlog.h  err_utils.h \
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -34,7 +34,6 @@
 #include <string.h>
 
 #include <glib.h>
-#include <time.h>
 
 #include "addcall.h"
 #include "addmult.h"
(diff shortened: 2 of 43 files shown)


dl1jbe:
> (sorry for the Big Bang-like commit, but time related code is scattered all around, not easy to make separate fixes.)

Wow, really quite a big patch. 

Very well. Thanks for sorting it out. Time handling was a mess before.  

> Number 1 is the ` gtime()` function that stores its result into a shared static structure. It's clearly not thread safe. Number 2 is `mktime()`: it updates the TZ and DST flag in the shared structure.
> 

That behavior of mktime is well documented in the man page, but it seems we did not pay enough attention to it before. Glad you found it.

> All time related operations (getting, formatting, parsing) were moved to `get_time.c` and are done in a thread safe way. Internally time is stored strictly in UTC epoch (seconds since 1970.01.01) instead of mktime's output, 

So it is a clean design now. I like it.

> As a byproduct of the time-related fixes I had to touch the simulator and figured out the it doesn't fully work as intended. The major issue was that repeating was triggered by looking for a `?` on the call line, but this we don't have since who knows when as `?` is caught already in `callinput.c` so `logit.c` won't get it.
> The code was separated from `background_process` and moved into `cqww_simulator` as this is what it actually does. I kept the funny time-based "random" generators. 

Feel free to replace it by rand() if you like.

> The countries come in a barely random order if callmaster file is sorted. It's no problem, one just has to shuffle the file for a different pattern. As a new feature I added a progressive slow down for repeat.
> A note on repeat: it's triggered by any message (F-key) but upon completing the logging the state is overwritten to FINAL. I.e. there is no explicit check for F5/F8/F9/F10 that one would typically use for this purpose.

OK. Let me test it a little bit more tomorrow. than I can pick it up.

PR #177, opened 2020-07-14
Arrl fd

k8lt:
ARRL Field Day logs can be submitted as a Cabrillo file, so I added a Cabrillo format definition.

When exporting FD logs in ADIF format I noticed a couple of things which might be improved:

- For a NO_RST contest, the <RST_SENT> and <RST_RCVD> tags can just be omitted.
- ADIF 3.1.0 defines the <SRX> and <STX> tags as integer serial numbers, so non-serial number exchanges should be
  written using <SRX_STRING> and <STX_STRING>.


dl1jbe:
Hi Gary. Thanks for the contribution and sorry for being late.
Good work. 73, de Tom

k8lt:
No problem.  Glad I could help.
73,  Gary

Issue #178, opened 2020-07-27
Scoring problems


dl1jbe:
SP3RXO reported some problems with the scoring logic.

CONTINENTLIST=EU,AS
COUNTINENT_LIST_POINTS=2
USE_CONTINENT_LIST_ONLY
COUNTRY_MULT

gives 2 points for AS stations but -1 (!!!) for EU stations.

dl1jbe:
Furthermore if USE_CONTINENT_LIST_ONLY gets dropped CONTINENT_LIST_POINTS will not be applied at all.

dl1jbe:
Part 1 of the problem got fixed by PR #189.

While thinking about the second part it showed that there is no easy solution. There are two main reasons for that:

Problem
------------

1.  It is not really clear (and not easy to see) which role a CONTINENT_LIST_POINTS should play and which priority it has if other keywords (MYCOUNTRY_POINT, MYCONTINENT_POINT , ...) are set.

2. The scoring mechanism has grown into a complicated peace of code. It started by points for my country, my continent and other continents. Due to new contests with more complex rules we added such things as COUNTY/CONTINENT_LIST and more. Besides not really defined priorities there are further problems as the continent and country lists are sometimes used for scoring points and sometimes for counting multis.

Possible solution
------------------------

I would like to suggest a complete new mechanism which (hopefully) will avoid these problems and can led to a much more clean scoring code.

1. There will be two list for point scoring of a) continents and b) countries

1.   In each list you can name a continent or country together with its points (e.g. EU 2 points, all other continents 3 point, DL in the other list with 1 point would give the known 1 point for own cty, 2 for own continent and 3 points for DX) . Different points for each continent/country are allowed.

1.  Continents/countries not in the lists will get no points at all.

1.  Country points will override continent points

Scoring will be simple:

a) Assume 0 points
b) If in Continentlist use the points from there
c) If in Countrylist override points so far by the points from the list

I suggest two new keywords:
- COUNTRYPOINTS=<cty1,cty2,cty3:points1>;<cty4,....,ctyn:points2>; ...
- CONTINENTPOINTS=<cnt1,cnt1:points1>;cnt3,...,cntn:points2>

More than one of each keywords are allowed in the config file. They are cumulative so you can split a long list of countries to more than one line.

Discussion
---------------

With the above changes we should be able to 

[...]

PR #189, opened 2020-11-10
Fix scoring

dl1jbe:
Correct scoring of CONTINENT_LIST_POINTS (see first part of issue #178)

### DIFF
--- a/src/score.c
+++ b/src/score.c
@@ -186,15 +186,14 @@ int scoreByContinentOrCountry() {
     extern char continent[];
 
     int points = 0;
-    int inList = 0;
+    bool inList = false;
 
     inList = exist_in_country_list();
     if (countrylist_only) {
-	if (inList == 1 && countrylist_points != -1)
+	if (inList && countrylist_points != -1)
 	    points = countrylist_points;
     } else {
-
-	if (inList == 1) {
+	if (inList) {
 	    if (countrylist_points != -1)
 		points = countrylist_points;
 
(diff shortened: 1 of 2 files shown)


zcsahok:
Just a related question: a number of related configs (CONTINENT_LIST_POINTS, MY_CONTINENT_POINTS, etc) do not allow overwriting, i.e.
```
            if (my_cont_points == -1)
                my_cont_points = atoi(c_temp);
```

Is this intentional?

I looked at the condition in USE_IF_SET and was wondering why ``!= -1`` is used, so checked other parts of the code. (my preference would have been ``>= 0``)
I'm working on improving config parsing, hence the question.

dl1jbe:
> Just a related question: a number of related configs (CONTINENT_LIST_POINTS, MY_CONTINENT_POINTS, etc) do not allow overwriting, i.e.
> 
> ```
>             if (my_cont_points == -1)
>                 my_cont_points = atoi(c_temp);
> ```
> 
> Is this intentional?

Good question. I did a lookup in old versions and found it already in tlf-0.9.10 dated from 2003.

So I do not really have an answer but assume it was so that only your first keyword got used if you put the keyword in the config twice. 

I think it would be ok to drop it and accept the last occurrence of the keyword instead only the first one.

> I looked at the condition in USE_IF_SET and was wondering why `!= -1` is used, so checked other parts of the code. (my preference would have been `>= 0`)
> I'm working on improving config parsing, hence the question.

The '!= -1' comes from the default initialization of the variables. But, yes any negative number could be used as an 'not set' indicator'.

PR #191, opened 2020-11-18
Test logcfg parsing

zcsahok:
Added test for logfcg parsing function. The goal is to refactor the monolithic switch-case statement into a data driven solution. For this all configs shall be covered with tests. As there are 250+ of them this will take some time. The migration can be done gradually, i.e. the ones tested can be changed to the new approach. This way we can always be sure the all is stable during the process.

The second unrelated commit just fixes some typos caught by ispell.


dl1jbe:
oh, oh, so much typos... I should use a spellchecker myself in future. Thanks for fixing that mess.

dl1jbe:
Tests are looking good. That will really help but also will be a lot of work.

PR #194, opened 2020-12-01
More tests and common handling for unkown/deprecated config keywords

zcsahok:
Deprecated ones are now marked with `NULL` and for this changed string comparison to `g_strcmp0`. 

Unknown keyword reports the keyword part instead of the whole line making error message more logical.

Extended a bit the test of cw_utils. (have plans to change the code...)

### DIFF
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -318,8 +318,8 @@ int parse_logcfg(char *inputbuffer) {
     extern int verbose;
 
     char *commands[] = {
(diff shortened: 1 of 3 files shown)


dl1jbe:
Good work. Shall I pick it up or wait a little bit more for things to come?

zcsahok:
I'll commit some more test today, then it can be merged.
Will use the same branch for further work on configs.

zcsahok:
Done for today.

dl1jbe:
Applied and pushed.

PR #195, opened 2020-12-09
Fix wpx usage

dl1jbe:
Fix usage of 'wpx' variable

Background:
- Variable gets set by setcontest() if 'whichcontest' == "wpx" or it is
  set if we use PFX_MULT keyword (sets 'wpx' together with 'pfxmult')
- The usage in tests is therefore mixed:
  * Whenever we test for 'wpx == 1' we want to execute the code for WPX
    contest AND also if PFX_MULT is used
  * All tests for 'wpx == 1 && pfxmult == 0' are than true if only WPX
    contest is in use.

Fix:
- To decouple things we have to:
  * drop setting 'wpx=1' for PFX_MULT
  * convert all use of 'wpx == 1' to 'wpx ==1 || pfxmult == 1' and
  * convert all use of 'wpx == 1 && pfxmult == 0' to a simple 'wpx ==1'

### DIFF
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -43,6 +43,7 @@
 #include "getctydata.h"
 #include "getpx.h"
 #include "get_time.h"
+#include "globalvars.h"
 #include "log_utils.h"
 #include "paccdx.h"
 #include "score.h"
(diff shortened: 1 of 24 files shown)


zcsahok:
OK with me. Once merged I'll create a PR for further config tests and the new config processing code.

airween:
I just review the code, but couldn't tested. It's good to see the simplified logic :).

airween (approved):

dl1jbe:
Thanks for the check to both of you. The old logic was quite twisted and I was not sure if I made some errors in transformation.
I am on the way to make the predefined contests mutual exclusive and table driven. That for we have to sort out these mixed logic. There will be a PR for it soon.

Zoli, just go ahead with your changes.

PR #196, opened 2020-12-09
logcfg parsing tests + new code

zcsahok:
More tests and the new code for parsing.

The easy configs are bools/ints/strings that can be described with their parsing+checker functions. The keyword specification is regexp that allows the use of capture groups to simplify handling of configs like F1..F12.

There are of course a number of issues in the current code that shall be fixed.
Just to name a few -
* some int values have no range checking, others have it and
  * ignore out-of-range values (CWTONE) or
  * take the limit in this case (WEIGHT) or
  * take some default (CQDELAY).
* strings and messages could be static or dynamic; some are chomped, some are not; some even require the trailing NL
* most (all?) errors are reported as warnings and allow continuing; others make a straight exit (but still just show a warning)

All this chaos has to be somehow sorted out or at least reduced.

Now around 65 keywords are still to be converted. These typically will require changes in the places where they are actually used, so a slower progress is expected.

Let me know if this approach is fine.

The same machinery can be later used in `changepars.c`.


zcsahok:
`g_ascii_string_to_signed` and co require GLib 2.54 or later

zcsahok:
CI passed on Bionic. So either we stay on it and require at least Bionic/Buster, or I can add a simpler implementation for integer parsing and then we can continue using GLib >= 2.40, i.e. even Strech.

dl1jbe:
As 2.54 came out in 2017 I fear there are quite some installations out there with an older code base (I remember complaints as I tried to use some newer GLib features some times ago :-) ). We jumped to >= 2.40 in 2019 - 5 years after it came out. If we want to keep that distance from the release the max we could use is >=2.45. 

For the code itself, please give me another day to look through. Anyway it looks promising.

zcsahok:
I agree with not pushing GLib version too fast. Changed parsing to use a more "ancient" function that is there since 2.12.

airween:
Hi Zoli,

it's a pretty nice patch, congrats!

About a year ago, in last year when the CQWW was during (our personal meeting :)) we talked abut this feature. The reason was [this fetaure request](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00036.html).

Now you reorganized the config keywords - wouldn't it be time to introduce the mandatory source file too for each keywords?

A small remarks [here](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00063.html).

dl1jbe:
@zcsahok Hi Zoli, finally found time for a deeper look into the code. Very good work, I like it.

I have some minor questions I would like to ask before merging:

- The comments in line 1545 and 1546 moved left for each commit you made. Any idea what happened?
- You replaced a lot of checks with your fine new machinery, but left the original code in place. Do you want to drop it all in one step?
- You added  lot of external variable definitions even when they were applied by '#include "globalvars.h"'. Maybe we should discuss how to handle it in general and try to convert all files to the same rule.
- and finally should we better squash the last 6 commits into one? I think that would make sense.

@airween From the discussion in your mentioned  [thread](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00036.html) I got the impression that you started working on the feature. Any news on that?

zcsahok:
It's astyle playing with those comment lines. I was also wondering what is going on.

The process is: create a test for a keyword, add new style parsing and then mark it with NULL in `commands[]`. Old code is not touched and kept for easier fallback during testing. Once all keywords have been migrated, then I'll remove the legacy code incl. the mega-switch in `parse_logcfg()`.

Ideally all vars shall come from a .h file. `globalvars.h` is probably the place for the ones defined in `main.c`, whereas other .c's provide their own .h. I was just blindly moving `extern`s out of `parse_logcfg()` when they were used. I think cleaning this up - and removing random externs generally - will be a next step.

Yes, of course, please squash the commits.

@airween finally I got a free slot for this topic. hopefully won't take another year to finish. :-)
Maybe as a recap you could open a issue on Cabrillo configs.

zcsahok:
@dl1jbe merge can wait, I'll try to complete the conversion in the coming days.

airween:
@dl1jbe Thomas,

> @airween From the discussion in your mentioned [thread](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00036.html) I got the impression that you started working on the feature. Any news on that?

not yet - I was confused as I wrote in my second e-mail. I think it would be good add these keywords that user can't place them everywhere, just to a predetermined location (eg. only to the rule file).

As you see my [summary](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00063.html), I listed few opinions there. I preferred (and still prefer) the 2/b, but as I wrote there, I think it can easily lead to the chaos. That's what I want to avoid, and the idea above (controls the source files of keywords), and I think Zoli's patch helps us in this.

dl1jbe:
> It's astyle playing with those comment lines. I was also wondering what is going on.

Ahh, that explains it. I already saw sometimes that astyle does not like heavy intended comments.
> 
> The process is: create a test for a keyword, add new style parsing and then mark it with NULL in `commands[]`. Old code is not touched and kept for easier fallback during testing. Once all keywords have been migrated, then I'll remove the legacy code incl. the mega-switch in `parse_logcfg()`.

Makes sense. I assume you will try to fix the remaining keywords in next days/weeks. So it will not too long until we drop it.
> 
> Ideally all vars shall come from a .h file. `globalvars.h` is probably the place for the ones defined in `main.c`, whereas other .c's provide their own .h. I was just blindly moving `extern`s out of `parse_logcfg()` when they were used. I think cleaning this up - and removing random externs generally - will be a next step.
> 
Ok, fine by me. I tried to move global varaible declarations belonging to a special part of code the according header file and all other into globalvar.h. It would be good to reduce the amount of globals as much as possible.

> Yes, of course, please squash the commits.
> 
OK.

> @airween finally I got a free slot for this topic. hopefully won't take another year to finish. :-)
> Maybe as a recap you could open a issue on Cabrillo configs.

dl1jbe:
> @dl1jbe merge can wait, I'll try to complete the conversion in the coming days.

OK. I will wait until then.

dl1jbe:
@airween

> @dl1jbe Thomas,
> 
> > @airween From the discussion in your mentioned [thread](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00036.html) I got the impression that you started working on the feature. Any news on that?
> 
> not yet - I was confused as I wrote in my second e-mail. I think it would be good add these keywords that user can't place them everywhere, just to a predetermined location (eg. only to the rule file).
> 
> As you see my [summary](https://lists.nongnu.org/archive/html/tlf-devel/2019-11/msg00063.html), I listed few opinions there. I preferred (and still prefer) the 2/b, but as I wrote there, I think it can easily lead to the chaos. That's what I want to avoid, and the idea above (controls the source files of keywords), and I think Zoli's patch helps us in this.

Ok, thanks for the clarification.

There is an additional thing we should keep in mind:
The code at the moment will parse the rules file AFTER completely reading the logcfg.dat. So we should declare the cabrillo entries only in one place.

zcsahok:
I'd stop parsing on the first error and exit. Is that OK? Currently there is option to continue, but IMO user shall first fix the config.

zcsahok:
For the parameter value: I'd strip trailing spaces and apply `g_shell_unquote` so that they could be specified if ever needed. Currently only leading space is stripped and each parser has to cope with trailing space/NL/CR.

dl1jbe:
> I'd stop parsing on the first error and exit. Is that OK? Currently there is an option to continue, but IMO user shall first fix the config.

I think there was a reason for that option, but I do not remember it atm. Just change it to exit as suggested. Just make sure the user has a chance to see the problem.

dl1jbe:
> For the parameter value: I'd strip trailing spaces and apply `g_shell_unquote` so that they could be specified if ever needed. Currently only leading space is stripped and each parser has to cope with trailing space/NL/CR.

Be careful in doing so. It was critical in some places to drop the trailing whitespace (especially the NL at the end of the call). Last time I had hands on it it was easier to keep it as is. Maybe we changed enough in the use of the parameters to allow the change now.

zcsahok:
So now for any reason `databases_load()` would fail, the program exits after user has pressed a key.

image

I think this version is good for merge. I'll continue working on fixing the issues marked (various buffer overflows, etc.), moving `extern`s to header files and the param stripping case. (maybe no in this order)

airween:
> Ok, thanks for the clarification.

yw.

If Zoli will finish the new parser and the code will merged, I'll try to add my idea: I should add a new sub-parser function, eg. `parse_cabrillo_template()`, which will called if the main parser founds the keyword `CABRILLO_TEMPLATE=/path/to/file`.

This function will search keywords with prefix `CBRTPL_`, here are few examples:
```
CBRTPL_CALLSIGN=HA2OS
CBRTPL_CATEGORY-ASSISTED=NON-ASSISTED
CBRTPL_CATEGORY-BAND=ALL
CBRTPL_CATEGORY-MODE=CW
CBRTPL_CATEGORY-OPERATOR=SINGLE-OP
``` 
Every lines will be put into the final cabrillo after the parser transforms it:
```
CALLSIGN: HA2OS
CATEGORY-ASSISTED: NON-ASSISTED
CATEGORY-BAND: ALL
CATEGORY-MODE: CW
CATEGORY-OPERATOR: SINGLE-OP
```

With this method, we can avoid the confusion: these keywords won't be part of generic config, but user can control the output, and can re-use its own templates as many times he/she wants.

Questions: do we really need to add the keywords with prefix? May be we can put the original cabrillo text, and the code just copy it as is....? Any idea?

dl1jbe:
> I think this version is good for merge. I'll continue working on fixing the issues marked (various buffer overflows, etc.), moving `extern`s to header files and the param stripping case. (maybe no in this order)

Applied, squashed to some degree and pushed. Also included the following commit 'fix tests' f49d2b0.
Thanks.

dl1jbe:
The merge conflict can be resolved by just rebasing  (or cherry-picking) the last 3 commits onto actual master.

zcsahok:
I'll just create another PR with the outstanding changes around lancode. (removed bogus variable `node` and the local index to band string mapping, added test for it)

dl1jbe:
All picked up and merged now. Closing.

PR #197, opened 2020-12-12
Introduce table driven contest configuration

dl1jbe:
Refactor 'setcontest()' to use predefined contest configurations. It replaces much of the variables to indicate the selected contest. 

The provided code should act as a starting point which introduces the mechanism. We can migrate in next time all contest related settings (scoring rules for points and multis, display variations, band timers and so on) to the  configuration struct and provide predefined settings for all supported contests.

For contests which are not hard coded  'config_unknow' is used which can be filled out by the keywords which were  already there (DX_&_SECTIONS, SSBPOINTS, ...).

I see several improvements with the new scheme:

- Contests are now mutually exclusive (No longer fear that 'cqww' and 'arrlss' are set simultaneously).

- It makes adding new contests easier - provide a configuration and maybe some helper functions which can be pointed to by the config).

- If needed we can provide dedicated functions for scoring, checking and so on and call them via function pointers in the config. That means the default functions can be simplified as special cases would be split out into these special functions.

As the code is quite fresh and contains quite some changes I would invite you to a thorough review. Comments and suggestions are welcome. 

There is no need to rush to merge it.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -86,7 +86,7 @@ int read_rules() {
     } else {
 	showstring("There is no contest rules file:", contest_conf);
 	showmsg("Assuming regular QSO operation. Logfile is qso.log");
-	setcontest("qso");		 /* default use general qso mode...
+	setcontest(QSO_MODE);		 /* default use general qso mode...
 					   (PA0R, 24 Sept. 2003) */
 	strcpy(logfile, "qso.log");
     }
--- a/src/rules.c
+++ b/src/rules.c
@@ -86,9 +86,8 @@ int read_rules() {
     } else {
 	showstring("There is no contest rules file:", contest_conf);
 	showmsg("Assuming regular QSO operation. Logfile is qso.log");
-	strcpy(whichcontest, "qso");	/* default use general qso mode...
+	setcontest("qso");		 /* default use general qso mode...
 					   (PA0R, 24 Sept. 2003) */
-	setcontest();
 	strcpy(logfile, "qso.log");
     }
 
--- a/src/setcontest.c
+++ b/src/setcontest.c
@@ -33,15 +33,45 @@
 #include "tlf.h"
 
 /* configurations for supported contest */
+contest_config_t config_unknown = {
+    .id = UNKNOWN,
+    .name = "Unknown"
+};
+
 contest_config_t config_qso = {
     .id = QSO,
     .name = "QSO"
 };
 
+contest_config_t config_cqww = {
+    .id = CQWW,
+    .name = "CQWW"
+};
+
+contest_config_t config_wpx = {
+    .id = WPX,
+    .name = "WPX"
+};
+
+
 /* table with pointers to all supported contests */
 contest_config_t *contest_configs[] = {
     &config_qso,
+    &config_cqww,
 };
+
+#define NR_CONTESTS (sizeof(contest_configs)/sizeof(contest_config_t*))
+
+/* lookup contest config by name in config table */
+contest_config_t *lookup_contest(char *name) {
+    for (int i = 0; i < NR_CONTESTS; i++) {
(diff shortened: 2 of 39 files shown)


src/rules.c
```diff
@@ -86,9 +86,8 @@ int read_rules() {
     } else {
 	showstring("There is no contest rules file:", contest_conf);
 	showmsg("Assuming regular QSO operation. Logfile is qso.log");
-	strcpy(whichcontest, "qso");	/* default use general qso mode...
+	setcontest("qso");		 /* default use general qso mode...
```
zcsahok, line 89: "qso" is in fact config_qso.name

test/test_contest.c
```diff
@@ -0,0 +1,56 @@
...
+    return 42;
+}
+
+contest_config_t config_focm;
+
+int setup_default(void **state) {
+    char filename[100];
+
+    strcpy(filename, TOP_SRCDIR);
+    strcat(filename, "/share/cty.dat");
```
zcsahok, line 23: could be written as
```
static char filename[] = TOP_SRCDIR "/share/cty.dat"
```
(to avoid eventual buffer overflow)

src/tlf.h
```diff
@@ -187,5 +187,31 @@ void refreshp();
...
+    SPRINT,
+    ARRLDX_USA,
+    ARRLDX_DX,
+    ARRL_SS,
+    ARRL_FD,
+    PACC_PA,
+    STEWPERRY,
+    FOCMARATHON,
+    OTHER,
+    UNIVERSAL,
```
zcsahok, line 204: OTHER, UNIVERSAL are currently not used

zcsahok (approved):
fine with me

src/rules.c
dl1jbe: Right. The idea behind setcontest(char *name) is to activate the contest (in these case the non-contest qso-mode) by name.

test/test_contest.c
dl1jbe: Good idea. See new commit.

src/setcontest.c
```diff
@@ -32,24 +32,107 @@
...
+
+#define NR_CONTESTS (sizeof(contest_configs)/sizeof(contest_config_t*))
+
+/** lookup contest config by name in config table
+ *
+ * ignore configs where .name is not set
+ */
+contest_config_t *lookup_contest(char *name) {
+    for (int i = 0; i < NR_CONTESTS; i++) {
+	if (contest_configs[i]->name != NULL) {
```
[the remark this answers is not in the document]
dl1jbe: I added the check to make code more robust. In case someone add a new contest_configuration we have no guarantee that she not leave out the name.

dl1jbe:
Thanks for the feedback. I will adapt to changed master code to resolve the conflicts and afterwards pull it in if no other problems gets found.

src/rules.c
zcsahok: I mean the string constant could be replaced by a reference to the actual source of definition.

dl1jbe: Yes, I understood. It seemed only weird to me as we lookup configurations by name and then use that waht we would fine to look itself up.

But maybe you are right. Atm the best it seems to define a symbol and use it in both places - as a name for the configuration and as a name to look it up

#define QSO_MODE "qso"

config_qso = { .name = QSO_MODE }  in setcontest.c

and finally

setcontest(QSO_MODE)                         in rules.c

N0NB:
Hi Tom.

I've not taken the time to try this but I really like the concept.  I see that it is merged so I may get a chance to look at it after mid January.  Merry Christmas!

dl1jbe:
Hi Nate,

thanks for the info. And i hope you had some good Christmas days as well.

PR #199, opened 2020-12-17
fix and test lancode

zcsahok:
The rest of #196 

- removed variable `node`
- use `band[]` is lancode instead of local list
- added test for lancode (covers only sendfreq)
- fix 60m in `test/data.c`
- use common macro to free a dynamic string

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -33,6 +33,7 @@
 
 #include "err_utils.h"
 #include "get_time.h"
+#include "globalvars.h"
 #include "lancode.h"
 #include "tlf.h"
 #include "tlf_curses.h"
@@ -62,7 +63,6 @@ char bc_hostservice[MAXNODES][16] = {
 };
 char sendbuffer[256];
 int nodes = 0;
-int node;
 int send_error_limit[MAXNODES];
 //--------------------------------------
 /* default port to listen for incomming packets and to send packet to */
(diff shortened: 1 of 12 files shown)


dl1jbe:
Any reason you let PR #196 open?

The new macro FREE_DYNAMIC_STRING introduces some new warnings if used with 'const char *', see the below:

> test_lancode.c:31:25: Warnung: Übergabe des Arguments 1 von »g_free« entfernt Kennzeichner »const« von Zeiger-Ziel-Typ [-Wdiscarded-qualifiers]
>   31 |     FREE_DYNAMIC_STRING(sendto_last_message);
>        |                                           ^~~~~~~~~~~~~~~~~~~
> ../src/tlf.h:190:56: Anmerkung: in Definition des Makros »FREE_DYNAMIC_STRING«
  190 | #define FREE_DYNAMIC_STRING(p)  if (p != NULL) {g_free(p); p = NULL;}

Hope you can cope with the german comment.

dl1jbe:
Fine, looks good now. Thanks. 

Applied and pushed.

zcsahok:
#196 can be closed. I though it better you do it as owner.

Committed fix for the warning.

dl1jbe:
> #196 can be closed. I though it better you do it as owner.
> 
Ok. It was not quite clear to me if you had plans to add further commits there. I had expected that you would close the PR if there is nothing more to pick up (as you moved the newer commits to the new PR).

> Committed fix for the warning.

The fix got merged with the rest in the master branch.

PR #200, opened 2020-12-21
Use Hamlib utility functions

zcsahok:
As we link against Hamlib anyways, we can use it's utility functions `locator2longlat()` and `qrb()`. Current code contains a local copy of these.

Also included previously added new data files to dist. (`test/Makefile.am`)


dl1jbe:
Looks good. Thanks.

PR #201, opened 2020-12-22
Drop numeric tncport

zcsahok:
Keep only fully qualified `tncportname` to specify the TNC port. Numeric `tncport` was only partially supported (for values 1 and 2, legacy DOS COM1 and COM2). Opening a port was reporting wrong text using `tncport` in case of using `tncportname`. IMO anyone using a TNC shall be able to specify the correct Linux port name.

The same code pops up in `debug_tty()` (`changepars.c`), but there `tncport` is an unused local variable. Cleaned this up, but I suggest to completely drop `debug_tty()` as it assumes an ASCII rig protocol and thus has very limited use. One shall instead serve of either Hamlib's utils or any serial terminal emulator.

### DIFF
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -894,7 +894,6 @@ int debug_tty(void) {
     extern int serial_rate;
 
     int fdSertnc;
-    int tncport = 0;
     int i;
     struct termios termattribs;
     char line[20] = "?AF\015";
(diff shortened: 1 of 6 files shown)


dl1jbe:
Looks good. I agree that debug_tty() should be drop without replacement. 
Can you add a second commit with that. I will then pick up the PR afterwards. Thanks

zcsahok:
Done.
(now adding the almost 100 missing externs to globalvars.h...)

dl1jbe:
Argh. Tried to add the comments from the PR to the commit but failed in doing so. Should be better next time.

Thanks.

PR #202, opened 2020-12-23
Update globalvars.h

zcsahok:
Moved all externs from `parse_logcg.c` to their respective `.h` files, mostly `globalvars.h`.
On mid term all other `.c` files shall be cleaned up. Additionally ints with a boolean semantics shall be converted to bools.

### DIFF
--- a/src/getexchange.h
+++ b/src/getexchange.h
@@ -21,6 +21,8 @@
 #ifndef GETEXCHANGE_H
 #define GETEXCHANGE_H
 
+extern int call_update;
+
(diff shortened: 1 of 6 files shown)


dl1jbe:
Any more to come here?

I agree that both goals (cleanup and move to bool) are on the agenda.

zcsahok:
It's all for now, I was focusing on `parse_logcfg`. Further update can be done when a file is touched anyway.

dl1jbe:
Thanks. Applied and pushed

PR #204, opened 2020-12-26
Clean lancode

dl1jbe:
The PR summarizes some work to cleanup the code

- drop unused variables, move others to local scope, drop dead code
- do not count inactive lan as errors during function return
- cleanup wrong error counting
- finally make lan_active bool and adapt all  tests of the variable

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -58,7 +58,7 @@ char default_lan_service[16] = "6788";
 /* lan port parsed from config */
 int lan_port = 6788;
 
-int lan_active = 0;
+bool lan_active = false;
 int send_error[MAXNODES];
 int send_error_limit[MAXNODES];
 int lan_mutex = 0;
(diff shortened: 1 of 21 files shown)


src/lancode.c
```diff
@@ -105,58 +92,66 @@ int resolveService(const char *service) {
...
+    struct sockaddr_in lan_sin;
+
+    if (!lan_active)
+	return 0;
 
     sprintf(default_lan_service, "%d", lan_port);
     bzero(&lan_sin, sizeof(lan_sin));
     lan_sin.sin_family = AF_INET;
     lan_sin.sin_addr.s_addr = htonl(INADDR_ANY);
     lan_sin.sin_port = htons(resolveService(default_lan_service));
```
zcsahok, line 107: just a side note: I find using `resolveService()` (i.e. `getservbyname()`) an overkill as no-one will use a non-privileged UDP service (port) by its name for intra-Tlf communication.
In other IP-related cases we just use plain numeric port, e.g.
```
address.sin_port = htons(netkeyer_port);
```

zcsahok (commented):
pse merge, I have another PR in the queue adding constants for rigptt bits and cleaning externs from a bunch of files.

src/lancode.c
dl1jbe: You are right. It just got inherited from Rein. We should clean it out in future work.

PR #205, opened 2020-12-26
Add enums for rigptt, removed externs, added QRA evaluation

zcsahok:
The main change is adding enums for `rigptt` bits. Changed its type to `int` to simplify setting it in parse_logcfg, also dropped the ORing as other bits are not used at this stage.

Boolified variables from main that are controlled by command line switches (debugflag, no_trx_control, verbose, convert_cabrillo, nopacket). Also converted trx_control.

Replaced opening voice keyer wav file by `access()` in `play_file()`. This function was already used to check the presence of `./play_wk`, so the code is more consistent now.

Removed externs from all affected files, added variable to header files (globalcfg + lancode).
Removed the deprecated keyer enums. These are only internally used, so renumbered them.

Added a feature: use specfied QRA locator to determine lan/long.

Found an issue in `networkinfo()`: the reported config file is always either `logcfg.dat` or the given on command line. The system config is never shown as `config_file` is never empty. In addition the system config location is hard coded.
This can be fixed, but then we'll have an issue with re-reading the config via `:SET`/`:CFG`. Btw these probably anyway do not work as proper re-reading requires releasing all resources (net sockets, ...) which is not done at the moment.

### DIFF
--- a/share/help.txt
+++ b/share/help.txt
@@ -86,7 +86,6 @@ $         Switch to MEM and clear it (pop)
 :INFo     Network status
 :TRXcont  Toggle TRX control on/off
 :CHAR     Autosend: number of chars before starting to send
-:DEBug_tty Test rig link
 :SOUnd    Record sound files
 :WRIte    Write cabrillo
 :ADIf     Write Adif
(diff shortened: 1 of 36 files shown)


dl1jbe:
Applied an pushed.

Quite a bunch of work. Very appreciated.

I did a little resorting (adding comments from the PR to commit 1, merged commit 3..5) before pushing to the tree.

Moved the comments about networkinfo() to issue #206.

PR #207, opened 2020-12-27
Make use of contest config

dl1jbe:
Added a command line switch (-l) to list all contests with builtin support.
Moved recall_mult and exchange_serial varaible to contest_config.
Point scoring is now controlled by the contest configuration in one of two variants
* have a fixed point scoring or
* use a contest related function to determine the points.

Further work will migrate the control of displayed scoring results (showscore() ), the rules for multi recognition and exchange checking.


src/tlf.h
```diff
@@ -205,14 +205,27 @@ typedef enum {
 
 #define QSO_MODE ("qso")
 
+typedef enum {
+    DEFAULT = 0,	/* undiefined -> DEFAULT */
```
zcsahok, line 209: typo in comment

dl1jbe: Thanks for pointing it out.

Issue #203, opened 2020-12-26
Node sync/deleting QSO


m5evt:
I am running 2 instances of tlf on the same machine over LAN. I can confirm that communication between the 2 nodes is occurring and I can see a QSO being added to the other node when it takes place. However, I notice that deleting a QSO in 1 node does not delete it in the other. This results in scores drifting apart. The only way I can fix this is manually deleting the QSO in all nodes.

I shutdown one instance of tlf, added some QSOs in the other node. Then I re-opened the TLF instance again. A SYNC command (":sync") did not pick up these new QSOs that occurred while tlf was "off". Am I missing something regarding behaviour of the LAN/node implementation?

dl1jbe:
Thanks for the report.

ATM only adding QSOs to the log is supported. Normally in contest you would rather not delete a qso but make a note to process it later. So until now there was no need to support a delete also. 
Maybe we should add that too. Can you give some information what you wanted to accomplish?

An automatic sync at startup was considered to difficult to implement. So Rein, the original author of TLF, decided to provide then :SYNC command only. There were not so much need for another solution in last years as not many people seems to use TLF for M/S or M/M operation.
Besides that :SYNC should work. Only problem is that you have to provide the logfile via FTP://.

zcsahok:
@m5evt what is your `SYNCFILE` setting? (logcfg.dat)

m5evt:
Thanks for the reply. There are 2 things I am trying to accomplish that are linked. I have done some digging and I think I was missing the SYNCFILE. I will give this a try.

I have been experimenting with mapping in Python and tlf log files. I have managed to map worked cqww zones on a map using cartopy in Python.  I've also managed this for US states too. For now, this means I have to read/process the tlf log file each time I want to update the map. I had an idea that I could make the Python script emulate as a tlf node and then update the map only when a new zone is worked. Hence, I have been trying to understand the LAN functions in tlf. Below are some of the maps I have produced.

![All](https://user-images.githubusercontent.com/60316466/103155446-01755580-4798-11eb-9408-9018167e93f3.png)

![z14](https://user-images.githubusercontent.com/60316466/103155440-e99dd180-4797-11eb-99e4-54068595766b.png)

[...]

PR #208, opened 2020-12-30
Fix payload in lan_send()

zcsahok:
When communicating with other TLF nodes `lan_send()` in `lancode.c` sends a fixed length, 256 bytes UDP payload.

Apart from wasting bandwidth the issue is that this payload originates from a smaller buffer
```
int send_lan_message(int opcode, char *message) {
    char sendbuffer[102];
```
which can potentially lead to reading from unallocated memory thus leading to SEGV and crasing TLF.

Here how it looks like when sending a talk message (actual payload is `A4hello\n`):
Screenshot_20201230_102351
There is a bunch of garabge from the stack after the payload.

After the fix:
Screenshot_20201230_103946

(Note the dreaded trailing NL. We have to iron it out later.)

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -249,7 +249,7 @@ static int lan_send(char *lanbuffer) {
     for (int node = 0; node < nodes; node++) {
 
 	bc_sendto_rc = sendto(bc_socket_descriptor[node],
-			      lanbuffer, 256,
+			      lanbuffer, strlen(lanbuffer),
 			      0, (struct sockaddr *) &bc_address[node],
 			      sizeof(bc_address[node]));
 


zcsahok:
BTW how about dropping time sync functionality?
I think nowadays any reasonable computer runs some kind of time sync SW (NTP,...) anyway. So this shouldn't be the task of TLF.

dl1jbe:
> BTW how about dropping time sync functionality?
> I think nowadays any reasonable computer runs some kind of time sync SW (NTP,...) anyway. So this shouldn't be the task of TLF.

Is that also true for a field day installation? I can imagine installations without any access to internet (or maybe only one station having a connection). NTP is fine but useless as long as there is no NTP server around. Alternatively  you have to establish port forwarding on the one station with network access.

A greater problem in these field seems to be the algorithm for giving out qso numbers for distributed stations (as well as the chance to drop a logged QSO - see dicussion in issue #203).  I can not remember any question to a true MS/MM contest setup in the last years. So maybe not so much people are using these mode at all.

zcsahok:
When working in FD with multiple computers without Internet one can set up one of them as a local NTP server. I think the clock on modern PCs/laptops does not drift considerably within a day or two.
But one can sync time with other non-Internet resources: GPS or RF clocks (like DCF). Here is a nice project: https://weberblog.net/ntp-server-via-dcf77-on-a-raspberry-pi/

The lan code part has to be definitely reviewed. Not sure why time sync would be needed. I also doubt that current code works correctly when running >2 instances at a high QSO rate due to lack of central serial number management. But that's another story.

airween:
> BTW how about dropping time sync functionality?
> I think nowadays any reasonable computer runs some kind of time sync SW (NTP,...) anyway. So this shouldn't be the task of TLF.

Just a quick note: in case of N1MM, if there are more stations (MULTI), there must be a "master" station, which gives the time sync. That can get the time through NTP, but none of the other stations.

Btw I should keep the time sync between nodes.

dl1jbe:
Clock drift should not be the problem today. You can setup an NTP server any time, but if we drop the time sync by TLF we should at least make clear that you have to care for the correct time yourself.

> I also doubt that current code works correctly when running >2 instances at a high QSO rate due to lack of central serial number management.

That is just what I had in mind with my above comment.

dl1jbe:
Back to the original PR. Good catch. Will apply shortly.

The fear for an SEGV seems only valid to me if someone else will call lan_send(). Actual the used buffer is in the stack frame of send_lan_message(). The greater problem may be that we leak information about the parts in the stack above the buffer. Anyway, the fix is needed and appreciated.

dl1jbe:
Applied an pushed

Issue #198, opened 2020-12-15
Add a way of specifying Cabrillo header values


zcsahok:
Currently Cabrillo header values (CONTEST, CATEGORY-*, ...) have to be entered online while exporting the log. This process is quite tedious and error prone.

As a user I expect a way to specify such mostly seldom changing infos both
* via config file (logcfg.dat / rule file)
* and by reusing previous or sample logs as a template.

The procesing of configuration shall fit into the current Cabrillo QSO format specification logic (keyword CABRILLO).

airween:
To anybody: a possible solution was introduced [here](https://github.com/Tlf/tlf/pull/196#issuecomment-745111450).

zcsahok:
It would be nice to have a way of setting mostly constant (or default) data in logcfg+rule, i.e. name, email, power, etc as
```
CABRILLO-NAME: Joe Doe
CABRILLO-EMAIL: j.doe@...
CABRILLO-CATEGORY-POWER: LOW
```

`CABRILLO-TEMPLATE` could point to an actual Cabrillo file from which all relevant fields  (excluding QSO, START/END, CREATED) could be taken over. Ideally a valid file should be usable as-is in order to be user friendly: just point to an example log and that's it.

Missing fields would be still asked for during Cabrillo creation.

`CABRILLO-TEMPLATE` would be logically part of rules. But even without a template one could describe the fields using `CABRILLO-*` configs.

The current `CABRILLO` QSO format spec could be also optionally merged to the template by adding the same kind of "QSO=" line manually. This would result in a compact spec at the expense of possibly duplicating formats per contest. (or QSO= could be even a reference to the format...)

E.g.
```
START-OF-LOG: 3.0  <- not used
CREATED-BY: tlf-1.3.0 <- not used
CONTEST: 9ACW
#CATEGORY-BAND: 80M ... commented out
CATEGORY-MODE: CW
CATEGORY-OPERATOR: SINGLE
CATEGORY-POWER: LOW
QSO=<FMT spec>

```

zcsahok:
_Further brainstorming..._
Tags fall into these categories -
* **ask**, with optional hint text in parentheses (or brackets?)
	`CABRILLO-CATEGORY-POWER=`
	or
	`CABRILLO-CATEGORY-POWER= (LOW,HIGH)` or `[LOW, HIGH]`
        Hint text would be shown when asking for a value.

* **fixed**, never or seldom changing value
	`CABRILLO-NAME= Joe Doe`
[...]

PR #209, opened 2021-01-03
Cabrillo improvements

zcsahok:
A lot of changes, see #198.

- Cabrillo header fields configurable via configs and templates
- Better code sharing for ADIF and Cabrillo generation
- EXCHANGE-SEPARATOR for splitting received exchanges into Cabrillo QSO fields

Still open is the man page update for the Cabrillo header specification.


dl1jbe:
Applied and pushed. Good start in the new year. Thanks and HNY.

PR #213, opened 2021-01-08
Fixed buffer overflow error when typed longer comment than at last qso

airween:
I ran into a problem in QSO mode when I typed a longer text in the comment field than 12 chars, and the call exists in list of previous contacts (aka `worked` list).

First, the available space in the WORKED window is 12 char wide, but the declaration contains only `char[12]`, so I incremented that value to 13.

Then I changed the `strcpy` by the `strncpy`, and set the max length. Finally, I set to `\0` at the max length position in the `worked[i].exchange`.

### DIFF
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -93,7 +93,7 @@ int addcall(void) {
     j = getctydata(hiscall);
     worked[i].country = j;
     if (strlen(comment) >= 1) {		/* remember last exchange */
-	g_strlcpy(worked[i].exchange, comment, 12);
+	g_strlcpy(worked[i].exchange, comment, sizeof(worked[i].exchange));
 
 	if (CONTEST_IS(CQWW) || wazmult == 1 || itumult == 1) {
 	    /*
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -93,7 +93,8 @@ int addcall(void) {
     j = getctydata(hiscall);
     worked[i].country = j;
     if (strlen(comment) >= 1) {		/* remember last exchange */
-	strcpy(worked[i].exchange, comment);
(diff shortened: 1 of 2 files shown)


src/addcall.c
```diff
@@ -93,7 +93,8 @@ int addcall(void) {
     j = getctydata(hiscall);
     worked[i].country = j;
     if (strlen(comment) >= 1) {		/* remember last exchange */
-	strcpy(worked[i].exchange, comment);
+	strncpy(worked[i].exchange, comment, 12);
```
zcsahok, line 96:
```
   g_strlcpy(worked[i].exchange, comment, sizeof(worked[i].exchange));
```
can replace both lines

zcsahok (commented):
glibc's string functions are more safe

src/addcall.c
airween: Thanks, fixed.

src/addcall.c
```diff
@@ -93,8 +93,7 @@ int addcall(void) {
     j = getctydata(hiscall);
     worked[i].country = j;
     if (strlen(comment) >= 1) {		/* remember last exchange */
-	strncpy(worked[i].exchange, comment, 12);
-	worked[i].exchange[12] = 0;
+	g_strlcpy(worked[i].exchange, comment, 12);
```
dl1jbe, line 96: The last parameter should be 13 or better sizeof(worked[i].exchange). 

g_strlcpy needs the size of the buffer, not the number bytes you want to copy.

airween: TNX, fixed.

I always forget that glib functions can be used here :)

dl1jbe:
Thanks. Applied and pushed.

PR #214, opened 2021-01-09
Minor improvement of exist_in_country_list() and score.c

zcsahok:
See my comment on #212 

- changed prefix match retries to a loop
  * BUT: the whole thing makes no sense as the actually called `country_found()` does not use its argument. To be fixed later on.
- increased the size of `prefix` to match the definition `char pxstr[11]`
  * use of global variable for implicit argument shall be fixed here too.

Additionally
- changed `portable_x2` to bool
- suffix match in `portable_doubles()` uses the more clear `g_str_has_suffix()`
- no point using `strncpy` in `calc_continent()`, changed to plain `strcpy`
- removed externs, also from test
- corrected typo in "weight"

### DIFF
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -170,7 +170,7 @@ extern int logfrequency;
 extern int bmautoadd;
 extern int bmautograb;
 extern int serial_or_section;
-extern int portable_x2;
+extern bool portable_x2;
 extern int clusterlog;
 extern int sprint_mode;
 extern int timeoffset;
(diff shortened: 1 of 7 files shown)


dl1jbe:
>     * changed prefix match retries to a loop
>       
>       * BUT: the whole thing makes no sense as the actually called `country_found()` does not use its argument. To be fixed later on.

Right.  We discussed that more than once but got distracted by more urgent stuff. It is overdue.

 
>     * increased the size of `prefix` to match the definition `char pxstr[11]`

Good catch.

PR #215, opened 2021-01-09
Change handling of config and rule files

dl1jbe:
* Copy over default logcfg.dat if no one found in actual directory.
    
    Copy the default config file into the actual directory if no config are given
    on command line AND no logcfg.dat are already found.
    Furthermore close TLF with an error message if the given file is not found.

* Add helper function to find in local or default directory


zcsahok:
We could also review the usage of the `.paras` file. It looks like a leftover from an early version. Currently is just written in `writeparas.c` but not read anywhere. The only use it to detect the first run of Tlf. But for this we could you the copying over of `logcfg.dat`.

src/parse_logcfg.c
```diff
@@ -66,19 +68,31 @@ int read_logcfg(void) {
...
+	    return PARSE_ERROR;
+	}
+    }else {
 	config_file = g_strdup(LOGCFG_DAT_FILE);
 
-    if ((fp = fopen(config_file, "r")) == NULL) {
-	if ((fp = fopen(defltconf, "r")) == NULL) {
+	if (access(config_file, R_OK) == -1) {
+	    showmsg("No logcfg.dat found. Copying default config file.");
+	    showmsg("Please leave and adapt your settings if needed!");
```
zcsahok, line 82: does "leave" here mean to exit Tlf? maybe just "Please adapt the settings to your needs." would be enough.

dl1jbe:
> We could also review the usage of the `.paras` file. It looks like a leftover from an early version. 

Right. It was already unused in tlf-0.9.10 from 2003 (the oldest version I could find on the net. Found it mentioned in old ChangeLog which tells that generation of .paras was added in 5.2.3

I think it got never used.

> The only use it to detect the first run of Tlf. But for this we could do when the copying over of `logcfg.dat`.
> The only problem is that checking of existence of `.paras` happens quite early in the code. But maybe it could be somehow reordered.

Second problem is that it may be you have an 'logcfg.dat' already prepared and there is no copying over.
Better indication would be a missing logfile. But for this we need to read 'logcfg.dat' and rules file before.

So how about an (empty) dotfile which gets written on first start?

zcsahok:
> So how about an (empty) dotfile which gets written on first start?

Probably that's the easiest and still compatible solution. `isFirstStart()` would check `.paras` and if it is not present then create an empty file.

`writeparas.c` can be then disposed of and `getmessages()` can be renamed to what it actually does (~`show_station_info()`). Not sure about `setparameters()`: the changes made there are not persisted (they were meant to be persisted in .paras, I guess).

dl1jbe:
> `isFirstStart()` would check `.paras` and if it is not present then create an empty file.

Maybe we find a better name, but until then lets stay with `.paras`.

> Not sure about `setparameters()`: the changes made there are not persisted (they were meant to be persisted in .paras, I guess).

I think that was the original. plan.

I will try to look into it in the evening.

dl1jbe:
Looking into it - `setparameters()` gets never called. Will drop it completely.

zcsahok:
Just another point regarding the first run feature: `.paras` is checked in the current directory, so if tlf is started in another dir then user gets a welcome again. Could be annoying.

This feature could be replaced by a command line option like '--licence'. (I made a quick check but none of curl, wget, gccm python3 has it)

Related is that inclusion of rules file happens from the current dir even if a config file was specified on command line. This could lead to strange effects: rule file not found or other rules are taken than expected. Maybe it would be more correct to open the rule from the base dir of the actual logcfg file + rules. Not sure, as the current option could also sometimes make sense.

dl1jbe:
For that problem we have to rethink how tlf gets used. Some people create a directory for each contest and start it there. Other people have only one directory with different logcfg's and a common rule directory.
It seems rather odd to mee to have the config in a completely different directory.

But to be honest the actual scheme really might not be the best one.

PR #218, opened 2021-01-17
Move inclusion of hamlib/rotator.h to C source

N0NB:
Try to avoid indirect includes as much as possible.


dl1jbe:
Makes sense. Thanks. Applied and pushed.

PR #220, opened 2021-01-17
Quell warnings from use of strncpy()

N0NB:
Poking around the Web I found:

https://github.com/libfuse/libfuse/commit/f39c71dcf99292c188bb6f0a117d7e118f92bfb1

Which states in part:

Recent GCC releases have warnings related to common strncpy(3) bugs.
These warnings can be avoided by explicitly NUL-terminating the buffer
or using memcpy(3) when the intention is to copy just the characters
without the NUL terminator.

The use of memcpy() in this context does quell the warnings but requires
testing to be sure the behavior has not changed.

The following warnings were quelled:

../../tlf/src/background_process.c: In function ‘background_process’:
../../tlf/src/background_process.c:186:4: warning: ‘strncpy’ output may be truncated copying 80 bytes from a string of length 253 [-Wstringop-truncation]
  186 |    strncpy(prmessage, lan_message + 2, 80);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../tlf/src/background_process.c:178:4: warning: ‘strncpy’ output may be truncated copying 80 bytes from a string of length 253 [-Wstringop-truncation]
  178 |    strncpy(prmessage, lan_message + 2, 80);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../tlf/src/clusterinfo.c: In function ‘loadbandmap’:
../../tlf/src/clusterinfo.c:211:6: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 12 [-Wstringop-truncation]
  211 |      strncpy(spottime, thisline + 70, 4); // how old?
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../tlf/src/log_to_disk.c: In function ‘log_to_disk’:
../../tlf/src/log_to_disk.c:80:2: warning: ‘strncpy’ output may be truncated copying 87 bytes from a string of length 253 [-Wstringop-truncation]
   80 |  strncpy(lan_logline, lan_message + 2, 87);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../tlf/src/scroll_log.c: In function ‘get_next_serial’:
../../tlf/src/scroll_log.c:50:2: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 65 [-Wstringop-truncation]
   50 |  strncpy(qsonrstr, logline4 + 23, 4);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../tlf/src/scroll_log.c:55:6: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 154 [-Wstringop-truncation]
   55 |      strncpy(qsonrstr, logline3 + 23, 4);
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../tlf/src/scroll_log.c:61:6: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 243 [-Wstringop-truncation]
   61 |      strncpy(qsonrstr, logline2 + 23, 4);
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../tlf/src/scroll_log.c:67:6: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 332 [-Wstringop-truncation]
   67 |      strncpy(qsonrstr, logline1 + 23, 4);
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -175,7 +175,7 @@ void *background_process(void *ptr) {
 			break;
 
 		    case CLUSTERMSG:
-			strncpy(prmessage, lan_message + 2, 80);
+			memcpy(prmessage, lan_message + 2, 80);
 			if (strstr(prmessage, my.call) != NULL) {	// alert for cluster messages
 			    TLF_LOG_INFO(prmessage);
 			}
@@ -183,7 +183,7 @@ void *background_process(void *ptr) {
 			addtext(prmessage);
 			break;
 		    case TLFSPOT:
-			strncpy(prmessage, lan_message + 2, 80);
+			memcpy(prmessage, lan_message + 2, 80);
 			lanspotflg = 1;
 			addtext(prmessage);
 			lanspotflg = 0;
@@ -248,4 +248,3 @@ void *background_process(void *ptr) {
     }
 
 }
-
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -176,6 +176,8 @@ void *background_process(void *ptr) {
 
 		    case CLUSTERMSG:
 			memcpy(prmessage, lan_message + 2, 80);
+			prmessage[80] = '\0';
+
 			if (strstr(prmessage, my.call) != NULL) {	// alert for cluster messages
 			    TLF_LOG_INFO(prmessage);
 			}
@@ -184,6 +186,7 @@ void *background_process(void *ptr) {
 			break;
 		    case TLFSPOT:
 			memcpy(prmessage, lan_message + 2, 80);
+			prmessage[80] = '\0';
 			lanspotflg = 1;
 			addtext(prmessage);
 			lanspotflg = 0;
--- a/src/clusterinfo.c
+++ b/src/clusterinfo.c
@@ -208,7 +208,7 @@ int loadbandmap(void) {
 
 	    g_strlcpy(spotcall, thisline + 26, 6);
 
-	    strncpy(spottime, thisline + 70, 4);	// how old?
+	    memcpy(spottime, thisline + 70, 4);	// how old?
 	    spottime[4] = spottime[3];
 	    spottime[3] = spottime[2];
 	    spottime[2] = ':';
@@ -418,5 +418,3 @@ int getclusterinfo(void) {
 
     return si;
 }
-
-
--- a/src/log_to_disk.c
+++ b/src/log_to_disk.c
@@ -78,6 +78,8 @@ void log_to_disk(int from_lan) {
     } else {			// qso from lan
 
 	memcpy(lan_logline, lan_message + 2, 87);
+	lan_logline[87] = '\0';
+
 	strcat(lan_logline, spaces(78));
 
 	if (cqwwm2 == 1) {
--- a/src/log_to_disk.c
+++ b/src/log_to_disk.c
@@ -77,7 +77,7 @@ void log_to_disk(int from_lan) {
 	cleanup_qso();		/* reset qso related parameters */
     } else {			// qso from lan
 
-	strncpy(lan_logline, lan_message + 2, 87);
+	memcpy(lan_logline, lan_message + 2, 87);
 	strcat(lan_logline, spaces(78));
 
 	if (cqwwm2 == 1) {
--- a/src/scroll_log.c
+++ b/src/scroll_log.c
@@ -48,23 +48,27 @@ void get_next_serial(void) {
 
     if (!log_is_comment(logline4)) {
 	memcpy(qsonrstr, logline4 + 23, 4);
+	qsonrstr[4] = '\0';
 	mm = atoi(qsonrstr);
     }
     if (!log_is_comment(logline3)) {
 	if (atoi(logline3 + 23) > mm) {
 	    memcpy(qsonrstr, logline3 + 23, 4);
+	    qsonrstr[4] = '\0';
 	    mm = atoi(qsonrstr);
 	}
     }
     if (!log_is_comment(logline2)) {
 	if (atoi(logline2 + 23) > mm) {
 	    memcpy(qsonrstr, logline2 + 23, 4);
+	    qsonrstr[4] = '\0';
 	    mm = atoi(qsonrstr);
 	}
     }
     if (!log_is_comment(logline1)) {
 	if (atoi(logline1 + 23) > mm) {
 	    memcpy(qsonrstr, logline1 + 23, 4);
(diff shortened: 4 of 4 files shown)


zcsahok:
While this quells the warning, I suggest using GLib's `g_strlcpy()` instead as it guarantees the dest to be nul-terminated (`memcpy` does not) and it is already used all around the (newer) code.

N0NB:
* On 2021 17 Jan 12:46 -0600, Zoltan Csahok wrote:
> While this quells the warning, I suggest using GLib's `g_strlcpy()`
> instead as it guarantees the dest to be nul-terminated (`memcpy` does
> not) and it is already used all around the (newer) code.

Did strncat nul terminate the string?  As I was looking at the code it
didn't seem obvious that the destination was being nul terminated.

73, Nate

-- 

Web: https://www.n0nb.us

N0NB:
There are also some places where strncat is used that are generating warnings.

dl1jbe:
Yes `strncat` guarantees a null-termination. 

Another way instead `g_strlcpy` is to make a dedicated copy with `g_strndup` but that has to be freed later.

zcsahok:
`strncpy` nul-terminates only if the first N chars of source contain a NUL. It's a documented feated, see for a test https://gist.github.com/zcsahok/8706fab5dfc16641ddbc2b70f6b39b8f .

On the other hand the warning is somewhat unclear to me.
```
../../tlf/src/scroll_log.c:67:6: warning: ‘strncpy’ output may be truncated copying 4 bytes from a string of length 332 [-Wstringop-truncation]
67 | strncpy(qsonrstr, logline1 + 23, 4);
```
We just want 4 bytes from offset 23, so what's the issue? Maybe it's about the case when the source (logline) is shorter and strncpy would truncate qsonrstr?

N0NB:
I agree, the warning text isn't clear as I would think copying 4 bytes should be safe in this context.  It's probably a generic warning string that gets spat out for these warnings by gcc.

N0NB:
Also, after poking a bit more, the destination strings should be nul terminated as they get handled by other functions so I'll close this in a couple of days unless otherwise requested and use `g_strlcpy()` as we just as well use glib functions since they're in use elsewhere.

N0NB:
Well, right away I see an issue with `g_strlcpy ()` and that is that the third parameter is the declared size of the destination string and not the number of bytes to copy.

`g_strndup ()` allows for specifying the number of bytes and then the code would be something like:

```
char * spottime;
.
.
.
spottime = g_strndup (thisline + 70, 4);
.
.
.
g_free (spottime);
```

The other option, as we know our buffer is large enough is to manually assign the nul terminator ourselves after using `memcpy()`:

```
memcpy(spottime, thisline + 70, 4);
spottime[4] = '\0';
```

Comments?

dl1jbe:
Look for `-Wstringop-truncation` in `man gcc`.  It discusses some of the problems they try to catch.
Interestingly the switch changed to `-Wno-stringop-truncation` in GCC 10.

I think the main problem is the lack of a proper `substring` operator in C string handling.

N0NB:
I see that `spottime` was a poor example as it eventually becomes nul terminated.

Red face...

N0NB:
Hi All.

I added another commit that makes sure all of the memcpy'd strings are nul terminated.

Tom, if this is acceptable, feel free to rebase them together when applying them to master.

N0NB:
> Look for `-Wstringop-truncation` in `man gcc`. It discusses some of the problems they try to catch.
> Interestingly the switch changed to `-Wno-stringop-truncation` in GCC 10.

I guess I understand what the GCC people are trying to accomplish but in our case and from my reading of the `strncat()` man page, I don't see an issue.  Their suggested way quelling the warning through an attribute does not seem applicable in this case.

> I think the main problem is the lack of a proper `substring` operator in C string handling.

The C library has long been taken to task for many naive string handling functions.

N0NB:
From [Stackoverflow there is this ugliness](https://stackoverflow.com/a/50198710) (example from `background_process.c`):

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-truncation"
			strncat(talkarray[4], lan_message + 2, 60);
#pragma GCC diagnostic pop
```

How an older GCC would handle this is questionable.

I certainly do not wish to turn off such warnings globally as doing so shortcuts any new code where a mistake is made and the warning is needed.  OTOH, quelling the warning through this mechanism is fine so long as we've carefully vetted the code in question and the buffers like `talkarray[4]` are never shortened to less than the total number of concatenated characters.

Of course, a long unwritten rule of C is, "Thou shalt not shorten the declarations of thy buffers", and if one does so, it has be carefully and thoroughly vetted.

dl1jbe:
> From [Stackoverflow there is this ugliness](https://stackoverflow.com/a/50198710) (example from `background_process.c`):
> 
> ```
> #pragma GCC diagnostic push
> #pragma GCC diagnostic ignored "-Wstringop-truncation"
> 			strncat(talkarray[4], lan_message + 2, 60);
> #pragma GCC diagnostic pop
> ```

Good find. Thanks for the pointer.
> 
> How an older GCC would handle this is questionable.
> 
And the other point is that we do not only have GCC here . Clang is used more and more and I am not quite sure if we only have linux systems trying to compile TLF out there.

> I certainly do not wish to turn off such warnings globally as doing so shortcuts any new code where a mistake is made and the warning is needed. OTOH, quelling the warning through this mechanism is fine so long as we've carefully vetted the code in question and the buffers like `talkarray[4]` are never shortened to less than the total number of concatenated characters.
> 
Maybe see how I quelled the warning in commit https://github.com/Tlf/tlf/pull/222/commits/cece425c5266089bd866c73ffb328e25431307b9

N0NB:
Yes, those look better and more simple, Tom.

Another thought I had was to use GString objects, but it would be a lot of work to convert the number of C strings in Tlf to them, I think.

You may close this request.  It has been a fun discussion.

dl1jbe:
I picked up your solution for scroll_log.c. It is correct and I had no code for it yet. Maybe we convert it later to some g_str... functions. See it in my PR.

Thanks for the help and the diskussion

PR #221, opened 2021-01-17
Use int for myrig_model

zcsahok:
To avoid warning in `parse_logcfg`. Also extended the range to match Hamlib4 rig numbering. (and dropped the default value as it's invalid anyway after the numbering change)


dl1jbe:
Yes, Looks like the easiest way out.

zcsahok:
No warnings with gcc10. We are good until int can be cast to rig_model_t.

PR #222, opened 2021-01-18
Fix warnings

dl1jbe:
First part of diverse changes to fix GCC 10 stringop warnings.

### DIFF
--- a/src/clusterinfo.c
+++ b/src/clusterinfo.c
@@ -204,18 +204,18 @@ void loadbandmap(void) {
     for (j = 0; j < nr_of_spots; j++) {
 
 	g_strlcpy(thisline, spot_ptr[j], sizeof(thisline));
-	if (strncmp(thisline, "DX de ", 6) == 0) {
+	if (strncmp(thisline, "DX de ", 6) == 0 && strlen(thisline) >= 74) {
 
 	    g_strlcpy(spotcall, thisline + 26, 6);
 
 	    /* read and convert hours and minutes to spotminutes */
-	    spottime[0] = *(thisline + 70);
-	    spottime[1] = *(thisline + 71);
+	    spottime[0] = thisline[70];
+	    spottime[1] = thisline[71];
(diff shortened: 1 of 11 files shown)


src/clusterinfo.c
```diff
@@ -208,12 +208,17 @@ int loadbandmap(void) {
...
 	    g_strlcpy(spotcall, thisline + 26, 6);
 
-	    strncpy(spottime, thisline + 70, 4);	// how old?
-	    spottime[4] = spottime[3];
-	    spottime[3] = spottime[2];
-	    spottime[2] = ':';
-	    spottime[5] = '\0';
-	    spotminutes = 60 * atoi(spottime) + atoi(spottime + 3);
+	    /* read and convert hours and minutes to spotminutes */
+	    spottime[0] = *(thisline + 70);
```
zcsahok, line 212: maybe `spottime[x] = thisline[y]` would be more readable

ideally one would need to check if `thisline` has enough content, something like strlen >= 74, on line 207 in the if()

src/tlf.h
```diff
@@ -149,7 +149,7 @@ typedef struct {
     char call[20];
     int countrynr;
     char continent[3];
-    char cqzone[3];
+    unsigned short cqzone;
```
zcsahok, line 152: I'd just use a plain `int`, even if wasting some bytes.

zcsahok (commented):
yesterday forgot to update test/data.c, will post a PR shortly

src/clusterinfo.c
dl1jbe: Right. Good. Idea. Wrt the length I will look into it.

src/tlf.h
dl1jbe: yes, we also had countrynr as int

N0NB:
As for scroll_log.c, you may as well match it to the fixes in the other files using `g_strndup()`.

dl1jbe:
No more comments -> will apply now

PR #228, opened 2021-01-24
Quell cppcheck warnings

N0NB:
Cleaning up some warnings generated by the cppcheck GUI tool.

### DIFF
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -161,16 +161,13 @@ void calledit(void) {
 		if (strlen(hiscall) + 1 == 12)
 		    break;	// leave insert mode
 
-		if (((i >= 65) && (i <= 90)) || ((i >= 47) && (i <= 57))) {
-		    call1[b] = i;
-		    call1[b + 1] = '\0';
-		    if ((strlen(call1) + strlen(call2)) < 12) {
-			strcat(call1, call2);
-//                      if (strlen(call1) + strlen(hiscall) >= 12) break;
-			if (strlen(call1) >= 12)
-			    break;
-			strcpy(hiscall, call1);
-		    }
+		call1[b] = i;
+		call1[b + 1] = '\0';
+		if ((strlen(call1) + strlen(call2)) < 12) {
+		    strcat(call1, call2);
+		    if (strlen(call1) >= 12)
+			break;
+		    strcpy(hiscall, call1);
 		}
 
 		if ((b < strlen(hiscall) - 1) && (b <= 12))
--- a/src/makelogline.c
+++ b/src/makelogline.c
@@ -144,7 +144,7 @@ void prepare_fixed_part(void) {
 	    ((strcmp(whichcontest, "qso") == 0) ||
 	     (strcmp(whichcontest, "dxped") == 0))) {
 	char khz[5];
-	sprintf(khz, " %3d", ((unsigned int)(freq / 1000.0)) % 1000);	// show freq.
+	sprintf(khz, " %3u", ((unsigned int)(freq / 1000.0)) % 1000);	// show freq.
 	strcat(logline4, khz);
 
     } else {
@@ -445,4 +445,3 @@ void fillto(int n) {
     if (len < n)
 	strncat(logline4, fillspaces, n - len);
 }
-
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -415,7 +415,7 @@ static int cfg_bandmap(const cfg_arg_t arg) {
 }
 
 static int cfg_cwspeed(const cfg_arg_t arg) {
-    int value = 18;
+    int value = 0;
     int rc = cfg_integer((cfg_arg_t) {.int_p = &value, .min = 6, .max = 60});
     if (rc != PARSE_OK) {
 	return rc;
@@ -425,7 +425,7 @@ static int cfg_cwspeed(const cfg_arg_t arg) {
 }
 
 static int cfg_cwtone(const cfg_arg_t arg) {
-    int value = 700;
+    int value = 0;
     int rc = cfg_integer((cfg_arg_t) {.int_p = &value, .min = 0, .max = 999});
     if (rc != PARSE_OK) {
 	return rc;
@@ -445,7 +445,7 @@ static int cfg_sunspots(const cfg_arg_t arg) {
 }
 
 static int cfg_sfi(const cfg_arg_t arg) {
-    int value = 67;
+    int value = 0;
     int rc = cfg_integer((cfg_arg_t) {.int_p = &value, .min = 0, .max = 1000});
     if (rc != PARSE_OK) {
 	return rc;
@@ -956,7 +956,7 @@ static int cfg_minitest(const cfg_arg_t arg) {
 	return PARSE_OK;
     }
 
-    int value = 60;
+    int value = 0;
     int rc = cfg_integer((cfg_arg_t) {.int_p = &value, .min = 60, .max = 1800});
     if (rc != PARSE_OK) {
 	return rc;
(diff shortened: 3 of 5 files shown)


src/makelogline.c
```diff
@@ -144,7 +144,7 @@ void prepare_fixed_part(void) {
 	    ((strcmp(whichcontest, "qso") == 0) ||
 	     (strcmp(whichcontest, "dxped") == 0))) {
 	char khz[5];
-	sprintf(khz, " %3d", ((unsigned int)(freq / 1000.0)) % 1000);	// show freq.
+	sprintf(khz, " %3u", ((unsigned int)(freq / 1000.0)) % 1000);	// show freq.
```
zcsahok, line 147: I'd prefer "%3d" with casting to int.
Unsigned int makes sense only if we'd like to log a 2+ GHz QSO on a 32 bit system. Quite unlikely.

A quick test for a potential negative double vs. unsigned int issue:
```
$ cat a.c
#include <stdio.h>

int main() {
    printf("%3d\n", (int)(12345.0/1000));
    printf("%3d\n", (int)(-12345.0/1000));
    printf("%3u\n", (int)(-12345.0/1000));
}

$ ./a.out 
 12
-12
4294967284

```

src/parse_logcfg.c
```diff
@@ -415,7 +415,7 @@ static int cfg_bandmap(const cfg_arg_t arg) {
 }
 
 static int cfg_cwspeed(const cfg_arg_t arg) {
-    int value;
+    int value = 18;
```
zcsahok, line 418: This assignment makes the false suggestion that the default speed is set here to 18 WPM.
This value is not used, as either `cfg_integer()` can parse the input and overwrites it or parsing fails and then error is returned.

The actual default WPM is set in `cw_utils.c`
```
int speed = 10;
```
No, it's not 10 WPM. It's then mapped using `char speedstr[]` and the final result is 30.

The general issue is that for historical reasons some numeric values are not stored in plain ints.
So the correct fix for such warnings is to refactor the code. I've it on my list.

src/writecabrillo.c
```diff
@@ -695,7 +695,7 @@ void write_adif_header(FILE *fp) {
     fprintf(fp, "Contest Name: %s\n", whichcontest);
     fputs("<adif_ver:4>3.10\n", fp);
     fputs("<programid:3>TLF\n", fp);
-    fprintf(fp, "<programversion:%ld>%s\n", strlen(VERSION), VERSION);
+    fprintf(fp, "<programversion:%ld>%s\n", (long int)strlen(VERSION), VERSION);
```
zcsahok, line 698: No point using a long int for abt 10. A simple `programversion:%d` would suffice.

src/calledit.c
```diff
@@ -161,16 +161,13 @@ void calledit(void) {
 		if (strlen(hiscall) + 1 == 12)
 		    break;	// leave insert mode
 
-		if (((i >= 65) && (i <= 90)) || ((i >= 47) && (i <= 57))) {
```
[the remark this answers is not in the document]
dl1jbe: No. Please have a look at line 157. That test is already done there.

src/makelogline.c
dl1jbe: Normally it would not make sense to have data type int for things which cannot be negative. Not having to test the lower border simplifies some tests.

But as we convert here from float anyway it would be good to keep the sign information.

src/calledit.c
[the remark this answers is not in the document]
zcsahok: OK, I see. (line 152)
BTW we are just lucky that `/` and `0-9` can be covered by a single ASCII range... Took me some time to figure it out.
If it were
```
 if ((i >= 'A' && i <= 'Z') || (i >= '0' && i <= '9') || i == '/') {
```
it would have been clearer to me.
(comment has a typo: _0_-9)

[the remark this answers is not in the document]
dl1jbe: Really old code here. If we touch the code to drop the extra check we should also clean up that test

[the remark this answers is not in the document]
N0NB: I stared at that for a good while and didn't catch the 1-9 in the comment and I have a 0 in my call!

While this is an example of a simple test using ASCII, I do wonder if we should be using standard library functions or glib functions like `g_ascii_isalnum ()` or `g_ascii_toupper ()`?  In many cases it seems like overkill when a simple `i = i - 32;` does the same thing.  Of course, should the project ever adopt using Unicode beyond the ASCII character set then all of this would need to be revisited.  In new code the use of library functions should be encouraged, IMO.

src/makelogline.c
N0NB: Funny you mention 2 GHz.  I having been considering a wishlist issue to bring Tlf out of the HF only world and be useful for VHF/UHF bands.  With the release of the Raspberry Pi 400 I have seen some hams write about using it for things like Field Day operations which in the USA permit operations on all bands though any bands above 23 cm are probably seldom used.

I see Tom prefers to simply cast to `int` and keep the `d` format specifier so that is what I'll go with.

src/parse_logcfg.c
N0NB: I think the issue is that cppcheck can't evaluate the expression at line 419 where the address of `value` is passed to be set and assumes that it is not initialized.

I was just picking values for fun but I'd also like to not have this come up every time I want to do some code checking.  Perhaps simply initializing the values to 0 would be acceptable?

src/writecabrillo.c
N0NB: That was my thought.  But, someone (Tom?) tossed that `l` in there so I worked around that.  :)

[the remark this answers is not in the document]
N0NB: Yes, I'll do that.  Made me reread the man page again!

src/makelogline.c
zcsahok: RPi 400 seems to be 64 bit, so one can tune until 9223372036 GHz, somewhere in the X-ray band. :-)

src/parse_logcfg.c
zcsahok: I guess cppcheck is afraid that the value is never assigned to.
Setting to 0 with an appropriate comment would be fine for the moment.

src/makelogline.c
N0NB: I made the change to:

```
147         sprintf(khz, " %3d", ((int)(freq / 1000.0)) % 1000);    // show freq.
```

and am greeted with this from gcc 10.2.1:

```
../../tlf/src/makelogline.c: In function ‘prepare_fixed_part’:
../../tlf/src/makelogline.c:147:20: warning: ‘sprintf’ may write a terminating nul past the end of the destination [-Wformat-overflow=]
  147 |  sprintf(khz, " %3d", ((int)(freq / 1000.0)) % 1000); // show freq.
      |                    ^
../../tlf/src/makelogline.c:147:2: note: ‘sprintf’ output between 5 and 6 bytes into a destination of size 5
  147 |  sprintf(khz, " %3d", ((int)(freq / 1000.0)) % 1000); // show freq.
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

While clang 11.0.1 says nothing.  Hmmm.

Now I prefer my original patch of simply changing the conversion specifier.

dl1jbe:
That only leaves the 'divide by 0' warning in line 965 of parse_logcfg.c. @zcsahok I would set the value to 1 in line 959 and add an comment. What do you think?

zcsahok:
OK. This is when compiler tries to outsmart the programmer...

N0NB:
Hmmmm,  My runs have not shown a divide by zero error and I missed it as I simply did not read the code.

Ooops....

dl1jbe:
That one comes from clang :-). I am curious what the coming GCC-11 will bring....

zcsahok:
If even current version already can divide by 0...
(there is of course no div-by-zero here)

N0NB:
Thanks, Tom.

PR #229, opened 2021-01-25
Implementing configuration statement HDRLINE_FKEYS= to allow the loca…

ha5se:
…lized

F-key assignments shown on the header line, instead of the constant F-key info.

### DIFF
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -1027,18 +1027,6 @@ static int cfg_cabrillo_field(const cfg_arg_t arg) {
     return rc;
 }
 
-// 2021-01-23 HA5SE implementing HDRLINE_FKEYS=
-static int cfg_headerline(const cfg_arg_t arg) {
-    char *str = g_ascii_strup(parameter, -1);
-    if (strlen(str) >= 58) {
-	g_free(str);
-	return PARSE_STRING_TOO_LONG;
-    }
-    strcpy(headerline+3, str);
-    g_free(str);
-    return PARSE_OK;
-}
-
 static config_t logcfg_configs[] = {
     {"CONTEST_MODE",        CFG_BOOL_TRUE(iscontest)},
     {"MIXED",               CFG_BOOL_TRUE(mixedmode)},
@@ -1155,6 +1143,7 @@ static config_t logcfg_configs[] = {
     {"SC_DEVICE",       CFG_STRING_STATIC(sc_device, 40)},
     {"INITIAL_EXCHANGE",       CFG_STRING_STATIC(exchange_list, 40)},
     {"DIGIMODEM",       CFG_STRING_STATIC(rttyoutput, 120)},
+    {"HDRLINE_FKEYS",   CFG_STRING_STATIC(headerline+3, 57)},
 
     {"CABRILLO",    CFG_STRING(cabrillo)},
     {"CALLMASTER",  CFG_STRING(callmaster_filename)},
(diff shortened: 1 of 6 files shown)


src/parse_logcfg.c
```diff
@@ -1027,6 +1027,18 @@ static int cfg_cabrillo_field(const cfg_arg_t arg) {
     return rc;
 }
 
+// 2021-01-23 HA5SE implementing HDRLINE_FKEYS=
+static int cfg_headerline(const cfg_arg_t arg) {
```
zcsahok, line 1031: Szia and thanks for the PR!

As it just came up in #228 simple parsing methods (int/bool/string) shall be preferred over using a specific one whenever possible. In this case I'd factor out the first 3 constant spaces from `headline[]` and add them explicitly in `show_header_line()`:
```
    mvaddstr(0, 21, spaces(3));
    mvaddstr(0, 24, headerline);
```
Then `headline` is just a "normal" static string.
Well, almost. See this screenshot:
image

ha5se: Hi Zoli,
thank you very much for your comment, very good hit. Hope it suits standards better now.
Admitting I am unfamiliar with both C and git, was already happy that after many failures, it compiled and worked  as expected ;-) . But fully agree that this is in fact easier to read and maintain. Should have been deeper consulting existing code before starting to modify ;-) .

dl1jbe: Hi and welcome. 

Some comments from my side:

- I would not hard code the 3 spaces into `show_header_line`. In the original string they make sure the text is centered. So maybe better to pick up  `headerline` and print it centered starting from position 21.  That could also fill the space right of the text with the right color.

- I would suggest a simpler keyword - how about just HEADERLINE or HDRLINE?

@ha5se Wrt being unfamiliar - we all started small. Keep your ideas going, we are here to help.

zcsahok: Maybe FKEY-HEADER would be a good name.

ha5se: Hi Thomas (Grüss Gott),

in contrast to Zoli's comment _(simple parsing methods shall be preferred...)_ , that I  agreed right at once, sorry but I feel forced to disagree all of YOUR points.

1) hdrline has three other fields left of F-Keys... You cannot simply call "headerline" only the right part of the header  line...

However, I do not like the name, either. Could contain any info, not only F-Keys...

HDRLINE_FKEYS= 1=CQ  2=DE  3=RST 4=73  5=HIS  6=MY  7=B4   8=AGN  9=?
HDRLINE_FKEYS= Darling, Love You, today contesting, pse ask my creditcard tomorrow

How about **WIFE_MSG**? -- well, about 50% we man do not have wife.... Not mentioning the few YL calls...
How about **SMALLINFO** ?

2) It is a matter of taste - however, for me, less then three spaces was visually unacceptable

3) centering is ... again a matter of taste. Obviously much beyond my scopes... It is against keeping simple, and IMHO also it is not needed. Personally, I find the room too limited, and  would be very happy with 4+ more spaces for my own... And again, anybody can shift/center her/his own wishes in the conf.

4) After all, I felt kind of fed up with this kind of developing (I am 63, my first git). Still I thank for everybody. Still I want to go on with my next idea... After a long discussion with Zoli, we agreed that he will take over this one and finish alongside centering, and then he is supposed to kill  this my attempt...

dl1jbe: >     1. hdrline has three other fields left of F-Keys... You cannot simply call "headerline" only the right part of the header  line...
> 
> However, I do not like the name, either. Could contain any info, not only F-Keys...
> 
So let us see if we can find something better.


>     1. It is a matter of taste - however, for me, less then three spaces was visually unacceptable
 
>     2. centering is ... again a matter of taste. Obviously much beyond my scopes... It is against keeping simple, and IMHO also it is not needed. 

The idea behind my suggestion was me thinking what would it be like if we are able to resize tlf not only in vertical direction but also in horizontal one. How should  these partial field be used - left stuck to left part of the screen, floating to the center of the available space or something different?
Personally, I find the room too limited, and  would be very happy with 4+ more spaces for my own... And again, anybody can shift/center her/his own wishes in the conf.
 
>     3. After all, I felt kind of fed up with this kind of developing (I am 63, my first git). 

Welcome in the club, me reaching these age in some days.

> After a long discussion with Zoli, we agreed that he will take over this one and finish alongside centering, and then he is supposed to kill  this my attempt...

Anyway thanks for your contribution.

zcsahok:
This one can be closed.

PR #230, opened 2021-01-25
Update docs for Cabrillo

zcsahok:
And general fixes mostly in man page.
Pse review.

- Tried to remove references like "since version 0.9.2" or "before 1.2.x", just focusing on what we have currently.
- Also removed info on deprecated RPC rigd, PI4TUE cluster, and the defunct old manual URL.
- Moved "unproductive" options (-d/-v/-V) to the end of the list.
- Fixed file locations.
- Ctrl-D doesn't exit. BTW we have to review the possible exit methods (so many of them: this is the program with most options to exit...) as I have the impression that not all work the same way. (some say 73 at the end, some don't)
- Added MARKERDOTS/CALLS to logcfg.dat
- capitalized Cabrillo


dl1jbe:
Good work. What should be added to the man page is that keywords in rule file takes precedence over similar keywords in logcfg.dat. Would be nice if could add it while documenting the grab feature (PR #231).

Wrt the exit options: It is quite true that there are too much (and there where even more before).  While trying to iron out these doubles we should check all the other keys - there are some more doubles around. It could free some combinations for other use.

zcsahok:
> What should be added to the man page is that keywords in rule file takes precedence over similar keywords in logcfg.dat.

It was already there at least in 1.3.2:

> RULES
>       The  contest rules can be put into separate files.  Tlf will first look
> ...
>       NB: The rules file overrides settings from the logcfg.dat file!

dl1jbe:
> > What should be added to the man page is that keywords in rule file takes precedence over similar keywords in logcfg.dat.
> 
> It was already there at least in 1.3.2:

Sorry. I did not look careful enough. It is still here.

Issue #31, opened 2015-01-17
Bandmap improvements


dh5fs:
One of the big weak points for realtime logging is still the bandmap display. In detail:
- the shown number of stations is too small > bigger terminal window possible, or even a seperate one?
- no connection of the bandmap to the actual frequency > if the number of calls stays limited there should be some sorting around the actual working frequency
- no filtering for Mults possible, not even for DXCC Mults
- no free movement in the bandmap possible > something like ALT-ARROW UP, DOWN to move up/down in frequency would be nice

73 Fred

dl1jbe:
Actual state - a short statement:

> no connection of the bandmap to the actual frequency > if the number of calls stays limited there should be 
> some sorting around the actual working frequency

That is implemented now. See commit  
https://github.com/Tlf/tlf/commit/3be8dd518b84d1e00209e6ce840ad274a3342720

> free movement in bandmap

needs quite some work on key input handling as ncurses limits the range of usable key combinations. It is on the ToDo list, but will take time.

> display of more spots, bigger window

Some ideas to allow more than 25 lines for the program window are in the work at the moment. Unluckily most display postions are hard coded so it is quite some work to make it more flexible.

dl1jbe:
bandmap display is now centered around center frequency of actual band and mode segment if you have no running rig connected.

dl1jbe:
HA2OS provided a patch to save and reload bandmap data. That will be helpful if tlf crashes or you have to temporarily leave tlf.

zcsahok:
Fred, are there still something that need to be fixed/added or can we close this issue?

dh5fs:
Still to move with CTRL-G only into one direction and not to move in the opposite frequency direction if needed would be a valuable improvement.

zcsahok:
Do you mean not reversing search direction at the end of the list, but rather jump to the beginning?

dh5fs:
I mean to change direction whenever needed, not just at the end of the list.

zcsahok:
I see, makes sense. Just have to come up with a key for it. Maybe `Ctrl-Shift-G` could work.
Eventually mouse clicking on the list could be supported too. I saw plans for this.

zcsahok:
How about a grab direction indicator as shown in the screenshot?
Arrows (well, pretty crude) pointing downwards mean grabbing up in frequency matching the visual representation of the band.
[...]

PR #231, opened 2021-01-26
Grab direction display and toggle

zcsahok:
As per #31.

Also
- In `bandmap_next()` merged the up and down loops into a single one.
- `spaces()` instead of strings of spaces.
- Fixed help for Ctrl-T.

Will update man page when #230  is closed.


zcsahok:
Updated man page.

The last commit fixes the issue that some test programs (run_*) requiring data files from share only work when executed from the test dir. Sometimes I use them for prototyping and it's annoying to always change dirs.

dl1jbe:
Wrt the merge of master: 

It keeps your branch mergeable. So it should normally be helpful. But I remember one or two occasions were I had to wrangle with the merge into master afterwards. Maybe I made a mistake there.

I am not so sure if it also works for the automatic `squash` provided by github. We used that often in last time for feature branches where the internal steps were not worth to be remembered. If that makes problem we can try to rebase your branch on top of master before squashing.

Issue #234, opened 2021-01-29
Long exchanges truncated


zcsahok:
Exchange input field allows to enter up to 25 characters.

![image](https://user-images.githubusercontent.com/15141948/106332274-be227100-6286-11eb-8a6a-0afd938aa1cf.png)

After logging the QSO the exchange is truncated to 14 characters.

![image](https://user-images.githubusercontent.com/15141948/106332393-0477d000-6287-11eb-969f-02fe1f0d6673.png)

Same in the log file:
![image](https://user-images.githubusercontent.com/15141948/106332578-5f112c00-6287-11eb-9f03-76bdaed755a0.png)

dl1jbe:
I assume you were in contest mode while testing.

The behavior is due to limitations in the TR log format we inherited from the original program. That format reserves a varying amount of line space for recording of multipliers and points depending on the contest/dxped/qso mode you use.

The exact amount which gets recorded is determined in makelogline.c in `prepare_specific_part`.

I agree that we need a better solution. Best would be to drop the old log file format and switch to a more flexible one - I would suggest  a database related approach here - but other schemes may be possible. Downside is that it breaks compatibility with older log version completely. 

As it looks we have quite some serious changes to make 

- log format
- handling of configuration storage and usage
- rework of point and multiplier scoring logic
- resizeable layout in horizontal and vertical direction

which may all break compatibility in a drastic manner.

Therefore I would suggest starting a discussion about a new TLF-2.xx version. Once the goals are clear we could prepare one last TLF-1.x.x version with all the features and fixes done in since TLF-1.4.1 and than start over to work on the new version.

zcsahok:
For me limiting the exchange input length to 14 chars would be perfectly fine. At least it must be consistent.

dl1jbe:
Good then let us keep the issue open to remind us to fix it.

N0NB:
Let me offer this, if a major rewrite of the log file format is to be done, do we want to retain the simple flat file format or go to something like an SQLite database?

A simple flat file format is easy to work with in any text editor but it is fragile in the sense that great care needs to be take to assure its integrity throughout the program.  Using an SQLite database may protect against this somewhat though it is not as strict (as I recall) enforcing data types as Mariadb or Postgresql, for example, but does not require a DB administrator either as everything is kept in one file.  SQLite has bindings for many languages so external log analysis tools could be written fairly easily.

[...]

PR #237, opened 2021-02-03
:INFo shows cty.dat version

zcsahok:
Recently came across this: https://www.country-files.com/contest/#Version

> In the list of prefixes/callsigns, you’ll find a “callsign” that looks like:VERyyyymmdd where “yyyymmdd” is the date of the release.

This is now stored and can be displayed with :INFo.

image

The other method works too, that needs no change. (enter VERSION and check the country)

image

### DIFF
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -35,6 +35,7 @@
 #include "cqww_simulator.h"
 #include "changepars.h"
 #include "clear_display.h"
+#include "dxcc.h"
 #include "editlog.h"
 #include "err_utils.h"
 #include "fldigixmlrpc.h"
@@ -330,7 +331,7 @@ int changepars(void) {
 	    read_logcfg();
 	    read_rules();	/* also reread rules file */
 	    TLF_LOG_INFO("Logcfg.dat loaded, parameters written.");
-            center_fkey_header();
+	    center_fkey_header();
 	    clear_display();
 	    break;
 	}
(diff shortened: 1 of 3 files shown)


airween:
> The other method works too, that needs no change. (enter VERSION and check the country)

This doesn't work for me. I typed `VE` then Tlf shows `VE Canada` in the bottom line, and when I typed the full word `VERSION` then it shows `3D2/c  Conway Reef`.

In the cty.dat I see `=VER20200405`.

airween:
Btw the PR is LGTM (Looks Good To Me).

zcsahok:
You have a very old version, it's even not present on the main history page, had to browse here
http://www.country-files.com/category/contest/page/2/
 

image

airween:
> You have a very old version

yes I know, but IMHO it doesn't matter - a version is a version :).

I changed it, now the `VERSION` now shows `ZS South Africa`.

But the `VER` shows `VER20200402` in the partial call window - which is weird, because I [downloaded](https://github.com/Tlf/tlf/issues/234#issuecomment-772819278) the cty.dat to CWD.

```
$ grep -n =VER cty.dat 
1170:    XL2[4],XM2[4],XN1,XN2(2),XO0(2)[4],XO1(1)[2],XO2,=VER20210127,
1328:    H5,S4,S8,V9,ZR,ZS,ZT,ZU,=VERSION;
```

zcsahok:
Partials are taken from the callmaster, so that's cm's version.

dl1jbe:
Ervin, just in cas you missed it - there are two ways to check the version:

1. look for VERyyyymmdd
2. enter VERSION as callsign and look for the country it reports

see https://www.country-files.com/contest/#Version

dl1jbe (approved):
Good idea. Looks good for me.

airween:
Thanks guys for tips - I'll check the versions in future :).

Thomas approved this patch too so I'm going to merge it - thanks for all.

PR #240, opened 2021-02-06
Speedup find_best_match function

dl1jbe:
Add a HashTable to prefix database to speedup lookup of station prefixes.

Using cluster for bandmap and/or xplanet display consumed a major part of time in find_best_function().
With the new lookup mechanism the time for these operation nearly vanished.

### DIFF
--- a/src/dxcc.c
+++ b/src/dxcc.c
@@ -81,6 +81,15 @@ prefix_data *prefix_by_index(unsigned int index) {
     return (prefix_data *)g_ptr_array_index(prefix, index);
 }
 
+/* lookup key in table of hashed prefixes
+ * \return - true, if found in HashTable
+ * \param key - part of call to look up
+ * \param value - the corresponding prefix index
+ */
+gboolean lookup_hashed_prefix(const char *key, void *value) {
+    return g_hash_table_lookup_extended(hashed_prefix, key, NULL, value);
+}
+
 /* add a new prefix description */
 void prefix_add(char *pfxstr) {
 
--- a/src/dxcc.c
+++ b/src/dxcc.c
@@ -30,6 +30,7 @@
 
 GPtrArray *dxcc;
 GPtrArray *prefix;
+GHashTable *hashed_prefix;
 bool have_exact_matches;
 char cty_dat_version[12];   // VERyyyymmdd
 
@@ -56,6 +57,11 @@ void prefix_free(gpointer data) {
 
 
 void prefix_init(void) {
+    if (hashed_prefix) {
+	g_hash_table_destroy(hashed_prefix);
+    }
+    hashed_prefix = g_hash_table_new(g_str_hash, g_str_equal);
+
     if (prefix) {
 	g_ptr_array_free(prefix, TRUE);
     }
@@ -141,6 +147,9 @@ void prefix_add(char *pfxstr) {
     new_prefix -> dxcc_index = last_index;
 
     g_ptr_array_add(prefix, new_prefix);
+    g_hash_table_insert(hashed_prefix,
+	    new_prefix->pfx,
+	    GINT_TO_POINTER(prefix_count() - 1));
 }
 
 
--- a/src/dxcc.h
+++ b/src/dxcc.h
@@ -22,6 +22,7 @@
 #define DXCC_H
 
 #include <stdbool.h>
+#include <glib.h>
 
 typedef struct {
     char *pfx;
@@ -55,6 +56,8 @@ void prefix_init(void);
 unsigned int prefix_count(void);
 
 prefix_data *prefix_by_index(unsigned int index);
+
+gboolean lookup_hashed_prefix(const char *key, void *value);
 
 void prefix_add(char *pfxstr);
 
--- a/src/getctydata.c
+++ b/src/getctydata.c
@@ -59,34 +59,34 @@ int find_full_match(const char *call) {
 
 /* search for the best mach of 'call' in pfx table */
 int find_best_match(const char *call) {
-    int bestlen = 0;
-    int i, w;
-    prefix_data *pfx;
-    int pfxmax = prefix_count();
+    void *value;
+    int w = -1;
 
-    w = -1;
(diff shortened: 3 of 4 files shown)


N0NB (approved):

N0NB:
LGTM, Tom.

Admittedly, this is a part of the code base I'm not familiar with, but your comments on speed increases sounds worthwhile.

Not too add to anyon'e workload, but I am revisiting the use of Doxygen in Hamlib and am ashamed to admit that I've ignored it for far too long.  I know some of the functions are documented with Doxygen type comments and I wonder if it would help others approach development if the various functions were so documented and a manual of the internals of sorts could be generated.  Just a thought as I've been digging into Doxygen the past couple of days.

dl1jbe:
Thanks for review.

Wrt Doxygen:

> Not too add to anyone's workload, but I am revisiting the use of Doxygen in Hamlib and am ashamed to admit that I've ignored it for far too long. I know some of the functions in the Tlf code base are documented with Doxygen type comments and I wonder if it would help others approach development if the various functions were so documented and a manual of the internals of sorts could be generated. Just a thought as I've been digging into Doxygen the past couple of days.

You are completely right. Note to myself: Would be good to add more doxygen comments - at least for the public declarations in the *.h file. Will try ....

dl1jbe:
> Not too add to anyone's workload, but I am revisiting the use of Doxygen in Hamlib and am ashamed to admit that I've ignored it for far too long. I know some of the functions in the Tlf code base are documented with Doxygen type comments and I wonder if it would help others approach development if the various functions were so documented and a manual of the internals of sorts could be generated. Just a thought as I've been digging into Doxygen the past couple of days.

dl1jbe:
@zcsahok Looks like I was a little bit to fast. Just saw you have some more changes in your repo. Please place a new PR when ready.

zcsahok:
_(got merged while writing this, but nevetheless)_
**Summary**: I suggest adding hashing on the first 2-chars of the prefix to get a quick hit for unique prefixes.
Code is here: https://github.com/zcsahok/tlf/commit/52836352762cba1ec1d3d64fee8889625e944b73

Here's my working version for the speed-up, just never had time to push it: https://github.com/zcsahok/tlf/commit/9824cc8607610d9ec61d7bf7d68e2d3ea937febc
It uses an more elementary approach by keeping the original linear search but partitioning the 5k long prefix list by the first 2-characters of the prefix.
I have also added a speed test  to test_getctydata.c to quantify the resulting gain. The test draws 20M (for the original code only 200k) random calls from the 35k callmaster entries. Just run
```
$ test/run_getctydata | head
[==========] Running 15 test(s).
[ RUN      ] test_xspeed
got 35419 calls from ...
took 7094277 us, rate is 0.354714 us/call
```

I've applied the same speed check to this PR and the 2-char pfx addition. Here are the results from my Core2 2.6 GHz test system. On a slowish RPi-like machine everything is factor 10 slower, so the data are consistent.
- original code: 35 us/call
- partitioning + lin. search: 0.28 us/call
- GHash + shortening call (this PR): 0.43 us/call
- adding 2-char hash to above: 0.35 us/call

The thing is that there are a lot of "easy" unique 2-char prefixes and the call shortening doesn't make use this fact.
Out of the 36*36 possible combinations (36 is the number of chars in '0-9A-Z') 30% are not existing.
From the existing prefixes almost 60% are unique (have no exception), like S5 or DL.
So first checking if the prefix is unique one can avoid making a bunch of hash lookups.

Good question why GHash + shortening is slower than the partitioned linear search.
I guess GHash being a universal tool does things internally that are not needed for our case.
Additionally the shortening search in the typical 2+1+3 (XY1ABC) call case makes 4 unsuccessful lookup, plus the first lookup, so in total has 5 misses before a hit.
The partitioned linear segments a quite short and the comparisons are quite easy (comparing short strings) that compensate the ineffectiveness of the search.

zcsahok:
@N0NB it would make sense to add such call lookup function to Hamlib as this is not a Tlf-specific functionality

N0NB:
> @N0NB it would make sense to add such call lookup function to Hamlib as this is not a Tlf-specific functionality

???

I'm not sure what you're referring to.

zcsahok:
Sorry, I mean having a function in Hamlib to look up country data for a call. Just like QRB calculation this is universal, of course provided a cty.dat is available.

N0NB:
I think that is out of scope for Hamlib.  However, I do think there are things that Tlf does that could be broken out into a separate library that could be reusable by other software.  Doing so would make testing that code independent of trying to simulate contest conditions in Tlf.

dl1jbe:
@zcsahok Thanks for the numbers. I only watched the relative workload with `perf top --pid <tlf-pid>`. With RBN spots from N8NOE during yesterdays contests the percentage of time spent in `find_best_match()` dropped from around 60..80% to below 0.1%. 

Will look into your branches shortly.

airween:
> I think that is out of scope for Hamlib.

Zoli's example is very good, why it is relevant to Hamlib :) (calculating QRB).

> However, I do think there are things that Tlf does that could be broken out into a separate library that could be reusable by other software. Doing so would make testing that code independent of trying to simulate contest conditions in Tlf.

Yes it makes sense.

N0NB:
The QRB functions were added as they were essentially "free" since the rot_ctl interface required lat/lon calculations.  This, processing CTY.DAT, is a whole new set of issues and is more limited in scope than the Hamlib API as I see it.  After processing CTY.DAT comes processing master.scp and so on.  I think such should be a separate library/project.

dl1jbe:
So, had a deeper look into it. Will pick up the two-letter hashing.
 
> Good question why GHash + shortening is slower than the partitioned linear search.
> I guess GHash being a universal tool does things internally that are not needed for our case.
> Additionally the shortening search in the typical 2+1+3 (XY1ABC) call case makes 4 unsuccessful lookups, so it has 4 misses before a hit.

Also the hash function has to operate on 2..6 characters instead of 2 and there may be some string compares if the hash is not unique.

PR #241, opened 2021-02-07
Change display of multis in bandmap for better readability

dl1jbe:
As multi display is only supported in CQWW contest please check it out there.


zcsahok:
ok for me, looks nice.

image

dl1jbe:
Very well. Pulling it in.

PR #242, opened 2021-02-07
Modify filter_spots()

dl1jbe:
* Set dupes for all spots in 'allspots' list
* Furthermore simplify the check logic

### DIFF
--- a/src/bandmap.c
+++ b/src/bandmap.c
@@ -657,11 +657,11 @@ freq_t bm_get_center(int band, int mode) {
 }
 
 /* small helpers for filter_spots() */
-bool band_matches(spot *data) {
+static bool band_matches(spot *data) {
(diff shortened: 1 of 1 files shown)


src/bandmap.c
```diff
@@ -656,6 +656,15 @@ freq_t bm_get_center(int band, int mode) {
     return centerfrequency;
 }
 
+/* small helpers for filter_spots() */
+bool band_matches(spot *data) {
```
zcsahok, line 660: unless used elsewhere these could be static to allow inlining by compiler

zcsahok (approved):
Thanks!

src/bandmap.c
dl1jbe: Yes, good idea.

PR #243, opened 2021-02-08
Add 2-char hash for further speedup

dl1jbe:
See PR#240 for discussion so far.

Moved `find_best/full_match()` to dxcc.c.

### DIFF
--- a/test/test_getctydata.c
+++ b/test/test_getctydata.c
@@ -1,3 +1,4 @@
+#include <inttypes.h>
 #include "test.h"
 
 #include "../src/tlf.h"
@@ -48,7 +49,7 @@ char *best_prefix(char *call) {
     return pfx->pfx;
 }
 
-char *full_prefix(char * call) {
+char *full_prefix(char *call) {
     prefix_data *pfx;
     int index;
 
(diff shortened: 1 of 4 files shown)


test/test_getctydata.c
```diff
@@ -60,6 +61,59 @@ char *full_prefix(char * call) {
     return pfx->pfx;
 }
 
+void test_xspeed(void **state) {
```
zcsahok, line 64: this could be removed (added for reference), or at least N reduced, as it takes precious seconds on each Travis run.
(not sure what is the current state at Travis, they were supposed to stop the free service end of 2020)

zcsahok (commented):
Thanks!

test/test_getctydata.c
dl1jbe: > this could be removed (added for reference), or at least N reduced, as it takes precious seconds on each Travis run.

I will look to keep it around but have a compile time switch (a simple #define) to make it a no-op if not needed.

> (not sure what is the current state at Travis, they were supposed to stop the free service end of 2020)

Just copied the following from travis-ci.com: 

"Testing your open source projects is always 100% free!"

PR #244, opened 2021-02-09
New config statement TUNE_SECONDS= to specify maximum length of tuning that was started with Alt-T. New code allows above the former 10 sec limit that is imposed by cwdaemon.

ha5se:


src/callinput.c
```diff
@@ -81,6 +81,7 @@
 #include "fldigixmlrpc.h"
 
 #define TUNE_UP 6	/* tune up for 6 s (no more than 10) */
+int tune_seconds = TUNE_UP;		// for back-ward compatibility, keep old default
```
dl1jbe, line 85: I would suggest moving the above line to main.c and the `#define' as well as the 'extern' declaration to globalvars.h

src/parse_logcfg.h
```diff
@@ -40,6 +40,7 @@ int parse_configfile(FILE *fp);
 int parse_logcfg(char *inputbuffer);
 
 extern char *error_details;
+extern int tune_seconds;
```
dl1jbe, line 44: Better declare variable needed 'during' parse_logcfg in the *.c file. 
But it should no longer be needed at all if you move it to globalsvars.h as suggested above.

dl1jbe (commented):
Seems to be useful.

src/callinput.c
ha5se: Thomas, thank you for the comments. However, while admitting being beginner and poor in C, still I cannot figure out your point. TUNE_UP (now tune_seconds) has always been defined in this module, and used here. No use in main.c . Why should it be moved there then?
Anyway,  being also in git an absolute rookie, now I am stuck here. So, I kindly ask you to put any final tidy-up into this to your liking. TIA

zcsahok: @dl1jbe I'll do the merge, it that's OK.

dl1jbe: @zcsahok: No problem. Please go ahead. 

@ha5se: 
> However, while admitting being beginner and poor in C, still I cannot figure out your point. TUNE_UP (now tune_seconds) has always been defined in this module, and used here. No use in main.c . Why should it be moved there then?

Let me explain.

First I was not talking about `TUNE_UP` but about the `tune_second` variable.

When defining global variables we have to choices:

* Variables which are only part of an module should be defined in the module itself and made known to other modules by declaring them `extern` in the modules header file 
* Alternatively if the variable is not clearly related to a functional module we have concentrated them in one place - at the moment in main.c (with test/data.c a shadow copy for testing purposes). The variable should then be declared in `globalvars.h`

My comment came from the first impression that `tune_seconds` is such an unrelated variable. A thorough rechecking of all references to the variable shows now that I am partly wrong. It can be seen as well as strongly related to the internal working of callinput() to be set from outside by parse_logcfg().

In light of that it may also stay where it is - but the extern declaration should go then to callinput.h and that one included in parse_logcfg.c. Let us see how Zoli will handle it.

Remark:
* There is quite a lot of `extern` declarations and variable placement which does not fit into the above. Most of it is inherited from old code and we are in process to sort it out.

> Anyway, being also in git an absolute rookie, now I am stuck here. 

Maybe just one hint how to handle the git process for reviews: Just add another commit on top of your branch (in that case TUNE_SECONDS) with the changes and push it up to your git repo. It will be automatically added to your pull request.


Anyway thanks for your PR. We will gladly help you to bring it in.

zcsahok:
I did the merge using command line using the suggestion at the bottom of the PR page. This resulted in 2 commits, not sure if this is optimal. Anyway.

Moved `tune_seconds` to main.c even if it belongs functionally to callinput as it would be the only variable exported from there. In main.c initialization happens in `init_variables` in order to allow re-initializing, in case we would need it for config reload. Also added a test.

ha5se:
Thanks for the explanation, and thanks for the merge. Considering the formerly forgotten re-init, with three modules now, it clearly belongs to globalvars in the current "architecture". However, I think it could be worth considering, that (at least to my taste), most of (or perhaps the complete[*] ) init_variables would logically much better fit into parse_logcfg.c where most of the variables could immediately be overwritten anyway... Why bother two modules when one could be enough?

[*] BTW, some existing constant values will surely similarly transfer into config variables in the future. (at least me having my next candidate, for sure ;-) ).

Issue #247, opened 2021-02-15
Plugin support


zcsahok:
_(just fyi, no action is needed)_

I'm working on a proof of concept implementation of using contest-specific plugins. As a test I have defined the multi rules for MWC (https://memorial-ok1wc.cz/index.php?page=rules2l) that takes place each Monday from 1630 UT. For multiplier they use the last character of the worked call. The rationale behind this is likely to allow higher mult counts even if only a handful of DXCC entities participate.
Such plugins would allow Tlf to support practically any contest and would also be a solution for #235.

Here is the screenshot after today's contest, I still had some unworked mults on 80 and even more on 40.

![image](https://user-images.githubusercontent.com/15141948/107980388-71e26b00-6fc0-11eb-9f5c-f07bf232a3da.png)

As plugin technology I use Python as it is quite widespread and has a lot of features. The plugin script is implicitly selected by the contest rule config: for RULES=xdx the file `rules/xdx.py` is loaded if available.

Currently the interface methods are:

- setup: called to reset the internal state
- add_qso: called when a QSO is logged (either interactively or by reading the log)
- nr_of_mults: number of mults per band or total, used for the display
- is_multi: to tell if a call would be a multi, the functionality of `bm_ismulti()`

There are still a number of rough edges and some stuff are not too tlf-ish (i.e. bypassing `multis[]`), so work is in progress.

The code for MWC:
```
import tlf

mults = set()

def setup():
    global mults
    mults = set()

def add_qso(qso):
    multi_key = get_multi_key(qso.call, qso.band)
    if multi_key in mults:
        return
    mults.add(multi_key)

def nr_of_mults(band):
    if band == tlf.BAND_ANY:
        return max(len(mults), 1)
   
    bk = f'{band};'
    n = 0
    for multi_key in mults:
        if multi_key.startswith(bk):
            n = n + 1
    return n

def is_multi(qso):
    multi_key = get_multi_key(qso.call, qso.band)
    return not multi_key in mults


#######################################
#
# internal functions
#

def get_multi_key(call, band):
    a = call.split('/')
    base = a[0]
    if len(a) >= 2 and len(a[1]) > len(a[0]):
        base = a[1]
    return f'{band};{base[-1]}'
```

airween:
Hi Zoli,

do you already work on it?

This was my idea too exactly :).

If you have any working part, could you share this with us? We should make a new branch here (eg: dev/plugins).

zcsahok:
[...]

PR #249, opened 2021-02-18
Clean-up exits (no Ctrl-C/Alt-Q)

zcsahok:
As discussed in #247

### DIFF
--- a/share/help.txt
+++ b/share/help.txt
@@ -8,18 +8,18 @@ Alt-T     Tune                          Alt-N   Note in log
 Alt-V     Band down (CW Speed?)         Alt-P   Turn PTT on/off
 Alt-H     Show help.txt                 Alt-W   CW Weight
 Alt-I     Show talk messages            Alt-J   See QRG's
-Alt-K     CW from Keyboard              Alt-X   Exit program (Alt-q)
-Alt-M     Multi display (Alt-c)         Alt-S   Score window (Alt-r)
-Alt-,     Show bandmap                  Alt-J   Show frequency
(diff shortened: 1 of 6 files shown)


share/help.txt
```diff
@@ -8,18 +8,18 @@ Alt-T     Tune                          Alt-N   Note in log
 Alt-V     Band down (CW Speed?)         Alt-P   Turn PTT on/off
 Alt-H     Show help.txt                 Alt-W   CW Weight
 Alt-I     Show talk messages            Alt-J   See QRG's
-Alt-K     CW from Keyboard              Alt-X   Exit program (Alt-q)
-Alt-M     Multi display (Alt-c)         Alt-S   Score window (Alt-r)
-Alt-,     Show bandmap                  Alt-J   Show frequency
+Alt-K     CW from Keyboard              Alt-X   Exit program
+Alt-M     Multi display (Alt-C)         Alt-S   Score window (Alt-R)
 Alt-Z     Show worked zones             Alt-G   Grab call (min. 2 letters from bandmap)
+Alt-,     Show bandmap
```
dl1jbe, line 14: Why did you drop `Alt-J`. It still works and is useful in multi op operation.

dl1jbe (commented):
The rest looks good. Go ahead and merge it in.

share/help.txt
zcsahok: It's still there, just dropped the second one and moved Alt-, to the end of the list.
```
   Alt-J   See QRG's
   Alt-J   Show frequency
```

dl1jbe: Silly me. I see, it was there two times. You dropped only the second one.

PR #250, opened 2021-02-19
Extend Multi check for bandmap display

dl1jbe:
Supports all builtin contests besides WPX, PACC_PA and FOC.

* Evaluates worked zones and countries 
* Honor DX_&_SECTIONS, COUNTRY_MULT, ITUMULT and WAZMULT keyword for other contests.

Multi based on exchange field (e.g. worked sections) can not be checked for bandmap display. 

Do not check in the code yet. I will add WPX and PACC shortly (next days).
Please test and give feedback.

### DIFF
--- a/src/bandmap.c
+++ b/src/bandmap.c
@@ -447,17 +447,17 @@ void bandmap_age() {
  *
  * \return true if new multi
  */
-bool bm_ismulti(spot *data, int band) {
+bool bm_ismulti(spot *data) {
 
     if (data == NULL || data->cqzone <= 0 || data->ctynr <= 0) {
 	return false;   // no data
     }
 
     if (contest->is_multi) {
-	return contest->is_multi(data, band);
+	return contest->is_multi(data);
     }
 
-    return general_ismulti(data, band);
+    return general_ismulti(data);
 }
 
 
@@ -594,7 +594,7 @@ void show_spot(spot *data) {
     printw("%7.1f%c", (data->freq / 1000.),
 	   (data->node == thisnode ? '*' : data->node));
 
-    if (bm_ismulti(data, data->band)) {
+    if (bm_ismulti(data)) {
 	attrset(COLOR_PAIR(CB_NORMAL));
 	printw("M");
 	attrset(COLOR_PAIR(CB_DUPE) | A_BOLD);
@@ -616,7 +616,7 @@ void show_spot_on_qrg(spot *data) {
 
     printw("%7.1f%c%c ", (data->freq / 1000.),
 	   (data->node == thisnode ? '*' : data->node),
-	   bm_ismulti(data, data->band) ? 'M' : ' ');
+	   bm_ismulti(data) ? 'M' : ' ');
 
     char *temp = format_spot(data);
     printw("%-12s", temp);
@@ -707,7 +707,7 @@ void filter_spots() {
 	    continue;
 
 	/* Ignore non-multis if we want to show only multis */
-	multi = bm_ismulti(data, data->band);
+	multi = bm_ismulti(data);
 	if (!multi && bm_config.onlymults)
 	    continue;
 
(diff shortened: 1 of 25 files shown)


dl1jbe:
Btw, code needs quite some refactoring, which can/should be done after test.

src/bandmap.c
```diff
@@ -447,20 +447,17 @@ void bandmap_age() {
  *
  * \return true if new multi
  */
-bool bm_ismulti(char *call, spot *data, int band) {
+bool bm_ismulti(spot *data, int band) {
```
zcsahok, line 450: maybe even `band` is not needed if it's always called with `data->band`.

src/setcontest.c
```diff
@@ -26,13 +26,65 @@
...
+    }
+
+    if (country_mult == 1) {
+	return ((countries[data->ctynr] & inxes[band]) == 0);
+    }
+
+    /* TODO: pfxmultab */
+
+    if ((itumult == 1) || (wazmult == 1)) {
+	return ((zones[data->cqzone] && inxes[band]) == 0);
```
zcsahok, line 80: typo? && -> &

src/setcontest.c
```diff
@@ -26,13 +26,65 @@
...
+#include "bands.h"
 #include "focm.h"
 #include "getctydata.h"
 #include "globalvars.h"
 #include "setcontest.h"
 #include "score.h"
 #include "tlf.h"
 
+/* contests where multiplier get determined from comment field can not be
+ shown in bandmap works also for modes with no multiplier at all */
```
zcsahok, line 39: missing comma is sentence?

zcsahok (commented):
:+1:
 
Not directly related:
- I had trouble understanding the code as `inxes[]` holds not indices as the name would suggest, but band mask values. Could we rename it to say `bandmask[]`?
- Do I see correctly that in `addmult()` the individual cases (if's) are mutually exclusive, just as in `score()`?

dl1jbe:
> +1
> 
> Not directly related:
> 
>     * I had trouble understanding the code as `inxes[]` holds not indices as the name would suggest, but band mask values. Could we rename it to say `bandmask[]`?

Yes. That name has gone a long way. Let it finally named right.
> 
>     * Do I see correctly that in `addmult()` the individual cases (if's) are mutually exclusive, just as in `score()`?

I hope so (but are not 100% sure yet). If we can guarantee mutual exclusion the code can be simplified much. 
Btw, one of the main reasons for the switch to contest configs was the stronger  guarantee of mutual exclusion between the different point and multi scoring rules.

One downside of the actual implementation in these PR is that we now have one additional place where the logic for multiplier handling is located. I will try to bring it to a common code base to make sure we have the same eval rules for each case later.

src/setcontest.c
dl1jbe: Good catch, thanks

src/bandmap.c
dl1jbe: Good idea.

src/setcontest.c
dl1jbe: Did a rewording to (hopefully) make it more clear.

zcsahok: Now clear, thanks.

dl1jbe:
Finished handling of wpx style prefixes. Should be complete now.

* Last commit also fixed a segmentation fault if callmaster file could not be found.

dl1jbe:
If all looks well, I would like to reorganize and squash the commits a little bit. Just give me a go if ok

zcsahok (approved):
Thanks!

zcsahok:
I'm just thinking if it could be possible to structure multi handling as it is done for scoring: a contest would define its `get_multi()` functions and the core would just do the "bookkeeping".

dl1jbe:
> I'm just thinking if it could be possible to structure multi handling as it is done for scoring: a contest would define its `get_multi()` functions and the core would just do the "bookkeeping".

That is one of the next goals. It should also be done for showscore() and get/checkexchange().

dl1jbe:
Applied and pushed. Thanks for the feedback.

PR #253, opened 2021-03-07
Added RAEM contest

zcsahok:
image

Added my.lat/lon to the tlf module as scoring needs it. `getstationinfo()` has to be called before initializing the plugin. Plugin's `score()` takes now precedence over built-in logic.

There is an issue with writing/reading point to the .log file: it assumes that score values fit on 2 digits. Upon reading the 3rd digit is cut away, the sum of the points in the example above is clearly not 195.
We could somehow relax this and allow a generic "contest specific" part in a form of something like this
```
5    6         7         8
56789012345678901234567890
exchange   mult mul2 ppp
```
If there are no multipliers (as in RAEM) or no second multi then the points field could be even wider. The width of the fields could be configured.


dl1jbe:
+1In relation to the discussed rescoring at every start we could drop recording of mults and points in the log at all. That would solve also the problem with the truncated comment field.Points and multipliers should be shown anyway for the worked contacts.The future format of the log file has still to be decided.Am 07.03.2021 20:23 schrieb Zoltan Csahok <notifications@github.com>:
Added my.lat/lon to the tlf module as scoring needs it. getstationinfo() has to be called before initializing the plugin. Plugin's score() takes now precedence over built-in logic.
There is an issue with writing/reading point to the .log file: it assumes that score values fit on 2 digits. Upon reading the 3rd digit is cut away, the sum of the points in the example above is clearly not 195.
We could somehow relax this and allow a generic "contest specific" part in a form of something like this
5    6         7         8
56789012345678901234567890
exchange   mult mul2 ppp

If there are no multipliers (as in RAEM) or no second multi then the points field could be even wider. The width of the fields could be configured.



You can view, comment on, or merge this pull request online at:
  https://github.com/Tlf/tlf/pull/253

Commit Summary
add RAEM contest

File Changes

    A
    rules/raem.py
    (45)
  
    M
    src/main.c
    (6)
  
    M
    src/plugin.c
    (59)
  
    M
    src/plugin.h
    (3)
  
    M
    src/score.c
    (4)
  
    M
    src/tlf.h
    (4)
  
    M
    test/test_addcall.c
    (1)
  
    M
    test/test_addmult.c
    (2)
  
    M
    test/test_contest.c
    (3)
  
    M
    test/test_getctydata.c
    (3)
  
    M
    test/test_parse_logcfg.c
    (3)
  
    M
    test/test_score.c
    (2)
  
    M
    test/test_searchlog.c
    (1)
  

Patch Links:
https://github.com/Tlf/tlf/pull/253.patchhttps://github.com/Tlf/tlf/pull/253.diff

zcsahok:
I find it informative for the operator to see the worked mults and the points. We should just not use it for any further evaluation ("write only" fields).

In the .log file we have 24 chars for exchange+multinfo+score. A contest would specify the exact width it needs for these fields. For example RAEM exchanges can be up to 14 chars (xxxx +xxN+xxxO) and the max score needs 3 digits. As there are not mults, this fits easily within the available 24, so it could have exchange width say 20 (probably left aligned) and score width 4 (right aligned).

Coup du REF has one (or two, depending on the modeling) mults + score. The exchanges are on the other hand much shorter, normally either 2 digits (for Dept) or up to 4 (for serial). So with exchange width 5 there would still ample space to store both the multi info and the score.

For re-reading only the exchange is needed, so its width has to be known, of course. For plugin contests this information could be delivered by `setup()` as it is called before reading the log file.

dl1jbe:
Sorry for my last badly formatted reply. I was away and had a smartphone only.

> I find it informative for the operator to see the worked mults and the points. We should just not use it for any further evaluation ("write only" fields).
> 
> In the .log file we have 24 chars for exchange+multinfo+score. A contest would specify the exact width it needs for these fields. For example RAEM exchanges can be up to 14 chars (xxxx +xxN+xxxO) and the max score needs 3 digits. As there are not mults, this fits easily within the available 24, so it could have exchange width say 20 (probably left aligned) and score width 4 (right aligned).
> 
...
I think that is not a good way. 
As long as we stay with the actual log format we will always struggle how to manage and allocate the precious space. Whatever we decide now there will sure be contest with other requirements. So we will work on that part again and again.

I see two ways out:

* Drop the multiplier fields from the log but provide a way to show/print a log with that information. Easy and quick to do but we will have to switch to a still different log format later.
* Decide shortly about a new log file format with variable fields and field sizes which gives us enough 'room' for future work.

PR #254, opened 2021-03-14
Prefix naming

dl1jbe:
Make clear that 'pxstr' stands not for dxcc country prefix but for WPX prefix


zcsahok:
Fine with me, thanks.

PR #255, opened 2021-03-22
Some small fixes and changes

dl1jbe:


src/readcalls.c
```diff
@@ -206,7 +206,6 @@ void count_contest_bands(int check, int *count) {
 int readcalls(void) {
```
zcsahok, line 206: in the future we could consider using `parse_logline()` in  this function (from `writecabrillo.c`)

zcsahok (approved):
I'll merge these and the previous changes to dev_plugin to keep the two branches in sync, if that's ok.

src/readcalls.c
dl1jbe: I am working in that direction. 
The used result struct is not quite ideal as it mixes parse results from loglines and QTC lines. Maybe I can decouple it.

Next step will be better handling of worked stations (collecting all infos about a call in the worked_t stuct). After adding some tests to readcalls() I want to use the parsed struct for addcall and addmult.

dl1jbe:
> I'll merge these and the previous changes to dev_plugin to keep the two branches in sync, if that's ok.

Fine, good idea.

PR #257, opened 2021-03-25
Parse line

dl1jbe:
Introduce a parse function for QSO line entries which returns a clean struct without unneeded parts.
Use it to implement the already existing parse_logline() function.


dl1jbe:
No bad feedback. Merging it in.

PR #258, opened 2021-03-29
Add exchange width for contest

### LINKED ISSUE #234: Long exchanges truncated

zcsahok:
To fix #234: a contest defines the max exchange width. Input field in then limited to this width.

image

Tried to come up with reasonable values for the built-in contest.

As the existing widths (e.g. the default 14) sometimes may be too short or too much, they have to be configurable. Will add it if this one is OK.

### DIFF
--- a/src/cleanup.c
+++ b/src/cleanup.c
@@ -43,7 +43,7 @@ void cleanup(void) {
     mvaddstr(12, 29, spaces(12));
 
     attron(COLOR_PAIR(C_WINDOW));
-    mvaddstr(12, 54, spaces(24));
+    mvaddstr(12, 54, spaces(contest->exchange_width));
 
     attron(COLOR_PAIR(C_LOG | A_STANDOUT));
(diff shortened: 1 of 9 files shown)


dl1jbe:
Fine.  +1

A  minor point which we could fix while we at it:
Entering a comment we now see the allowed input field witdh in different color. Starting a new QSO or drop a just started qso (by pressing ESC some times) drops that display of the allowed comment field. Would be nice to have it always shown.

zcsahok:
OK, I will check it.

zcsahok:
image

Couldn't come up with a reason why exchange was shown for CQWW and ARRLDDX only, so removed lines 145-154. `refresh_comment()` does it anyway for all cases.

dl1jbe:
Very well. Had not though it to be so easy. Thanks.

PR #259, opened 2021-03-29
Move hamlib debug setting before loading configs

zcsahok:
In order to avoid debug output from locator functions appearing while/after loading config.
(RAEM plugin evaluates QRA)

Can be merged to master too.


dl1jbe:
+1 
Applied and pushed.

PR #260, opened 2021-04-07
Fix wrong field size for pfxnummulti[]

dl1jbe:
Field size corresponsd to number of possible prefix numbers (0..9), not the number of bands.


zcsahok (approved):
thanks

Issue #248, opened 2021-02-16
TLF should calculate points and multis on every start


dl1jbe:
On restart with an existing not empty logfile TLF does not calculate points according to the actual contest rules but just reads them from what was recorded in the logfile in former runs (see discussion in Issue #178).

dl1jbe:
Same problem exists for scoring multis.

zcsahok:
At least for the display it would make sense to keep the score/multi info in the log line. (the latter was missing in the screenshot in #247 as I bypassed the native multi handling) Whether to write the info back to the log (update the record) or just keep it in `qsos[]` (in-memory) is to be decided.

dl1jbe:
Hmm. The log line display reads the data from the file not from qsos[] (see `scroll_log()`). So there is some work to be done either in rewriting the log file on every start or in the display code.

zcsahok:
I see. But if qsos[] holds the same data (minus recalculated fields) as the log file, then it could be used to get the lines.
Just an idea.

dl1jbe:
Right. I am just looking to make sure it has no side effects so that we do not break something different.

dl1jbe:
Some further thoughts:

* As soon as we calculate points and multis on every start writing them in the log file makes no longer sense. That can be gone in the long run, but having a different log format would be a severe change in TLFs external behavior. The same goes for the automatically recalculation on every start.

* The planned change has a second effect. The pure log file without points and multis are worthless without the corresponding CTY.dat and rules file. That would make the way some people organize their contests (all logs in a single directory) nearly useless - updating any of the files would change the contest results for ANY contest in that directory.

Both points - changed log file and dependency on a complete set of related files - is a serious change of TLFs behavior and usage. Such a change is normally related to a major version change to TLF-2.x and should not done `on the fly` in a TLF-1.5.x or something. And if we want to go that way to a new major version (I am not against it) we should make sure to incorporate ALL the serious changes we already see now (Multi handling, external scripting, .....). Otherwise we are on the way to TLF-3.x, 4.x, 5.x  ... shortly after.

airween:
_As soon as we calculate points and multis on every start writing them in the log file makes no longer sense._ - agree, but I'm not afraid this can makes any problem. Just write back the correct points and multis to logfile. To make a backup during the re-score event can avoid the data loss.

ha5se:
Let me add some more aspects:

[...]

PR #261, opened 2021-04-07
Evaluate on start

dl1jbe:
This is a first part to fix issue #248.

Besides some cleanup, simplification and restructuring it changes addcall() to accept a qso_t record of relevant 
information. That gets collected by a new function collect_qso_data() from the bunch of global variables.
Long term goal is to use the same addcall() function for evaluation of newly made QSOs and of contacts red from the log file during readcalls().

Furthermore the patch uses parse_qso() to get these information from QSOs coming in from LAN (in addcall2() ) or from reading the already written log file (in readcalls() ).

Next steps will be the use of the qso_t record for addmult() and maybe makelogline(). Furthermore the parsed qso_t struct and the evaluation results for all qsos should be kept in memory instead( or in complement) of the actual qsos[] array which stores only the plain log lines and needs a reevalution every time some value is needed from them.

### DIFF
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -62,6 +62,19 @@ int excl_add_veto;
  * addcall2() is need to separate the points and multipliers.
  */
 
+
+/* collect all relevant data for the actual QSO into a qso_t structure */
+struct qso_t *collect_qso_data(void) {
+    struct qso_t *qso = g_malloc0(sizeof(struct qso_t));
+    qso->call = g_strdup(hiscall);
+    qso->mode = trxmode;
+    qso->bandindex = bandinx;
+    qso->timestamp = get_time();
+    qso->comment = g_strdup(comment);
+    return qso;
+}
+
+
 int addcall(void) {
 
     int cty, zone = 0;
--- a/src/addcall.c
+++ b/src/addcall.c
@@ -119,7 +119,7 @@ int lookup_country_in_pfxnummult_array(int n) {
 int addcall(struct qso_t *qso) {
 
     int cty, zone = 0;
-    int add_ok;
+    bool add_ok;
     int pfxnumcntidx = -1;
     int pxnr = 0;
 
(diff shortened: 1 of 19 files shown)


dl1jbe:
There are some severe changes. Could use a thorough review.

src/addcall.c
```diff
@@ -112,96 +163,53 @@ int addcall(void) {
...
-			    == 0) {
-			pfxnummulti[pfxnumcntidx].qsos[pxnr] |= inxes[bandinx];
-			addcallarea = 1;
-			countryscore[bandinx]++;
-			zonescore[bandinx]++;
-		    }
-		}
-		break;
 
+    if (add_ok == 1) {
```
zcsahok, line 191: `add_ok` could be a boolean

src/addcall.c
```diff
@@ -290,70 +287,39 @@ int addcall2(void) {
...
-			zones[zone] |= inxes[bandinx];
-		    }
-		    break;
-
+	    if (pfxnumcntidx < 0) {
+		if (cty != 0 && (countries[cty] & inxes[bandinx]) == 0) {
+		    countries[cty] |= inxes[bandinx];
+		    countryscore[bandinx]++;
+//                  new_cty = cty;
+	    }
```
zcsahok, line 310: strange indent, maybe astyle is need

src/readcalls.c
```diff
@@ -240,20 +203,22 @@ int readcalls(void) {
...
+	update_worked(station, qso);
 
-	countrynr = getctydata(presentcall);
 
 	if (continentlist_only) {
-	    if (!is_in_continentlist(continent)) {
+	    if (!is_in_continentlist(dxcc_by_index(countrynr)->continent)) {
 		qsos_per_band[bandindex]++;
+		free_qso(qso);
+		qso = NULL;
```
zcsahok, line 221: is this needed? or just a preventive measure in case `qso` would be used later on?

src/tlf.h
```diff
@@ -189,12 +190,14 @@ typedef struct {
 struct qso_t {
     char *logline;
     int band;
+    int bandindex;
```
zcsahok, line 193: (can't comment on unchanged line)
```char exchange[13];```  should be 24 ((77-55+1)+1) to be able to store a QSO-mode exchange

zcsahok (approved):
Found mostly just minor stuff. Thanks.

src/addcall.c
dl1jbe: Thanks. Just fixed.

src/addcall.c
dl1jbe: Good catch. Vi's autoindent had difficulties with the commented out line above.

src/readcalls.c
dl1jbe: Not really. But it helps to catch 'use after free' errors during development.

src/tlf.h
dl1jbe: You mean in declaration of worked_t? You are right - corrected.

zcsahok (approved):
Thanks, will merge into dev.

PR #263, opened 2021-04-10
Fix Cabrillo keywords in rules

### LINKED ISSUE #262: Keyword 'CABRILLO-MODE' not supported error for default cqww contest
After a pull and recompile I've started to get the following error when using the default logcfg:

Keyword 'CABRILLO-MODE' not supported error for default cqww contest

If I remove that from the cqww rules file TLF starts up ok. Version from "git describe" is "tlf-1.4.1-165-g080f551".

zcsahok:
Fixes #262


dl1jbe:
+1 Applied and pushed

PR #265, opened 2021-04-25
Convert own callsign to upper case

zcsahok:
Just in case one writes it in lower case... Search in cty.dat works for upcase only.


dl1jbe:
Applied and pushed.

PR #269, opened 2021-06-23
Fix when using a call with slash (e.g. portable)

zcsahok:
Having a call containing a slash causes two issues:

1. Cabrillo file can't be saved (or imported) as '/' is not a valid file name character.
2. Lookup of own call fails on startup. Trailing NL interferes somehow.

The fix drops the legacy trailing NL and in Cabrillo file name slash is replaced by underscore.

### DIFF
--- a/test/test_sendbuf.c
+++ b/test/test_sendbuf.c
@@ -87,7 +87,7 @@ int setup_default(void **state) {
     trxmode = CWMODE;
     cwkeyer = 1;
     digikeyer = 1;
-    strcpy(my.call, "dl1jbe\n"); 		// !!! do not forget trailing \n
+    strcpy(my.call, "dl1jbe");
     strcpy(hiscall, "lz1ab");
     strcpy(sent_rst, "579");
     shortqsonr = LONGCW;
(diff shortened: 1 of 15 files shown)


test/test_sendbuf.c
```diff
@@ -87,7 +87,7 @@ int setup_default(void **state) {
     trxmode = CWMODE;
     cwkeyer = 1;
     digikeyer = 1;
-    strcpy(my.call, "dl1jbe\n"); 		// !!! do not forget trailing \n
+    strcpy(my.call, "dl1jbe");
```
airween, line 90: Don't we want to test here the call sign with `/`?

airween (approved):
I reviewed, have only one question - see the comment.

Btw, looks good to me.

test/test_sendbuf.c
zcsahok: Pushed the updated test of getctydata(). The original issue (the seconds one) can be reproduced by adding a trailing NL to the calls on lines 175 and 178 of `test_getctydata.c`. The test will fail on line 178, i.e. "LA3BB\n" leads to the same (and correct) value as for "LA3BB/QRP", but "LA3BB/P\n" delivers zero.
The correction was done by dropping the legacy trailing NL, not by changing the function.

PR #275, opened 2021-08-31
Rework checkexchange()

zcsahok:
Work in progress on `checkexchange()` in order to have a clear path for log re-evaluation upon startup or rescoring.

Started with ARRL SS contest. There are some issues caused by the makeshift regex engine demonstrated by the test added at  fe36abfff. E.g. partial section sometimes taken for precedent value
image
Or `sectionpats[7]` has an extra _i_ that makes this entry overflow into the next one.

I'll tackle CQ WW next. Does anyone has a hint regarding the use of zone_fix and zone_export? I've seen the comment on this and agree that it's quite obscure.


dl1jbe:
The change to a generic regex parser is really overdue. Good work.

Comparing with the old code the new one seems to be more restrictive. You have to input the fields in just the right order and it seems the spaces between the fields are required now.

Citing from Rein's old web page: 
'For the arrl_ss contest the exchange input is intelligent, incl. call update in the exchange field.E.g. in SS you can enter 23A 98OH and tlf knows what you mean.'

I am not sure how a typical station handles Sweep Stake input in reality - it seems there is a need to allow a more relaxed input. Maybe Nate can provide some background.

Anyway, I would not care too much for that  at the moment. Seems you started with the most complex case.

N0NB:
IMHO, CT really had an elegant way of dealing with SS and other events by having separate input fields for all elements.  <Space> was the magic "tab" key that moved between relevant input fields.  Tr's two field approach and intelligent deciphering always seemed to me to be prone to royal screwups, though never having used Tr I am speaking well beyond my base of knowledge!

Enough of that.

Zoli, are you validating in real time or only when the op signals to log the QSO?  I think if the later the confusion might be lessened.  As an example, if the exchange is supposed to be `84 A N0NB 83 KS` but it is typed in as `A 83 84 KS N0NB`, how are we supposed to know which number is the serial and which is the check?  This may be why I have seen the precedent directly appended to the serial number to avoid such confusion, `84A N0NB 83 KS`.

In SS the serial number is variable from 1 to 4 digits and the check is always two digits.  Ambiguity exists for the entire range where the serial number is two digits, though many ops send cut leading zeroes, it is not mandatory.

zcsahok:
Thanks for the feedback!

I have relaxed somewhat the pattern and moved some validations to the code.  Now even a comment like `123AK1ABC73OH` can be parsed (into "123 A K1ABC 73 OH"). Of course this is generally ambiguous, as this could also be "123 AK1ABC 73 OH" if the user didn't enter the precedent. Added a bunch of tests for various corner cases. Without an explicit grouping using e.g. spaces it is impossible to always figure out the actual exchange parts.

As I read the rules the exchange has to be in fixed form, like "43 M W1AW 31 CT". Typing the fields in a different order is not supported. I don't see an issue with this, but of course I'll probably never participate in SS.

CALLUPDATE is an undocumented feature used only for SS, CQWW and other zone-based contests. @N0NB may I ask you to add a line on this to the man page?

zcsahok:
`checkexchange()` is called on each key press. It parses the comment entered so far and stores data in variables like `ssexchange`, `callupdate` or `zone_fix`/`zone_export`. `callupdate` is used on-the-fly, whereas `ssexchange` replaces the original exchange in the log once the QSO has been logged. I plan to rename `ssexhange` to something like `normalized_comment` as this feature shall also be used for CQWW and other contests.

N0NB:
> Thanks for the feedback!
> 
> I have relaxed somewhat the pattern and moved some validations to the code. Now even a comment like `123AK1ABC73OH` can be parsed (into "123 A K1ABC 73 OH"). Of course this is generally ambiguous, as this could also be "123 AK1ABC 73 OH" if the user didn't enter the precedent. Added a bunch of tests for various corner cases. Without an explicit grouping using e.g. spaces it is impossible to always figure out the actual exchange parts.

Agreed.  Of course those of us who know the arcane rules surrounding USA callsigns will immediately notice that AK1ABC is not a legal US callsign.  Only the A precedent is part of a legal US callsign prefix, so the others, Q, B, U, M, and S are invalid as the first letter of a US callsign.  All could be a part of the two or three letter section, however.

> As I read the rules the exchange has to be in fixed form, like "43 M W1AW 31 CT". Typing the fields in a different order is not supported. I don't see an issue with this, but of course I'll probably never participate in SS.

Correct.  On the air some stations try to short-circuit this a bit by leaving out their callsign part of the exchange.  I even ran across a program, perhaps this one (which is forgivable as it was originally more EU centric), that left it out as part of the default macro.  Over the years there have been pointless arguments on mailing lists over this!  It is stated in the rules as such and as such must be sent in the correct and full order or the exchange is incomplete.

However, a logging program is not obligated to follow the order exactly, though it must generate the Cabrillo file correctly.  CT had an interesting input method.  On screen the fields were S/N, P, Call, Check, Sect.  For a new QSO the cursor was in the Call field, from there pressing space cycled the cursor to S/N, P, Call, Check, Sect. then back to Call.  It actually worked well until you got one field off!

> CALLUPDATE is an undocumented feature used only for SS, CQWW and other zone-based contests. @N0NB may I ask you to add a line on this to the man page?

Sure.  I'm working through the man page updates now so I'll add it.

dl1jbe:
Maybe we need not go as far as to eliminate all requirements for spaces. How about some simple:

- The fields have to be in the request order (S/N, precedent, call, check, section)!
- The callsign (and only it) has to be surrounded by spaces.

That eliminates any uncertainty which number is S/N and which is check. Furthermore it fixes any problems with the recognition of the callsign. And it allows some freedom with regards to the needed spaces.

N0NB:
I would say there is little need to duplicate the received call in the exchange field.   Yes, it is part of the SS exchange as given on the air, and yes, the Cabrillo file must be formatted correctly, but aside from that, it's okay if the exchange field only contains `001 A 83 KS`, for example and then it is all combined later.

For our TX, the only part that changes is the serial number.  The rest is a static value, e.g. `# A % 83 KS`will be the F3 TX macro, etc.

zcsahok:
I think parsing is now reasonably well-behaving. What worries me more is that for some reason the call is removed from the exchange during normalizing (it was like this before, as on my first screenshot) and thus Cabrillo file doesn't contain the full exchange. Would such a Cabrillo be accepted by the organizers?

For CQWW spaces will be a must, as there it would be very hard to separate zone numbers and calls.

N0NB:
Probably everything you wanted to know about November Sweepstakes and Cabrillo:  http://www.arrl.org/cabrillo-format-tutorial

zcsahok:
Thanks for the link. An example QSO line looks like
```
QSO: 14000 CW 2009-11-07 2100 W1AW          1 M 38 CT K8MM        1 Q 92 MI
```
so the call must be removed from the exchange, just as we do it. Then we are all OK for ARRL SS.

dl1jbe:
Fine. Just merge it in Zoli.

zcsahok:
I'll go on with CQWW and other use cases covered in `getexchange()`.

dl1jbe:
If I remember correctly some checks ultimatively require that you have the correct exchange format before it allows to log the qso. Some others are more relaxed as they only warn but let you log the contact anyway. In the moment the behavior is hard coded, but it may be better to have a general keyword to control that behavior (e.g. STRICT_CHECK or similar).

zcsahok:
Yes, checks happen upon leaving the input field. These I'd keep as-is for the moment (but let's see :-) ). My ultimate goal is to make it explicit that `checkexchange()` in some cases determines the multiplier value that shall be later used in `addmult()`.

N0NB:
> Thanks for the link. An example QSO line looks like
> 
> ```
> QSO: 14000 CW 2009-11-07 2100 W1AW          1 M 38 CT K8MM        1 Q 92 MI
> ```
> 
> so the call must be removed from the exchange, just as we do it. Then we are all OK for ARRL SS.

Well, it is still there, though out of order.  The reporting format is not quite reflective of the required on-the-air format.

dl1jbe:
Yes, at the moment there is to much magic happening before we can score the multis. So it is a good idea to make it explicitAm 09.09.2021 08:15 schrieb Zoltan Csahok ***@***.***>:
Yes, checks happen upon leaving the input field. These I'd keep as-is for the moment (but let's see :-) ). My ultimate goal is to make it explicit that checkexchange() in some cases determines the multiplier value that shall be later used in addmult().

dl1jbe:
I did some code reading again. If I understand it right then zone_export gets generated from the callsign by getctydata(). Zone_fix on the otherhand gets parsed from the comment line. It seems the idea was to allow to override the zone (zone_export) derived from the callsign by entering another one as comment. Not sure if it works at all.. Am 31.08.2021 20:38 schrieb Zoltan Csahok ***@***.***>:makes this entry overflow into the next one.
 Does anyone has a hint regarding the use of zone_fix and zone_export? I've seen the comment on this and agree that it's quite obscure.



You can view, comment on, or merge this pull request online at:
  https://github.com/Tlf/tlf/pull/275

Commit Summary
add test_getexchangerework checkexchange for ARRL_SSfix callupdate size

File Changes

    M
    src/calledit.c
    (2)
  
    M
    src/callinput.c
    (6)
  
    M
    src/getexchange.c
    (288)
  
    M
    src/initial_exchange.h
    (3)
  
    M
    src/tlf.h
    (3)
  
    M
    test/data.c
    (5)
  
    M
    test/test_addcall.c
    (2)
  
    M
    test/test_addmult.c
    (4)
  
    M
    test/test_cabrillo.c
    (4)
  
    M
    test/test_getctydata.c
    (4)
  
    A
    test/test_getexchange.c
    (160)
  
    M
    test/test_readcalls.c
    (8)
  
    M
    test/test_score.c
    (4)
  
    M
    test/test_searchlog.c
    (7)
  

Patch Links:
https://github.com/Tlf/tlf/pull/275.patchhttps://github.com/Tlf/tlf/pull/275.diff

zcsahok:
I'm trying to figure out what might have been the original intention showing a zone number in lower right corner of the worked display.
image
In this example K1PDL gave 06 and then any DL or PD station will have 06 in the display.
But once Enter is pressed the "correct" exchange  is taken:
image
So what's the point of showing 06?

Would the following make sense?

1. if still in call field and exchange field is empty, then show the zone determined from previous QSOs or other sources (the "correct" value that will be taken on Enter).
2. else show whatever can be parsed from the exchange field, just in case of ARRL SS.

(Point 1 could be even dropped as is an typical contest situation one hits Enter right after entering the call without checking the worked window)

src/getexchange.c
```diff
@@ -333,7 +332,15 @@ int getexchange(void) {
 		CONTEST_IS(CQWW) ||
 		CONTEST_IS(STEWPERRY)) {
 
-	    checkexchange(comment);
+	    checkexchange(comment, true);
+
+	    if (call_update && strlen(callupdate) >= 3) {
+		strcpy(hiscall, callupdate);
+		mvprintw(12, 29, "%s", spaces(MAX_CALL_LENGTH));
```
dl1jbe, line 339: How about using 'printcall()' for these two lines?

src/getexchange.c
```diff
@@ -615,277 +770,34 @@ void checkexchange(char *comment) {
...
-	}
-
-	if (strlen(hiscall) >= 2)
-	    OnLowerSearchPanel(32, zone_export);
-
-
-	for (ii = 0; ii < LEN(callpats); ii++) {
-
-	    hr = getlastpattern(callpats[ii]);	// call update ?
+    normalized_comment[0] = 0;
```
dl1jbe, line 773: These two lines are redundant. See above at line 737 and 738.

src/getexchange.c
```diff
@@ -615,277 +770,34 @@ void checkexchange(char *comment) {
...
+    // ----------------------serial+section--------------------------
+    if (serial_section_mult) {
 
-	return;		// end arrlss
+	checkexchange_serial_section(comment, interactive);
+	return;
     }
 
-    // ----------------------serial+section--------------------------
     if ((serial_section_mult == 1) || (sectn_mult == 1)
```
dl1jbe, line 797: 'serial_section_mult' got handled above (line 793). Should be dropped here.

src/getexchange.c
zcsahok: Good point. I see know that call update interferes with autosend.

src/getexchange.c
zcsahok: Yes, forgot that I added them already.

src/getexchange.c
zcsahok: It's still work in progress, will drop all unneeded code once ready. It's just kept for reference. Next commits will cover sectn_mult + sectn_mult_once.

test/test_addmult.c
```diff
@@ -313,14 +315,23 @@ void test_arrlss(void **state) {
...
 
     setup_multis("SC\nSCV\n");
-    strcpy(ssexchange, "SCV");
-    set_this_qso("");   // NOTE: ssexchange is not part of qso_t
+    strcpy(mult1_value, "S");   // incomplete value (normally doesn't happen)
+    set_this_qso("");   // NOTE: mult1_value is not part of qso_t
     addmult(this_qso);
-    strcpy(ssexchange, "97A23SCV");
+    assert_int_equal(nr_multis, 0);
+#if 0
```
zcsahok, line 322: @dl1jbe Tom, could you check this? Here an invalid mult ("SCX") is provided but it counts as "SC". (the test fails, hence the commeting out) This is related to the looking for the longest `get_matching_length()` in `addmult`. Is this really correct?

src/getexchange.c
dl1jbe: Was just a reminder from my side...

test/test_addmult.c
dl1jbe: Will do, but give me a day

dl1jbe: Original code from Rein picked up first matching multi (even in an exchange without spaces) - so if SC and SCX are in that order in the list of possible mults it choose SC even SCX where in the exchange. It got changed around '18 to pick up the longest matching multi. But it seems that having a wrong multi in the exchange got not handled properly.

In your actual code you have the section as standalone string in 'mult1_value'. With that we should convert the handling in 'addmult()' from finding the longest match to a check for a full match ('get_exact_mult_index()'). That would be better and fixes your problem in the test code.

zcsahok: Pushed fix. Probably `get_matching_length()` could be dropped later on.

src/getexchange.c
N0NB: I guess I need to set up for SS and play a bit to see the changes.

N0NB:
Just pulled in this branch and built it and entered a few fake QSOs and it seems to work.

One thing, is that now in place of the RST values are three dashes, `---`, I think it would be better just to print spaces to the screen instead.  Doing so may avoid confusion when the `NO_RST` keyword is used.

Other than that, Tom, please merge this so that we can announce the changes on the list as the SS SSB weekend is just over a week away and others can test this a bit more.

zcsahok:
Thanks for checking, Nate. Could you post a screenshot with dashes? I'll check it then.

N0NB:
Screenshot from 2021-11-12 15-28-41

zcsahok:
I see, it's in `makelogline.c`
```
    if (no_rst) {
        strcat(logline4, "---  ---  "); /* instead of RST */
```

src/getexchange.c
zcsahok: Added to b6ec0f8

dl1jbe:
@Zoli: Looks like you are done with the major part. Can we merge it in (see Nates comment above)?

zcsahok:
I tested it in the OK-OM contest, so yes, it can be merged.

Next I'll look into re-evaluating multis during reading the log.

There is one open issue: the length of section is currently 4 which is less than the length of special DOKs (6). So 4 shall be increased to 6 in a backward compatible way.

dl1jbe:
> Next I'll look into re-evaluating multis during reading the log.

Ok. I did some first look into that during summer holidays but got not far. 

The code to evaluate the red lines in readcalls() are similar, but not quite the same as in addcall() and addmult(). It seems desirable to have one common set of functions which evaluates the qso line during start and also during contest. 

From my summer experiments I found some minor changes that may be helpful: see https://github.com/dl1jbe/tlf/tree/addcall_improvements. Feel free to pick up what seems appropriate.

PR #276, opened 2021-09-01
Fix stopping voice keyer messages

dl1jbe:
Start voice keying in an separate process so TLF can look for ESC key and send the play_vk script an TERM signal.

To allow the script to handle that signal the actual sound playing has to be run in background by the added '&'.

You have to adapt the autocq timing as the time the sound is running does no longer add to that time.

Please review and test that sound is stopped and rig switched back to receive mode.


dl1jbe:
> For the SoX version of `play` this line should be:
> 
> ```
> play -q $1 &
> ```


Thanks for the hint. According to man page you are right. 

I do have Sox-14.4.2 here and de facto both version do work. I will change it anyway.

N0NB (commented):
Sorry I missed this.

I've yet to try it but will.  I noted the syntax for SoX `play` in the commit for `play_vk`.

N0NB:
I actually have these lines in my local `play_vk`:

``` shell
# play file on the default sound device

# OSS play utility syntax:
#play $1 -q 2>/dev/null

# SoX play utility syntax
play -q $1
```

scripts/play_vk
```diff
@@ -28,7 +28,7 @@ trap 'kill -s INT $PID' TERM INT
 
 # play file on the default sound device in backgroudn, so signals can be catch
 # by the 'trap' instruction above
-play $1 -q 2>/dev/null &
+play -q $1 2>/dev/null &
```
N0NB, line 31: Locally I found that redirecting `stderr` to `/dev/null` was not necessary with the `-q` option properly placed.  Please test that, Tom.

dl1jbe: > Sorry I missed this.
> 
Just found the following in 'man play':

"These options can be specified on the command line at any point  before the first effect name."

That may explain the behavior. With -q before the file name it is quite more robust. Thanks.

dl1jbe: > Locally I found that redirecting `stderr` to `/dev/null` was not necessary with the `-q` option properly placed. Please test that, Tom.

It may be left over from former versions of SoX which left an error message when killed. Just tested it with 'aplay -q' - that would need the redirection.

N0NB: I was basing my placement of `-q` on the output of `play -h` which provides a usage output.  Yes it says that gopts can be specified any place before the first effect.  Earlier in the output is this line:

```
Usage summary: [gopts] [[fopts] infile]... [fopts] [effect [effopt]]...
```

So I presumed it would be best to place the `-q` in front of the filename.  However, just testing it now I see it is honored in either place on the command line, so disregard my "advice".  :)

dl1jbe: I just tested the aplay command. But as agreed let us stay with SoX as 'official' supported.

And sure there has to be more documentation. At the moment it is just to test that the sound gets stopped and rig is going back to RX mode.

N0NB: I am going through the man page and have noted that area as one that will need to be updated.

dl1jbe:
Please test and report back again. If no problems we can add some keywords to allow users to specify the commands to use.

zcsahok:
It does not stop playing on my system (Debian 11 i386). I'm using a plain play_vk script with a single line
```
play $1 -q 2>/dev/null
```
Upon Esc "Terminated" is written into the call input field (after the current cursor) and this text can't be cleaned by further Esc's, just by typing it over. Playback continues and terminates normally (not interrupted).
Looks like pkill only kills the play_vk process itself, but the actual play command keeps running.
Replacing the pkill with
```
pkill -SIGTERM -n play 2>/dev/null
```
fixes both issues (playback is terminated and there is no extra output). But "play" seems to a bit too generic to me, easy to inflict collateral damage...

N0NB:
While it may seem that `play` is generic, apparently it doesn't conflict in Debian or they would rename it (that happened some years with the `node` program in the ax25 packages).  Also, as it is a command line program, I doubt that it would be used at the same time `tlf` is used, at least on the same computer.

To be absolutely certain, its PID or the script's PID needs to be known to kill the correct process that `tlf` spawned and then kill by PID.

zcsahok:
If PID is know then
```
pkill -9 -P <PID>
```
seems to do the right job. (tested in a terminal)

dl1jbe:
Hi Zoli, you need to use the changed play_vk from the PR. It catches the sigterm and in turn kills the play command. Am 09.09.2021 19:22 schrieb Zoltan Csahok ***@***.***>:
It does not stop playing on my system (Debian 11 i386). I'm using a plain play_vk script with a single line
play $1 -q 2>/dev/null

Upon Esc "Terminated" is written into the call input field (after the current cursor) and this text can't be cleaned by further Esc's, just by typing it over. Playback continues and terminates normally (not interrupted).
Looks like pkill only kills the play_vk process itself, but the actual play command keeps running.
Replacing the pkill with
pkill -SIGTERM -n play 2>/dev/null

fixes both issues (playback is terminated and there is no extra output). But "play" seems to a bit too generic to me, easy to inflict collateral damage...

N0NB:
If I understand things, `system()` doesn't return the PID.  A more manual method is required, I think.

dl1jbe:
It is difficult to get the pid of a process started by system(). Therefore we use the pkill command. There is one switch to kill only the last started command by pkill - i just do not remember...Am 09.09.2021 19:54 schrieb Zoltan Csahok ***@***.***>:
If PID is know then
pkill -9 -P <PID>

seems to do the right job. (tested in a terminal)

N0NB:
It's the `-n` option, Tom.

N0NB:
https://manpages.debian.org/bullseye/procps/pkill.1.en.html

zcsahok:
OK, with the new `play_vk` it does stop indeed, all is fine.
Just users need to be reminded to fix their scripts in case they don't use the distributed ones.

dl1jbe:
I will prepare more documentation in next time. I just needed a confirmation for correct function before continuing in these direction.To be honest the play_vk is not really needed - the mechanism can work with scripts or with plain binary programs (e.g. play). More on that later (using just my phone for that conversation is just annoying....)Am 09.09.2021 20:14 schrieb Zoltan Csahok ***@***.***>:
OK, with the new play_vk it does stop indeed, all is fine.
Just users need to be reminded to fix their scripts in case they don't use the distributed ones.

zcsahok (commented):
added 2 comments

scripts/play_vk
```diff
@@ -21,11 +21,17 @@
...
 # logcfg.dat file) can be adapted and has precedence over the standard install
 
+trap 'kill -s INT $PID' TERM INT
+
 # mute mic
 #amixer -c 0 set Mic mute -q
 
-# play file on the default sound device
-play $1 -q 2>/dev/null
+# play file on the default sound device in backgroudn, so signals can be catch
```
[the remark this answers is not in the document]
N0NB: s/backgroudn/background/
s/catch/caught/

N0NB:
There is a slight delay if I hit F1 and then Escape immediately afterward but this much, much better than we had and than what my hack was.

N0NB (approved):
LGTM

dl1jbe:
Zoli, Nate, thanks for the spell check.Atomic_bool is a good idea. I fear the small delay in stopping is difficult to avoid as we are running some threads and real scheduling is controlled by the kernel only.Btw, the delay is shorter than for the cw keying here.Am 10.09.2021 03:36 schrieb Nate Bargmann ***@***.***>:
There is a slight delay if I hit F1 and then Escape immediately afterward but this much, much better than we had and than what my hack was.

zcsahok:
> There is a slight delay if I hit F1 and then Escape immediately afterward but this much, much better than we had and than what my hack was.

Try killing with SIGKILL in `play_vk`:
```
trap 'kill -s KILL $PID' TERM INT
```
For me it resulted in an immediate stopping.

N0NB:
It doesn't seem to work here, Zoli.

Screenshot from 2021-09-10 20-48-39

src/main.c
```diff
@@ -413,6 +413,9 @@ int bandweight_multis[NBANDS] = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0};
 pthread_t background_thread;
 static struct termios oldt, newt;
 
+bool vk_running = false;
```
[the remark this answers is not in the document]
dl1jbe: I am not sure if i is needed hee, but using  atomic_bool seems a good idea. 

Btw, there are a lot more places where we have to care for thread safety. As thre were more urgent problems thread safety problems got neglected for quite some time.

scripts/play_vk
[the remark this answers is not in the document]
dl1jbe: fixed

src/main.c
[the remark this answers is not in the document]
dl1jbe: See last commit

scripts/play_vk
```diff
@@ -21,13 +21,14 @@
...
 
+# accept SIGTERM and SIGINT signals and kill play command as soon as possible
 trap 'kill -s INT $PID' TERM INT
 
 # mute mic
 #amixer -c 0 set Mic mute -q
 
-# play file on the default sound device in backgroud, so signals can be caught
-# by the 'trap' instruction above
+# play file on the default sound device in backgroud, so signals can be
```
zcsahok, line 30: typo (backgroud)

src/audio.c
```diff
@@ -273,3 +284,85 @@ void record(void) {
...
+    /* play sound in separate thread so it can be killed from the main one */
+    if (pthread_create(&vk_thread, NULL, play_thread, (void *)audiofile) != 0) {
+	    TLF_LOG_INFO("could not start sound thread!");
+    }
+}
+
+#define NO_KEY -1
+
+void stop_vk() {
+    IGNORE(system("pkill -SIGTERM -n play_vk2"));
```
zcsahok, line 349: typo (play_vk2)

src/audio.c
```diff
@@ -273,3 +284,85 @@ void record(void) {
 	}
     }
 }
+
+
+/* playing recorded voice keyer messages */
+void *play_thread(void *ptr) {
```
zcsahok, line 290: With playing in background one can start many parallel playbacks by hitting Fn several times.
This is more a bug, than a feature.

scripts/play_vk
```diff
@@ -21,13 +21,14 @@
 # An executable copy of 'play_vk' in your contest directory (containing the
 # logcfg.dat file) can be adapted and has precedence over the standard install
 
+# accept SIGTERM and SIGINT signals and kill play command as soon as possible
```
zcsahok, line 24: For me using `-s INT` gives a slight (around 0.5 s) delay, before stopping playback.
`-s KILL` stops playback immediately.

zcsahok (commented):
Added some comments.

src/audio.c
dl1jbe: Without running it in background we are not able to see the ESC key, so we can not avoid that.
What we can do is reject (simply ignore) new play commands as long as the old one is running.

scripts/play_vk
dl1jbe: Sure SIGKILL kills the execution of the script directly by the kernel while SIGINT will be catched by the script and sent to the play command which in turn has to catch the signal and then stops the sound.

I fear the downside of SIGKILL will be that the script gets no chance to do an work you may want it to do AFTER playing the sound. I will check it in the evening.

Comparing it with stopping a CW message the delay for SSB is shorter here.

src/audio.c
dl1jbe: Thanks.

scripts/play_vk
dl1jbe: After a quick check I have to confirm some of my fears about the use of SIGKILL. It can not used in stop_vk(). 

But we can use it in the trap line inside the play_vk command. That seems to fix the delay but keeps the script running.

zcsahok: The latest play_vk and code works just fine for me.

zcsahok (approved):
All OK.

dl1jbe:
Very well. Let me do some fine tuning (comments, better naming, maybe documentation,..) before I merge it in, Should be done in next one or two days.

dl1jbe:
It was more than the expected one or two days. I took the chance to rewrite most of the audio handling code to cleanup functionality and improve readability. Furthermore the following changes were made.

* Dropped the no longer needed SC_DEVICE keyword
* Added user configuration for commands to record and play back voice key messages and soundlog.
  (see VK_PLAY_COMMAND, VK_RECORD_COMMAND, SOUNDLOG_PLAY_COMMAND and   
  SOUNDLOG_RECORD_COMMAND)
* Added user configuration for the directory containing soundlogs (SOUNDLOG_DIRECTORY).

I would appreciate a review and feedback for the changes so far.

Open points to do and to discuss:
* Documentation (my part)
* Should we also make the file ending (and therefore the audio file type) configurable? It is still fixed
  to Sun au format. Other parts of TLF use different formats.
* Soundlog recording could use an automatic filename suggestion (at the moment still done via the  
  'soundlog' shell script). Do we keep the existing format (ddhhmm) or should we choose a different one?
* The actual 'soundlog' script supports starting a new record any hour or so. Should we keep that in the 
  script or code it into TLF.
* Do we really need the command to replay the soundlog files? I would suggest to drop the command at 
  all.
  It seems more appropriate to use an external player which allows to interactively move to interesting 
  points in the file, to replay some parts again and so on.  There are good solutions out there (graphical 
  and also with text UI).

dl1jbe:
To complete the started PR and to fix the existing conflicts I did a rebase of the code onto master and split it in two parts.

* First part enables a really quick stop of running voice keyer messages via stop_tx.
* The second part has most of the audio code rewritten and makes all sound recording and playing user configurable. It will also move soundlogs to ./soundlogs directory by default.

I will add pull requests for both, link back to these PR for reference and close these PR afterwards.

zcsahok (approved):

dl1jbe:
All code from these PR got moved to PR #317 and PR #318. Therefor closing these request now.

Issue #266, opened 2021-05-30
Info line going missing


dh5fs:
version: test with newest master build, in WPX
whats wrong: in ESM mode after entering the call, the lowest line with (distance heading etc.) disappears, this can be annoying when you need to turn the beam because Sig went down. Does not happen when changing with space to the number field.
how it should be: line just stays until number was entered

dl1jbe:
That behavior is already seen at least since tlf-1.4.1. But you are right, it would be better to keep that information visible  as long as possible. 
I will look into it.

dl1jbe:
As #279 is merged in let us close the issue now.

PR #279, opened 2021-09-11
Fix clearing info line by keyer

zcsahok:
A fix for #266
Sending messages calls `displayit()` down the line and it then calls `clear_display()` in turn.
Simply added re-displaying DX info to `displayit()`. Not sure if it's the best fix, though.
Also added clearing display after logging a QSO so that a new one starts with a fresh view.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,16 @@ before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y libhamlib-dev
   - sudo apt-get install -y libxmlrpc-core-c3-dev
-  - wget https://cmocka.org/files/1.0/cmocka-1.0.1.tar.xz
-  - tar -xvf cmocka-1.0.1.tar.xz
-  - cd cmocka-1.0.1
+  - wget https://cmocka.org/files/1.1/cmocka-1.1.5.tar.xz
+  - tar -xvf cmocka-1.1.5.tar.xz
+  - cd cmocka-1.1.5
   - mkdir build
   - cd build
   - cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Debug ..
   - make
   - sudo make install
   - cd ../..
-  - rm -rf cmocka-1.0.1 cmocka-1.0.1.tar.xz
+  - rm -rf cmocka-1.1.5 cmocka-1.1.5.tar.xz
   
 script: 
   - autoreconf -i && ./configure && make
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -108,7 +108,7 @@ void calledit(void) {
 		hiscall[j] = hiscall[j + 1];	/* move to left incl. \0 */
 	    }
 
-	    showinfo(getctydata_pfx(hiscall));
+	    show_call_info(hiscall);
 
 	    if (cnt > 1)
 		searchlog();
@@ -126,7 +126,7 @@ void calledit(void) {
 		    hiscall[j] = hiscall[j + 1];
 		}
 
-		showinfo(getctydata_pfx(hiscall));
+		show_call_info(hiscall);
 
 		if (cnt > 1)
 		    searchlog();
@@ -173,7 +173,7 @@ void calledit(void) {
 		else
 		    break;
 
-		showinfo(getctydata_pfx(hiscall));
+		show_call_info(hiscall);
 
 		searchlog();
 
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -26,6 +26,7 @@
 
 #include <string.h>
 
+#include "callinput.h"
 #include "getctydata.h"
 #include "globalvars.h"
 #include "keystroke_names.h"
@@ -108,7 +109,7 @@ void calledit(void) {
 		hiscall[j] = hiscall[j + 1];	/* move to left incl. \0 */
 	    }
 
-	    show_call_info(hiscall);
+	    update_info_line();
 
 	    if (cnt > 1)
 		searchlog();
@@ -126,7 +127,7 @@ void calledit(void) {
 		    hiscall[j] = hiscall[j + 1];
 		}
 
-		show_call_info(hiscall);
+		update_info_line();
 
 		if (cnt > 1)
 		    searchlog();
@@ -143,11 +144,9 @@ void calledit(void) {
 	} else if (i != ESCAPE) {
 
 	    // Promote lower case to upper case.
-	    if ((i >= 97) && (i <= 122))
-		i = i - 32;
+	    i = g_ascii_toupper(i);
 
-	    // Accept A-Z or / and 1-9
-	    if (((i >= 65) && (i <= 90)) || ((i >= 47) && (i <= 57))) {
+	    if (valid_call_char(i)) {
 
 		call2[0] = '\0';
 
@@ -173,7 +172,7 @@ void calledit(void) {
 		else
 		    break;
 
-		show_call_info(hiscall);
+		update_info_line();
 
 		searchlog();
 
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -150,13 +150,12 @@ int callinput(void) {
 
     attron(modify_attr(COLOR_PAIR(NORMCOLOR)));
 
-    printcall();	/* print call input field */
-    searchlog();
-
     while (strlen(hiscall) <= 13) {
 
 	show_zones(bandinx);
-	printcall();
+	update_info_line();
+	searchlog();
+	printcall();    // note: calls refreshp()
 
 	/* wait for next char pressed, but update time, cluster and TRX qrg */
 	/* main loop waiting for input */
@@ -184,8 +183,8 @@ int callinput(void) {
 		    if (fabs(freq - freqstore) > 500) {
 			add_to_spots(hiscall, freqstore);
 			hiscall[0] = '\0';
-			HideSearchPanel();
 			freqstore = 0;
+			break;
 		    }
 		}
 	    }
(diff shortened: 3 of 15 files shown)


dl1jbe:
As `clear_display()` does refresh call and comment maybe it would be a better place to also re-display DX info. Downside: We need to check all places we use `clear_display()' if we have some relevant data in the info line we do not like to be overwritten.

zcsahok:
As there are no defined specs how the bottom info line shall behave how about this

- if the entered call (`hiscall`) has at least 2 chars and country information can be determined then show this info
- else if there are recent WWV info available the show it
- else info line is filled with blanks (cleared).

With this it would be enough to move calling of `showinfo()` to from `displayit()` [what a misleading name] to `clear_screen()` and make (hopefully) small fixes to showinfo and/or getctydata_pfx.
BTW showinfo is always used as `showinfo(getctydata_pfx(hiscall))`.

dl1jbe:
> As there are no defined specs how the bottom info line shall behave how about this
> 
>     * if the entered call (`hiscall`) has at least 2 chars and country information can be determined then show this info
> 
>     * else if there are recent WWV info available the show it
> 
>     * else info line is filled with blanks (cleared).
> 
Agreed. And to make it explicit let us add these specs to the code.

> With this it would be enough to move calling of `showinfo()` to from `displayit()` [what a misleading name] to `clear_screen()` and make (hopefully) small fixes to showinfo and/or getctydata_pfx.

Yep. Lets go in that direction. (I agree about the naming of `displayit()` - let us look for a better one.)

> BTW showinfo is always used as `showinfo(getctydata_pfx(hiscall))`.

Not quite - there is `showinfo(SHOWINFO_DUMMY)` to clear the line. But we can move these case to a separate function (e.g. clear_infoline()`).

zcsahok:
The display of WWV info via `time_update()` will also be affected. Currently WWV info has precedence over country info. This will be the other way around. I see no problem with this as WWV will anyway shown once country info is cleared at the end of a QSO.
WWV info is and will be not actively removed if the info line is idle: last info is shown unless country info overwrites it.

There is a minor bug that will be more visible: the DX time part of country info is static and off by +/-DST offset. (Currently at 21h local time in HA it shows DX time 20h for DL, although we are in the time zone)
There are many options:

1. live with it; take it for a rough estimate (+/- 1 hr)
2. don't show DX time
3. update DX time in `time_update()` and apply the correct offset.

I'm fine with 1) or 2). The correct solution is 3), but getting the right offset can be non-trivial as DST zones (and rules) do not necessarily match DX zones.

N0NB:
I suspect that to do option 3 that someone will end up living deep in the system tzdata, or it might not be that much of a problem.  What I think amateur radio is lacking is a cross reference of DXCC prefixes with ISO country designators.  Developing that probably would be a tiresome and thankless job!

I think the best is the use the data in `cty.dat` which, if I recall, gives coordinates and time offsets from UTC for each nation's capital (I think he has the time offsets reversed as Washington, D.C. is actually UTC -5 hours whereas he shows it as 5 hours).  Scratch that on the coordinates I as checked the ones for the US and that point is in south central Missouri.  Maybe it's the population center.  I will note that the longitude values appear to be reversed as well.  Sigh.

dl1jbe:
I vote for option one. It corresponds to the real daylight situation.
Anyway it should be documented somewhere.

zcsahok:
This time zone issue is kind of can of worms.
This map shows that time zone (pink) and CQ zone (blue) boundaries don't coincide.
image
Time zone even don't follow state boundaries, just check Tennesse for example.
Worse, Tennessee could be even part of two CQ zones (it wasn't quite clear from the map).
Now, even `cty.dat` has provisions to provide time zone overrides (enclosed with tildes), it does not contain them. So we have to live with the fixed 5 hr offset for US.
`cty.dat` even doesn't distinguish US call areas, where 1 would belong to CQ zone 5 and 6 to zone 3. This could be quite annoying for CQ WW (I'm checking it in #275).
I had a quick look into other formats provided by AD1C and XML seems promising as it contains more fine granular information. Anyway, a feature for version 2.0 ....

For the moment we then stay with option 1).

N0NB:
> This time zone issue is kind of can of worms.

Without question.

> This map shows that time zone (pink) and CQ zone (blue) boundaries don't coincide.
> image
> Time zone even don't follow state boundaries, just check Tennesse for example.

There are several counties in western Kansas in the Mountain time zone as well as all of the Nebraska pan handle.  Half of the Florida pan handle is in the central time zone and so on.  In short, it's really a mess and not easily definable without a lot of exceptions.

If the time zone boundaries came straight through, Mountain time would start about 30 miles west of me.

> Worse, Tennessee could be even part of two CQ zones (it wasn't quite clear from the map).

I've never seen a CQ map that was more precise than that one.  I think that those who live close to a boundary either interpolate or simply select a zone that looks good to them and move on.  If there is a list of states and counties within states that belong in one CQ zone or another, I've not seen it.

> Now, even `cty.dat` has provisions to provide time zone overrides (enclosed with tildes), it does not contain them. So we have to live with the fixed 5 hr offset for US.

Still, for our purposes it usually is not a critical issue.  Those interested in gray line propagation have other tools to offer precise times when another entity is in the gray line.  Even then it is relative depending on the time of year.

> `cty.dat` even doesn't distinguish US call areas, where 1 would belong to CQ zone 5 and 6 to zone 3. This could be quite annoying for CQ WW (I'm checking it in #275).

US call areas are pretty much meaningless as a definition of where a station is located as we are no longer required to change our calls when moving our station location to another call area.  The vanity call system does not enforce call areas either, at least for the lower 48 where an available call from anywhere may be requested (DXCC entities administered by the FCC excepted, i.e. a mainland station cannot request a call with a prefix reserved for Guam).  `cty.dat` would easily grow by five times in size just for the US exceptions if that were attempted.  Even now the US zone exceptions is ridiculously large.

> I had a quick look into other formats provided by AD1C and XML seems promising as it contains more fine granular information. Anyway, a feature for version 2.0 ....

I was unaware he was offering an XML format.  XML is so '00s.  The cool cats use JSON these days.

> For the moment we then stay with option 1).

Agreed.  Don't put too much effort into this that could be better applied elsewhere.

IMO, more important is showing the correct short path azimuth to the DX station.  In country azimuth calculation from a simple file like `cty.dat` is hopeless, but that is not its intent.  Even within the NA continent `cty.dat` is useless for azimuth calculations to Canada.  To Mexico it is marginally helpful.  To the Caribbean more so.

When a grid square is known, calculations can be more precise, but grid squares are not popular for HF work.

N0NB:
Just a little more fuel for this fire.

hsustimezones

This picture shows the US time zones by county.  Notice that IN and AZ are outliers as neither participate fully in DST although I think IN mostly does now.  Then look at Idaho, which is really a mess!

Judging by the CQ zone image above, it looks like the boundary of Zones 4 and 5 divide TN and FL where it follows the time zone line.  The boundary between Zones 3 and 4 is even less clear to me.

N0NB:
Perhaps not relevant, but this Reddit thread is someone else diving into this sort of thing:  https://www.reddit.com/r/hamdevs/comments/lwy2b1/dxcc_entity_list_in_json_and_csv_ive_added/

zcsahok:
Please check if this solution is OK.

dl1jbe:
First tests are looking good. At the moment I see some minor points for discussion:

- `show_call_info()` is only called with `hiscall` as parameters. I am not sure if we need it here or should hard code it. 
- That would also hide the weird usage when `hiscall` is cleared and `show_call_info()` is used to empty the line.
- The name of the function seems also a bit misleading. Sometimes it display the call info, sometimes the WWV info and sometimes it clears the line. How about something like `update_info_line()` instead?

Thanks also for the change to cmocka-1.1.5. I am using it here for quite some times.

At the moment I am trying to understand the changes to dxcc.c and getctydata.c. Is there a sitution where getctydata is called with a NULL pointer instead of a call string?

zcsahok:
Changed info line function names and dropped `hiscall` param. `update_info_line()` has a similar working as `searchlog()`: shows result if possible else clears field (ok, except WWV).

Moved the historically scattered printcall/searchlog/update_info_line blocks to the beginning of the mega loop in `callinput()'. This function shall in fact "simply" look like
```
while() {
   update_screen_according_to_current_call(); // = searchlog+update_info+printcall
   key = wait_for_key(); // internally handle grabbing, etc.
   key = process(key);
   if (key == ENTER || ...) break;
}
```
Now complexity is reduced by 1%...
Added `valid_call_char()` to be shared with `calledit`. That function is also worth a review later. As a bonus I found an unused function in there: `insert_char()`.

`getctydata` + NULL: an early idea was to use it for the dummy case. But then is turned out that no explicit dummy handling is needed. All is done correctly implicitly. Anyway, the check for NULL doesn't hurt.

`dxcc`: all data for prefixes is filled in to avoid later special handling. Only potential drawback is that continent must be allocated for all entries. But if allocation unit is small (say 4 bytes) then it is still smaller than a 64-bit int.

dl1jbe:
New function and restructured code looks good. +1.

Just wondering if we can drop the separate `show_wwv_info_line()`. As showing the call info has precedence over wwv anyway why not just call `update_info_line`.  If a call is present its info will be refreshed, if not the wwv info is shown.

Otherwise feel free to merge it in.

And thanks for NULL ptr and dxcc clarification

zcsahok:
For some reason I thought that WWV info is displayed from the backround task. Now I realised that it is in fact called from `time_update()` that is in turn called from the main task via various functions. So no mutex is needed and there is also no issue with `getctydata` setting global variables.

We shall not forget to look into the background task as it does call functions that modify global variables (fldigi handling, net handling) potentially without mutexes. As these happen not too frequently we experience no major issues, but nevertheless it should be checked and fixed, if needed.

dl1jbe:
> We shall not forget to look into the background task as it does call functions that modify global variables (fldigi handling, net handling) potentially without mutexes. As these happen not too frequently we experience no major issues, but nevertheless it should be checked and fixed, if needed.

I did fix the most frequently in last years, but you are right - there are quite some more.

dl1jbe:
Did a little rearrangement and squashing of the commits and merged them in. 

These also closes issue #266

PR #285, opened 2021-10-29
Use Github Actions for CI build

zcsahok:
Some time after closing down travis.org I migrated my account to travis.com. Builds continued to work as before. Travis' free plan grants 1000 minutes and once they're used up there's no automatic extension, one has to request it via email.
Recently Tlf builds error out claiming that cmocka.org's certificate has expired which it in fact did not.
So, instead of debugging what's wrong at Travis I looked into Github Actions and figured out that the migration from Travis is quite straightforward. The default Linux instance uses Ubuntu 20, so it has the right cmocka anyway.

Details about billing for Actions can be found here
https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

Unlike at Travis it's clearly stated that 2000 minutes are granted per month.
As one Tlf build takes ~1 minute it'll take a long way to run into this limit.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,0 +1,31 @@
+name: CI build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    env:
+      TERM: xterm
+      AM_COLOR_TESTS: always
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Show OS info
+      run: cat /etc/os-release
+    - name: Install dependencies
+      run: |
+        sudo apt-get -qq update
+        sudo apt-get install -y libhamlib-dev libxmlrpc-core-c3-dev libcmocka-dev     
+    - name: Autoreconf and basic make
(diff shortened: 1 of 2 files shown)


dl1jbe:
Thanks for looking into the problem and providing a solution.

I had a quick look into github actions and it seems that it is the way to go. 
I aggree that 1000 mins should be enough for CI tests. Do you know if we somehow can see how much of that is already used during the month?

zcsahok:
Current billing info is on this page https://github.com/settings/billing
For some reason for me it still shows 0 used minutes even though I ran some builds. Maybe public repos counted differently?

image

image

dl1jbe:
It is the same here. Seems it gets only displayed for private repos. 

It would be good to get a feedback for the time consumed from that free limit. But as long as we use the gh actions only for CI it should be more than enough.

zcsahok:
Just found it:
_GitHub Actions usage is **free for** both **public repositories** and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account._ 

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions

dl1jbe:
Good to hear. Thanks for looking it up.

PR #289, opened 2021-11-17
Convert mvprintw to mvaddstr or clear_line

zcsahok:
Related to #284: following transformations were done to get away from mvprintw

- `mvprintw(y, x, "%s", "const string")` and `mvprintw(y, x, "const string")` into `mvaddstr(y, x, "const string")`
- `mvprintw(y, x, backgrnd_str)` into `clear_line(y)`

Test had to be fixed as `mvaddstr` is (unlike `mvprintw`) a macro not a function.

On the occasion the code duplication in `rtty.c` was reduced. (it's not an exact duplicate, probably due to a bug).

Will also check `mvwprintw` and `wprintw`.

### DIFF
--- a/src/addspot.c
+++ b/src/addspot.c
@@ -74,7 +74,7 @@ void addspot(void) {
 
 	attron(COLOR_PAIR(C_LOG) | A_STANDOUT);
 
-	mvprintw(13, 20, "freq.: ");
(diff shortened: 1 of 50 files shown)


zcsahok:
Pushed mvwprintw part.

dl1jbe:
Applied and pushed.

PR #290, opened 2021-11-23
Convert to bool

dl1jbe:
Convert most variables which controls scoring to bool data type.

Drop not used variables


src/parse_logcfg.c
```diff
@@ -1036,47 +1036,47 @@ static config_t logcfg_configs[] = {
     {"USEPARTIALS",     CFG_INT_ONE(use_part)},
```
zcsahok, line 1036: Definitely improves readability in my opinion.
Practically every `CFG_INT_ONE` config is a bool.

dl1jbe: True, there are quite some more changes to do. Will do it in next days.

zcsahok (approved):
all OK

PR #292, opened 2021-11-24
Rescore upon reading the log

zcsahok:
Too simple to be true. :-)  But let's see.

Here is an example: in the OK-OM contest I didn't set ONE_POINT, so all QSOs were scored with zero points. Mults were configured correctly, so that part is fine.
image

Added the missing ONE_POINT to the rule file and restarted Tlf: now I've 39 points for the 39 QSOs.
image

All QSOs were on 80, so as a test I moved the one with section MYJ to 40. I worked MYJ on 80 before, so the net result is that I "earn" it as a new one on 40. This operation was done from within Tlf using :EDIt, no restart was required.
image

As discussed in #248 the log itself is not updated (currently neither on disk nor in memory), hence the lines still show 0 points in the screenshots above. This could be seen as bug, but the fix must be discussed.


dl1jbe:
> Too simple to be true. :-) But let's see.
> 
Yes, that step was obvious. If you compare old code in readcalls() and the code in addcall() carefully you will find some subtle differences in handling of corner cases like arrl contests for the w/ve side. I was not quite sure which behavior is the right one, so I hesitated from the simple change before understanding all implications. But lets go on.

> As discussed in #248 the log itself is not updated (currently neither on disk nor in memory), hence the lines still show 0 points in the screenshots above. This could be seen as bug, but the fix must be discussed.

I think it can be made simple (but not without some changes). Let us keep the parsed records ('struct qso_t') and add them to a ptr array. That array would be a complete internal representation of all qsos and comments. 

Than we can reconstruct the display line from the record when needed for display. That way it will always be up to date. 

As an additional side effect it enables us later to generate the display line in different width (and detail) as soon as we support a variable display width in xterm

Both ideas were the main reason to introduce the 'struct qso'. To keep it simple I just started in spring with only the struct and left out the collection of all structs in the array for later.


As a sidenote, let me ask to keep the commits a little more clean. Having the unrelated changes for station info or scoring in the same commit makes it harder in the review to focus on the real core of the commit.

zcsahok:
re station info: it had to be moved before reading the log (country has to be known) and since it used plain printw's it got partially obscured by readcalls' output (via showmsg with its internal line number tracking), hence the rework.

dl1jbe:
> re station info: it had to be moved before reading the log (country has to be known) and since it used plain printw's it got partially obscured by readcalls' output (via showmsg with its internal line number tracking), hence the rework.

Oh, thanks. I missed that point.

src/log_to_disk.c
```diff
@@ -61,12 +61,6 @@ void restart_band_timer(void) {
     }
 }
 
-/* score QSO and add to total points */
-void score_qso(void) {
```
zcsahok, line 65: This has been moved to `score.c` as this single function required a bunch of dummies to be implemented in the test. It anyway fits better scoring than writing to disk.

src/getexchange.c
```diff
@@ -419,7 +419,7 @@ int getexchange(void) {
 		       && ((x != TAB) && (strlen(section) < 1))) {
 		if (!serial_or_section
 			|| (serial_or_section && country_found(hiscall))) {
-		    mvaddstr(13, 54, "section?X");
```
zcsahok, line 422: A leftover from testing my previous PR...

src/addmult.c
```diff
@@ -57,7 +57,7 @@ void addmult(struct qso_t *qso) {
 
 	idx = get_exact_mult_index(mult1_value);
 	if (idx >= 0) {
-	    remember_multi(get_mult(idx), bandinx, ALL_BAND);
```
zcsahok, line 60: It was using the global variable. (which we shall drop later, I guess)

src/readcalls.c
```diff
@@ -161,19 +164,14 @@ void count_contest_bands(int check, int *count) {
...
-    bool veto;
     struct qso_t *qso;
     int linenr = 0;
 
-    int bandindex = BANDINDEX_OOB;
-
     FILE *fp;
 
-    showmsg("Reading logfile... ");
+    char *info = g_strdup_printf("Reading logfile: %s\n", logfile);
```
zcsahok, line 172: Showing log file name as other "Reading..." prompts do.

src/readcalls.c
```diff
@@ -185,7 +183,7 @@ int readcalls(const char *logfile) {
 	exit(1);
     }
 
-    while (fgets(inputbuffer, LOGLINELEN + 1, fp) != NULL) {
+    while (fgets(inputbuffer, sizeof(inputbuffer), fp) != NULL) {
```
zcsahok, line 186: Just cosmetics (and trying to have a single point of definition). Could also change the next line too.

src/readcalls.c
```diff
@@ -197,206 +195,30 @@ int readcalls(const char *logfile) {
...
-		pfxnummulti[pfxnumcntidx].qsos[pxnr] |= inxes[bandindex];
-	    }
-	}
 	free_qso(qso);
 	qso = NULL;
     }
 
     fclose(fp);
 
-    /* all lines red, now build other statistics */
```
zcsahok, line 314: All this must also happen during `addcall()` & co. Will add some tests to check it.

src/readcalls.c
```diff
@@ -197,206 +195,30 @@ int readcalls(const char *logfile) {
...
-    }
-
     return linenr;			// nr of lines in log
 }
 
 int log_read_n_score() {
-    int nr_qsolines;
+    int nr_qsolines = readcalls(logfile);
 
-    total = 0;
```
zcsahok, line 398: `total` is reset in `readcalls()`

src/readcalls.c
```diff
@@ -197,206 +195,30 @@ int readcalls(const char *logfile) {
...
-    nr_qsolines = readcalls(logfile);
     if (qtcdirection > 0) {
 	readqtccalls();
```
zcsahok, line 223: On startup `readqtccalls()` is called before `readcalls()`. Hopefully the order is not relevant.

test/test_readcalls.c
```diff
@@ -18,22 +19,48 @@
...
 // OBJECT ../src/zone_nr.o
 
-char section[8] = "";       // defined in getexchange.c
+bool lan_active = false;
 
 // dummy functions
 void readqtccalls() {}
 void shownr(char *msg, int x) {}
-void spaces(int n) {} /* needs more care */
+const char *spaces(int n) { return "";} /* needs more care */
```
zcsahok, line 39: Dummies for `getexchange`...

zcsahok:
Will add (probably next week) updating the in-memory log using `makelogline`.
If there were changes the user will be asked whether to update the log on disk.

zcsahok:
Fixed (or at least improved) WPX rescoring. For this the `add_pfx` call hidden in `makelogline` had to be moved to `addcall` where it logically belongs. The somewhat obscure `excl_add_veto` flag thus could be made local. The condition for WPX has been changed to a hopefully more readable form (if not excluded || not pfxab). I'm not sure if this is actually correct, though.
For the sake of `addcall2` tests `excl_add_veto` has been kept as `excl_add_veto2` (will be removed later together with `addcall2` as soon we get the the networking part...). Plain `addcall` tests now check `countryscore` instead.

src/readcalls.c
```diff
@@ -84,105 +80,54 @@ void init_scoring(void) {
 
     InitPfx();
 
-    if (pfxnummultinr > 0) {
```
zcsahok, line 84: Dropped unnecessary check

src/readcalls.c
```diff
@@ -84,105 +80,54 @@ void init_scoring(void) {
...
-    refreshp();
+    if (interactive) {
+	showstring("Reading logfile:", logfile);
+    }
 
     init_scoring();
 
     if ((fp = fopen(logfile, "r")) == NULL) {
 	showmsg("Error opening logfile ");
-	refreshp();
```
zcsahok, line 183: Already done in showmsg()

test/test_readcalls.c
```diff
@@ -76,7 +84,7 @@ int pacc_pa(void) {
     return 0;
 }
 
-#define QSO1 " 40SSB 12-Jan-18 16:34 0006  PY9BBB         599  599  15                    10         \n"
+#define QSO1 " 80SSB 12-Jan-18 16:34 0006  PY9BBB         59   59   15            PY   15  3  14025.0\n"
```
zcsahok, line 87: Use a valid log line. Moved to 80 as 40 is the default band and this may hide errors.

zcsahok:
Log lines are re-created using `makelogline()` and checked against the ones in the file. If a change is detected then once all lines are loaded user is prompted if they want to save (update) the log. Previous log is backed up as <timestamp>_<logfilename>.

Next `makelogline()` has to be "de-globalized". This will also involve replacing `linedata_t` with `qso_t`.

src/readcalls.c
dl1jbe: Not quite sure. That needs to be checked.

src/readcalls.c
dl1jbe: addcall() should be fine in building the statistics. The statistics code at the end of readcalls() were written so as to mimic the logic represented by addcall().

dl1jbe:
> The somewhat obscure `excl_add_veto` flag thus could be made local. 

I did some checks for that variable and tried to make some comments about its working - see  https://github.com/dl1jbe/tlf/commit/842feaf84c496b20841f8149328d718b3c2a5146


The condition for WPX has been changed to a hopefully more readable form (if not excluded || not pfxab). I'm not sure if this is actually correct, though. For the sake of `addcall2` tests `excl_add_veto` has been kept as `excl_add_veto2` (will be removed later together with `addcall2` as soon we get to the networking part...). Plain `addcall` tests now check `countryscore` instead.

dl1jbe:
> The condition for WPX has been changed to a hopefully more readable form (if not excluded || not pfxab). I'm not sure if this is actually correct, though. 

I will try to check the handling in the existing code and compare with older versions. Maybe we can sort it out.

src/readcalls.c
```diff
@@ -84,105 +80,54 @@ void init_scoring(void) {
...
 	sleep(2);
 	exit(1);
     }
 
+    bool log_changed = false;
+    // array of qso's
+    // FIXME use this instead of qsos[] (make it global and don't free it here)
+    // FIXME also include comments/notes into qso_t
```
dl1jbe, line 128: We could just add a 'iscomment' flag and a pointer to the comment string or save some memory by using a union union from the actual qso_t and the comment field pointer.

src/readcalls.c
```diff
@@ -191,8 +136,11 @@ int readcalls(const char *logfile) {
 	// remember logline in qsos[] field
 	g_strlcpy(qsos[linenr], inputbuffer, sizeof(qsos[0]));
 	linenr++;
+	// FIXME check for overflow....
```
dl1jbe, line 139: Above two lines can be dropped as soon as we use the qso_array to keep all information.

zcsahok: True.

zcsahok:
Ideally scoring and mult handling shall be independent of whether the QSO originated locally, received from network or read from a file. Therefore any veto logic have to be an internal detail of the processing.

dl1jbe:
We are on the way to it. 
I think the best point is that the rules are now conentrated on a single place. In old code we had two places to  keep in sync.

src/makelogline.c
```diff
@@ -211,120 +209,120 @@ void prepare_fixed_part(void) {
...
 
     } else if (CONTEST_IS(CQWW) || wazmult || itumult) {
 	//-------------------------cqww----------------
-	if (trxmode == DIGIMODE && CONTEST_IS(CQWW) && strlen(comment) < 5) {
-	    comment[2] = ' ';
-	    comment[3] = 'D';
-	    comment[4] = 'X';
-	    comment[5] = '\0';
+	char buf[80];
+	strcpy(buf, qso->comment);
```
zcsahok, line 256: Change a private copy of comment.

src/makelogline.c
```diff
@@ -106,69 +115,58 @@ void makelogline(void) {
...
-	strcat(logline4, "SSB");
+    if (qso->mode == CWMODE)
+	strcat(logline, "CW ");
+    else if (qso->mode == SSBMODE)
+	strcat(logline, "SSB");
     else
-	strcat(logline4, "DIG");
+	strcat(logline, "DIG");
 
-    if (do_cabrillo == 0) {
```
zcsahok, line 114: Unified handling, dropped Cabrillo specific variables.

src/readcabrillo.c
```diff
@@ -421,15 +422,14 @@ int readcabrillo(int mode) {
     if (!cabdesc) {
 	show_readcab_msg(mode, "Cabrillo format specification not found!");
 	sleep(2);
-	do_cabrillo = 0;
-	return (2);
-    } else {
```
zcsahok, line 426: Straightened else branch.

src/store_qso.c
```diff
@@ -40,13 +40,10 @@ int store_qso(char *loglineptr) {
     }
     strcpy(qsos[nr_qsos], loglineptr);
     nr_qsos++;
-    strcat(loglineptr, "\n");	/* pa3fwm, 20040113: this looks suspicious,
```
zcsahok, line 43: Honored comment from 2004...

zcsahok:
Let me know if you see something open here. If there are none then I'd merge this and also close #248.

src/readcalls.c
```diff
@@ -81,322 +80,128 @@ void init_scoring(void) {
...
-	    if (add_ok == false) {
-		qsos_per_band[bandindex]++;
-	    }
-
-	    hiscall[0] = '\0';
+	dupe = is_dupe(qso->call, qso->bandindex, qso->mode);
+	addcall(qso);
+	score_qso();    //FIXME argument?
+
+	char *logline = makelogline(qso);
```
dl1jbe, line 160: After finishing all migration away from qsos[], we do not need to record such things as points, multis or prefixes in the written log. 
The we could drop these check for a change of logfile. The file would be the same as long as we do not change some of the qso information (time, call, exchange, band...).
But that leads us again to a discussion of logfile format. We should keep that for later.

src/readcalls.c
```diff
@@ -81,322 +80,128 @@ void init_scoring(void) {
...
+    if (log_changed && interactive) {
+	showmsg("Log changed due to rescoring. Do you want to save it? Y/(N)");
+	if (toupper(key_get()) == 'Y') {
+	    // save a backup
+	    char prefix[40];
+	    format_time(prefix, sizeof(prefix), "%Y%m%d_%H%M%S");
+	    char *backup = g_strdup_printf("%s_%s", prefix, logfile);
+	    rename(logfile, backup);
+	    // rewrite log
+	    remove(logfile);
```
dl1jbe, line 186: There should be no 'logfile' after the rename.

test/test_cabrillo.c
```diff
@@ -70,10 +66,16 @@ int getctynr(char *call) {
 }
 
 /* some spies */
-int bandinx_spy;
-
-void makelogline() {
-    bandinx_spy = bandinx;
+struct qso_t *qso_spy = NULL;
+
+char *makelogline(struct qso_t *qso) {
```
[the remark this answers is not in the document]
dl1jbe: I see

dl1jbe (approved):
Besides the question about 'freq' if no rig control is active I think all should be well.

Please double check and feel free to merge it in afterwards.

src/addcall.c
```diff
@@ -207,13 +207,14 @@ int addcall(struct qso_t *qso) {
...
 
-    addmult(current_qso);       /* for wysiwyg */
+    addmult(qso);           /* for wysiwyg */
 
     return cty;
 }
 
 /* ----------------------for network qso's-----------------------------------*/
 
+int excl_add_veto2;
```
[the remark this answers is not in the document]
zcsahok: Yes, this is part of network code. Just didn't want to touch it now. It'll be a can of worms for sure.

src/readcalls.c
dl1jbe: True

src/makelogline.c
dl1jbe: Ack

src/addcall.c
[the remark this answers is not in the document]
dl1jbe: ack

src/makelogline.c
```diff
@@ -53,32 +65,36 @@ void fillto(int n);
...
     }
 
-    fillto(80);
+    FILL_TO(80);
 
     /* add freq to end of logline */
-    if (trx_control) {
-	snprintf(freq_buff, 8, "%7.1f", freq / 1000.0);
-	strcat(logline4, freq_buff);
+    if (qso->freq > 0) {
```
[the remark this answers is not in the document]
zcsahok: `trx_control` tells whether rig control is currently in use, whereas `qso->freq > 0` checks if freq was known at the time of the QSO. Generally this could have been earlier or even currently but on another Tlf node. So relying on `trx_control` is not correct.

If trx control is disabled while Tlf is running then `freq` is not set to zero (previous value is kept). Similarly if trx control is enabled then rig connection is not initialized, freq stays zero. IMO both are bugs.

src/makelogline.c
```diff
@@ -99,69 +115,58 @@ void makelogline(void) {
...
-	    sent_rst[2] = ' ';
-	    recvd_rst[2] = ' ';
-	}
-
-	strcat(logline4, sent_rst);	/* till 54 */
-	strcat(logline4, "  ");
-	strcat(logline4, recvd_rst);
-	strcat(logline4, "  ");
+	sprintf(buf, "%-3d  %-3d  ",
+		qso->rst_s ? qso->rst_s : default_rst(qso),
```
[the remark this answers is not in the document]
zcsahok: ok, renamed to `standard_rst`

src/readcalls.c
zcsahok: fixed

dl1jbe:
Am Tue, 04 Jan 2022 07:01:50 -0800
schrieb Zoltan Csahok ***@***.***>:

> @zcsahok commented on this pull request.
> >    
>      /* add freq to end of logline */
> -    if (trx_control) {
> 
> `trx_control` tells whether rig control is currently in use, whereas
> `qso->freq > 0` checks if freq was known at the time of the QSO.
> Generally this could have been earlier or even currently but on
> another Tlf node. So relying on `trx_control` is not correct.
>
Ok, I overlooked the situation wrt other tlf nodes which may report the
frequency-
 
> If trx control is disabled while Tlf is running then `freq` is not
> set to zero (previous value is kept). Similarly if trx control is
> enabled then rig connection is not initialized, freq stays zero. IMO
> both are bugs.
> 
Right. Enabling trx_control should try to initialize the rig
connection. Needs to be done. Both bugs should be fixed. Will put it on
todo list.



-- 
"Do what is needful!"
Ursula LeGuin: Earthsea
--

src/readcalls.c
dl1jbe: See issue #310

PR #295, opened 2021-11-26
Make interrupting auto_cq more responsive

df7cb:
Instead of sleeping 500ms, sleep 10 times 50ms, checking for input in between.

### DIFF
--- a/src/autocq.c
+++ b/src/autocq.c
@@ -51,12 +51,48 @@ static int get_autocq_time() {
     return (int)(1200.0 / GetCWSpeed()) * cw_message_len;
 }
 
-
-int auto_cq(void) {
-
 #define NO_KEY -1
 
+static int wait_50ms_for_key() {
+
+    usleep(50 * 1000);
+
+    const int inchar = key_poll();
+    if (inchar > 0 && inchar != KEY_RESIZE) {
+	return inchar;
+    }
+
+    return NO_KEY;
+}
+
+#define TIME_UPDATE_MS  500
+
+// Wait for given ms, polling each 50 ms for a key and calling time_update()
+// each 500 ms. Pressing a key terminates the waiting loop.
+// Note: this works best if ms >= 500 (or ms = 0)
+static int wait_ms(int ms) {
     int key = NO_KEY;
+    int update_timer = TIME_UPDATE_MS;
+    int wait_timer = ms;
+
+    while (wait_timer > 0 && key == NO_KEY) {
+
+	key = wait_50ms_for_key();
+
+	wait_timer -= 50;
+	update_timer -= 50;
+
+	if (update_timer <= 0) {
+	    time_update();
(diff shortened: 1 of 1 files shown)


src/autocq.c
```diff
@@ -91,13 +91,17 @@ int auto_cq(void) {
 	    mvprintw(12, 29, "Auto CQ  %-2d ", delayval - 1);
 	    refreshp();
 
-	    usleep(500000); // 500 ms
-	    time_update();
+	    // sleep 10 * 50 = 500ms unless a key is pressed
```
zcsahok, line 94: Let's handle waiting while sending CQ (line 78) and waiting after CQ uniformly.
I mean

- wait for a total of `message_time` + `delayval`*(1/2 sec)
- in the loop poll for a key each 50 ms
- and call `time_update()` each 500 ms

dl1jbe: Needs additional logic to handle the countdown display during 'delayval' time.

Anyway it seems sensible to use the same 50ms increment in each case.

zcsahok:
Pushed an update: 
- first it waits for the message to finish, then starts the usual 1/2 sec countdown
- each 50 ms check for a key and `time_update` is called each 500 ms

Message time is considered with 50 ms granularity, but that's fine as 50 ms is just a single dot time at 24 wpm.

There is a minor change in the countdown.
Original code had this for CQDelay=4: 3, delay, 2, delay, 1, delay, 0, delay,  start cq again
I find more logical this sequence: 4, delay, 3, delay, 2, delay, 1, delay, start cq again

dl1jbe:
Sorry for being late, Zoli. Not much time in last days.

The code works as expected. The move to a uniform 50ms interval is good.

My only concerns are maintenance aspects. The new code is more complex and more difficult to understand than the old one (see for example the 3 needed tests for message_wait in the inner while loop). Anyone who needs to touch the code in future will have to think about it for quite some time. 

So maybe I do not see the point. What do we gain with the new code?

zcsahok:
The goal was to reduce code duplication (two similar loops) and keep the "wait 50 ms + get key" part somewhat central.
Let me come up with another solution.

dl1jbe:
I see. Maybe some refactoring which hides some of the logic in sub-functions may make the code more clean.

zcsahok:
pse check https://github.com/zcsahok/tlf/tree/responsive-autocq2

dl1jbe:
+1 Nice and clean. I think that is much better. 

> There is a minor change in the countdown.
> Original code had this for CQDelay=4: 3, delay, 2, delay, 1, delay, 0, delay, start cq again
> I find more logical this sequence: 4, delay, 3, delay, 2, delay, 1, delay, start cq again

I agree.  Just go ahead and merge it in.

And Merry X-mas.

df7cb:
Thanks for polishing this up and merging it!

PR #296, opened 2021-11-26
Don't print a complaint if either Alt-PgUp/Dn or Ctrl-PgUp/Dn work

df7cb:
There is an annoying delay on startup if the terminal doesn't support
one of Alt-PgUp Alt-PgDn Ctrl-PgUp Ctrl-PgDn, even if the other
combination works. If fact this even happens with rxvt-unicode which the
manpage advertises as "does not eat Ctrl-PgUp".

Change logic such that the delay only happens if none of the keys work.

### DIFF
--- a/src/ui_utils.c
+++ b/src/ui_utils.c
@@ -95,7 +95,8 @@ int lookup_key(char *capability) {
 
 /** lookup all needed special keys not defined by ncurses */
 void lookup_keys() {
-    int not_ok = 0;
+    bool alt_not_ok = 0;
+    bool ctrl_not_ok = 0;
 
     key_kNXT3 = lookup_key("kNXT3");
     key_kPRV3 = lookup_key("kPRV3");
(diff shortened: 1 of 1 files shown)


src/ui_utils.c
```diff
@@ -95,7 +95,8 @@ int lookup_key(char *capability) {
 
 /** lookup all needed special keys not defined by ncurses */
 void lookup_keys() {
-    int not_ok = 0;
+    int alt_not_ok = 0;
```
zcsahok, line 98: Pse use bool type.

df7cb: It's been like that before, but it's bool now. Thanks!

zcsahok: (we are trying gradually improve code quality as we touch it. One of common topics is to use bool type if it's really meant to be a bool, see e.g. #290. This also involves using true/false literals and converting checks like ` if (a == 1)` into plain `if (a)`.)

zcsahok:
I'd merge this with minor cosmetic updates, if that's OK with you.

dl1jbe:
Fine. Just go ahead.

dl1jbe:
Ah, maybe the comment was not for me. Sorry.

zcsahok:
No problem. :-)

df7cb:
Sorry for the laggish responses, I am around but didn't have time to follow up. Will do that shortly with the other PRs.

Please just go ahead with whatever additions you have.

PR #298, opened 2021-12-07
Drop newline

dl1jbe:
Cleanup trailing newlines on all CW messages.

Furthermore fix a bug in message listing if no \n is present.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -221,9 +221,9 @@ char message[25][80] = /**< Array of CW messages
 			* message[1]  (F2)  - insert pressed
  			*/
 {
-    "TEST %", "@ DE %", "@ [", "TU 73", " @", "%",
+    "TEST %", "@ DE %", "@ [", "TU 73", "@", "%",
     "@ SRI QSO B4 GL", "AGN",
-    " ?", " QRZ?", " PSE K", "TEST % %", "@ [", "TU %",
+    "?", "QRZ?", "PSE K", "TEST % %", "@ [", "TU %",
     "", "", "", "", "", "", "", "", "", "", ""
 };
 
--- a/src/main.c
+++ b/src/main.c
@@ -221,9 +221,9 @@ char message[25][80] = /**< Array of CW messages
 			* message[1]  (F2)  - insert pressed
  			*/
 {
-    "TEST %\n", "@ DE %\n", "@ [\n", "TU 73\n", " @\n", "%\n",
-    "@ SRI QSO B4 GL\n", "AGN\n",
-    " ?\n", " QRZ?\n", " PSE K\n", "TEST % %\n", "@ [\n", "TU %\n",
+    "TEST %", "@ DE %", "@ [", "TU 73", " @", "%",
+    "@ SRI QSO B4 GL", "AGN",
+    " ?", " QRZ?", " PSE K", "TEST % %", "@ [", "TU %",
     "", "", "", "", "", "", "", "", "", "", ""
 };
 
--- a/src/main.c
+++ b/src/main.c
@@ -234,7 +234,10 @@ char ph_message[14][80] = /**< Array of file names for voice keyer messages
 			   */
 { "", "", "", "", "", "", "", "", "", "", "", "", "", "" };
 
-char qtc_recv_msgs[12][80] = {"QTC?\n", "QRV\n", "R\n", "", "TIME?\n", "CALL?\n", "NR?\n", "AGN\n", "", "QSL ALL\n", "", ""}; // QTC receive windowS Fx messages
+char qtc_recv_msgs[12][80] = {
+    "QTC?", "QRV", "R", "", "TIME?", "CALL?",
+    "NR?", "AGN", "", "QSL ALL", "", ""}; // QTC receive windows Fx messages
+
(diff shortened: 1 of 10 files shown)


zcsahok:
Trailing NL can have a role in digital modes. Will have a look at it.
I'm also not sure if there is a way to specify extra NL via logcfg.

dl1jbe:
Right. Digimodes has to be sorted out.

Old behavior for 'digi_message':

- No digi_messages are preset in main.c
- digi_messages not explicitly set get a copy of the CW message with \n replaced by ' ' (rules.c:100)
- While parsing logcfg.dat the same happens (CFG_MESSAGE_DYNAMIC)
- 'messagechange' would drop a trailing Enter key
- 'sendbuf ' in contrast honors \n and converts it to a CR character

So \n would be honored during set, but we have no chance to insert it into the message string. 

Maybe we should allow the string '\n' (2 characters) in message text and extend 'sendbuf' to replace that string with CR.

dl1jbe:
> So \n would be honored during set, but we have no chance to insert it into the message string.

...honored while sending, but ....

zcsahok:
An option would be to apply some unquoting to the message values (or generally config values) to allow adding leading/trailing spaces and NL. I've checked `g_shell_unquote` but it doesn't fit for our purpose.
(see at line 245: https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gshell.c)

Maybe a simple replace would suffice unless we find some library solution.

zcsahok:
Maybe `g_strcompress` would do.

dl1jbe:
Looking further it seems we already have what is needed.  We just have to document it.

'ExpandMacro()' accepts a '|' character in digi messages text as symbol for 'CR' and replaces it before sending the string. It is just missing in the documentation (see 'Macro characters in the message' in tlf.1).

src/main.c
```diff
@@ -221,9 +221,9 @@ char message[25][80] = /**< Array of CW messages
 			* message[1]  (F2)  - insert pressed
  			*/
 {
-    "TEST %\n", "@ DE %\n", "@ [\n", "TU 73\n", " @\n", "%\n",
-    "@ SRI QSO B4 GL\n", "AGN\n",
-    " ?\n", " QRZ?\n", " PSE K\n", "TEST % %\n", "@ [\n", "TU %\n",
+    "TEST %", "@ DE %", "@ [", "TU 73", " @", "%",
+    "@ SRI QSO B4 GL", "AGN",
+    " ?", " QRZ?", " PSE K", "TEST % %", "@ [", "TU %",
```
zcsahok, line 226: Some messages here have leading space. In case the user would like to have such space (or trailing space) it should be possible to do this via config. I assume that config values will be chomp'ed, so an option would be to have a replacement character for space the same way as for NL. Underscore would a natural candidate.
Something like
```
F9 = _?
```

src/parse_logcfg.h
```diff
@@ -109,10 +109,6 @@ typedef struct {
         (cfg_arg_t){.char_pp=&var, .chomp=true, \
                     .string_type=DYNAMIC}
 
-#define CFG_MESSAGE(var, i)     NEED_PARAM, cfg_string, \
```
zcsahok, line 112: Maybe we could keep CFG_MESSAGE name instead the CHOMP. (as there is no need to distinguish between the two)

zcsahok:
With F1 defined as
```
F1=CQ CQ DE % % % |CQ CQ DX % % % K
```
Fldigi works as expected (on the right, RTTY mode), but the keying window (on the left) shows the second line only. Probably due to pipe getting replaced by CR.
image

src/parse_logcfg.h
dl1jbe: Yes, good idea.

dl1jbe:
Will look into it.

src/parse_logcfg.h
dl1jbe: done. see commit below

src/main.c
zcsahok: After fiddling a bit with digi messages I think leading/trailing spaces are not really needed or can be effectively replaced by NL (|).
So how about

- drop leading spaces in CW messages too (incl. built-ins, here the op has more control on pacing the sending anyway)
- keep chomping messages in `parse_logcfg` and in `listmessages`
- add note in man page that unspecified digimode messages will have the value of their CW counterpart prepended with |. (`rules.c` line 87)

zcsahok:
Added "value too long" message to parse_logcfg, it's in 52d37c9a8bd185c90a43da90159c13f8fe7d8072

src/main.c
dl1jbe: Done.

dl1jbe:
> Added "value too long" message to parse_logcfg, it's in [52d37c9](https://github.com/Tlf/tlf/commit/52d37c9a8bd185c90a43da90159c13f8fe7d8072)

Did cherry pick and add it to the branch.

dl1jbe:
> Will look into it.

That needs a little bit more work as I would add some cleanup to 'sendbuf()' and others. I think it should go into an separate PR afterwards.

zcsahok:
OK. There's always something to fix here...

PR #299, opened 2021-12-23
Fix display of keyer messages containing embedded NL/CR characters

dl1jbe:
Furthermore clean up 'sendbuf()' and keyer terminal display

### DIFF
--- a/src/clear_display.c
+++ b/src/clear_display.c
@@ -159,22 +159,21 @@ void clear_display(void) {
 }
 
 /*
- *  scroll the loglines of the keyer terminal and show them
+ * add 'buffer' to the keyer terminal while scrolling one line up
+ * and refreshing the display
  */
-void displayit(void) {
-    extern char termbuf[];
+void add_to_keyer_terminal(char *buffer) {
 
-    char term2buf[81] = "";
+    char term2buf[81];
 
-    g_strlcpy(term2buf, termbuf, sizeof(term2buf));
+    g_strlcpy(term2buf, buffer, sizeof(term2buf));
     g_strchomp(term2buf);
-    g_strlcat(term2buf, backgrnd_str, sizeof(term2buf));	/* fill with blanks */
+    g_strlcat(term2buf, backgrnd_str, sizeof(term2buf));  /* fill with blanks */
 
     strcpy(terminal1, terminal2);
     strcpy(terminal2, terminal3);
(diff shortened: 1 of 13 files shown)


src/clear_display.c
```diff
@@ -48,20 +48,68 @@
...
+    mvaddstr(4, 0, terminal4);
+}
+
+/*
+ * add 'buffer' to the keyer terminal while scrolling one line up
+ * and refreshing the display
+ */
+static void add_line_to_keyer_terminal(char *buffer) {
+
+    if (terminal1)
```
zcsahok, line 87: just a minor note: g_free does the NULL check internally.

dl1jbe: Thanks for the reminder.

src/clear_display.c
```diff
@@ -48,20 +48,67 @@
...
+    clear_display();
+}
+
+/*
+ * add 'buffer' to keyer terminal, splitting it into separate lines if needed
+ */
+void add_to_keyer_terminal(char *buffer) {
+    gchar **lines = g_strsplit_set(buffer, "\n\r", 0);
+    for (int i = 0; lines[i] != NULL; i++)
+        add_line_to_keyer_terminal(lines[i]);
```
zcsahok, line 106: Using this setting
```
F1= CQ CQ DE % % %|CQ CQ DX % % % K
```
and calling CQ twice Tlf shows this
image

whereas fldigi shows
image

There seem to be a leading CR in the digi keyer strings which leads to an empty line on Tlf keyer terminal.
How about moving the chomp'ing into the for loop before calling `add_line_to_keyer_terminal` and skipping if line is empty? This of course means not displaying embedded CRs ("text1|||text2"), but we have just 4 lines anyway...

dl1jbe: Good point. I agree - the extra newline has to be dropped. Question is if we drop embedded newlines / carriage returns completely or cut them down to at most one newline.

Will look into it tomorrow. Today are other things to do :-) here.

dl1jbe: Actual solution is as recommended - just drop empty lines from display. They will still be sent via the digit keyer.

As far as I remember the discussion some years ago digi mode users are very seldom use more newlines than really needed. They try to use vertical space very sparingly - maybe leftover from paper printers :-).

zcsahok: Thanks, works OK.
HNY!

PR #300, opened 2022-01-08
Fix leading zeros and Hamlib debug

zcsahok:
Fixing 2 minor issues:
* QSO number is sent without leading zeros due to a glitch in 5fca304
* Hamlib log level not set to DEBUG when starting without rig control. Setting got lost in 90e9c98
image

First comes a (failing) test for leading zeros...


zcsahok:
(commited wrong file, starting over)

PR #301, opened 2022-01-08
Fix leading zeros and Hamlib debug

zcsahok:
Fixing 2 minor issues:

 *   QSO number is sent without leading zeros due to a glitch in 5fca304
*    Hamlib log level not set to DEBUG when starting without rig control. Setting got lost in 90e9c98
image

First comes a (failing) test for leading zeros...

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -711,6 +711,8 @@ static int databases_load() {
 
 static void hamlib_init() {
 
+    rig_set_debug(RIG_DEBUG_NONE);
+
     if (no_trx_control) {
 	trx_control = false;
     }
--- a/src/sendbuf.c
+++ b/src/sendbuf.c
@@ -197,18 +197,18 @@ void ExpandMacro(void) {
 
     if (NULL != strstr(buffer, "#")) {
 	int leading_zeros = 0;
-	int lead = 1;
+	bool lead = true;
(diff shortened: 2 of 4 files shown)


zcsahok:
... and the fix. (incl. boolifying a variable)

zcsahok:
And re-applying 810dbb4 to fix Hamlib logging.

src/main.c
```diff
@@ -711,6 +711,8 @@ static int databases_load() {
 
 static void hamlib_init() {
 
+    rig_set_debug(RIG_DEBUG_NONE);
+
```
dl1jbe, line 715: Maybe better in hamlib_init(). Not sure if the details of setting hamlibs  debug level is a responsibility for main().

dl1jbe:
Otherwise all fine.

src/main.c
zcsahok: Did you mean `init_tlf_rig()`?
It is not called if trx control is disabled.
(`rig_set_debug` is already in `hamlib_init()`)

zcsahok: I think I was reading the comments in wrong order, sri. So is this fine?

dl1jbe: No, sri. Was my fault. I thought you put it back to main(). It is fine at the start of hamlib_init().

PR #302, opened 2022-01-10
Rescore edit of last qsos

dl1jbe:
Tlf supports editing the last 5 qsos. As we now generate the score of each qso programmatically that should also happen after any edit. 

The patch drops the possibility to edit multiplier and point score as they get generated automatically. It furthermore respects now the contest specific exchange_width - first commit.

After any edit (insert, delete or overwrite of any character) a reload and rescore is executed - second commit.

The third commit forces an automatic rewrite of the log file without user intervention in such an case. That works also in the case of a change while :EDITing the log directly.

### DIFF
--- a/src/edit_last.c
+++ b/src/edit_last.c
@@ -33,6 +33,7 @@
 #include "globalvars.h"		// Includes glib.h and tlf.h
 #include "keystroke_names.h"
 #include "logview.h"
+#include "readcalls.h"
 #include "scroll_log.h"
 #include "tlf_curses.h"
 #include "ui_utils.h"
@@ -99,6 +100,7 @@ void putback_qso(int nr, char *buffer) {
 
 void edit_last(void) {
 
+    bool changed = false, needs_rescore = false;
     int j = 0, b, k;
     int editline = NR_LINES - 1;
     char editbuffer[LOGLINELEN + 1];
@@ -141,7 +143,11 @@ void edit_last(void) {
 	} else if (j == KEY_UP) {
 	    if (editline > (NR_LINES - nr_qsos) && (editline > 0)) {
 		unhighlight_line(editline, editbuffer);
-		putback_qso(nr_qsos - (NR_LINES - editline), editbuffer);
+		if (changed) {
+		    putback_qso(nr_qsos - (NR_LINES - editline), editbuffer);
+		    needs_rescore = true;
+		    changed = false;
+		}
 		editline--;
 		get_qso(nr_qsos - (NR_LINES - editline), editbuffer);
 	    } else {
@@ -154,7 +160,11 @@ void edit_last(void) {
 
 	    if (editline < NR_LINES - 1) {
 		unhighlight_line(editline, editbuffer);
-		putback_qso(nr_qsos - (NR_LINES - editline), editbuffer);
+		if (changed) {
+		    putback_qso(nr_qsos - (NR_LINES - editline), editbuffer);
+		    needs_rescore = true;
+		    changed = false;
+		}
 		editline++;
 		get_qso(nr_qsos - (NR_LINES - editline), editbuffer);
 	    } else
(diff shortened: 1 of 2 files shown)


zcsahok:
Will check it over the weekend (HA-DX).

dl1jbe:
Fine. GL in contest.

zcsahok:
No issues found, but when we touch this then it'd be worth fixing potential problems caused by filling the space between the fields.
Say a typo around the QSO number and time (adding an extra leading digit, thus removing the space) can lead to catastrophic effects:
image
image
image

Limiting editing to the the actual fields would solve this.

dl1jbe:
Yes I see. 
How about only allowing to modify call and exchange. As the other parts are generated automatically, there should be no need to edit these fields.

ha5se:
Well, still I would strongly suggest to also allow freq/mode modify. It happened to me several times that the RIG CAT control has been lost, unnoticed. Then the freq/mode logging has been frozen to the last one from CAT. Then, frozen CAT by me continued to remain unnoticed even during band or mode change. Fortunately, usually I noticed it within some half an hour more or so, and then I still could more or less correctly remember the freq change and fortunately could right once correct the last QSOs.

Still I absolutely agree that into field separators should never be allowed to type in.

dl1jbe:
Thanks for the feedback. I see the need for correction in that case. 

There are some more things to think about: Actual solution only allows to edit the last five lines from within TLF. Earlier contacts needs an editor for modification. In such situation we have no way to force a correct placement of the line elements.  The actual logline format (extended TRLog format)  is quite brittle in that relation. 

I think it needs a full fledged log editor which is able to modify each logline (and maybe also allow a post contest input) is needed in the long run.

The question for now is if we allow to edit ANY field (besides points and multis) or if we can concentrate on a part of it (band, mode, call, exchange , ...)? That would leave date, time and qso number untouched.

zcsahok:
I consider editing the log file as an expert use case, where the user knows what they are doing. Editing the last 5 on the other hand is provided by the UI, so at least some minimal protection against error is expected.

I'll come up with a solution where editing is restricted to the actual fields.

ha5se:
Well, I agree with Zoltán, I also think that nobody wants to shoot self on the foot. I think the file format is self-explaining. To implement all extra protection would be a waste of time, IMHO. However, a short warning in README could perhaps be  useful, suggesting making a backup before manual editing. Moreover, by reading the file, base checking for fields would reveal trivial typos, and would be rather easy to implement. E.g. files with the above screenshot example `"20:2700002"` could easily be rejected upon verifying the time field for  regex `"^[0-2]?[0-9]:[0-5]?[0-9]$"` , or even upon number of fields. (much simple time verification would do as well, `"^.?.:..$"` is also good enough for typos: if somebody insists on HH:MM as `34:99` it is his business ;) ).

Another consideration could be the maximum editing of last five. I think it would be easy to allow cursor-up and cursor-down keys to navigate the "current" line in the whole file, thus allowing edit, and perhaps END key to go back to actual EOF (current) again.

dl1jbe:
> I consider editing the log file as an expert use case, where the user knows what they are doing. Editing the last 5 on the other hand is provided by the UI, so at least some minimal protection against error is expected.
> 
> I'll come up with a solution where editing is restricted to the actual fields.

Well, just go ahead. I will merge in the actual solution as it is no regression relative to the existing code in master. It will be need for the next PR to make the qso_array permanent.

Next step will be to use 'qso' struct for scoring.

dl1jbe:
> Well, I agree with Zoltán, I also think that nobody wants to shoot self on the foot. I think the file format is self-explaining.

That is mostly true, but it is to easy to wrongly configure your editor to insert tabs instead of spaces - we already had such problems coming up.

> To implement all extra protection would be a waste of time, IMHO. However, a short warning in README could perhaps be useful, suggesting making a backup before manual editing.

That should be done. Good idea.

> Another consideration could be the maximum editing of last five. I think it would be easy to allow cursor-up and cursor-down keys to navigate the "current" line in the whole file, thus allowing edit, and perhaps END key to go back to actual EOF (current) again.

That is what I was speaking about, when I mentioned a full fledged log editor. There is more to it than just simple going up and down. Imagine a log with some hundreds qsos - I would at least quickly opt for a search function.

PR #303, opened 2022-01-13
Fix for rescore on log read

dl1jbe:
Scoring for contests with country or continent based scoring (MY_CONTINENT_POINTS of DX_POINTS set) results always MY_COUNTRY_POINTS. 

Reason is empty 'hiscall' variable during readcalls(). country_found() will therefore always use operators own country for evaluation.

The fix provides the call sign of evaluated station in 'hicall' variable.

That is only a quick fix. The underlying problem is the code in country_found() which needs to be fixed.


zcsahok (approved):
OK. `score_qso`'s missing argument will be probably the solution.

dl1jbe:
Yes. That should be one of our next steps.

PR #305, opened 2022-01-22
Keep parsed qsos

dl1jbe:
Keep 'qso_array' as parsed representation of the log in memory. 

Keep array up to date while adding or editing new contacts.
Each entry keeps also a copy of the original log line. so we can move away from using the fixed size 'qsos[]' array.

### DIFF
--- a/src/last10.c
+++ b/src/last10.c
@@ -28,17 +28,15 @@
 #include "get_time.h"
 #include "globalvars.h"		// Includes glib.h and tlf.h
 
-
 int last10(void) {
-
-    char input[LOGLINELEN + 1];
 
     int minsbefore;
     int minsnow;
     int span;
-    int counter;
+    int index;
     int qsocount = 0;
     int thisband;
+    struct qso_t *qso;
 
     if (nr_qsos < 10)
 	return (-1);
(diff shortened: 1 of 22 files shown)


dl1jbe:
All done. Should be ready for merge now.

src/last10.c
```diff
@@ -28,49 +28,44 @@
 #include "get_time.h"
 #include "globalvars.h"		// Includes glib.h and tlf.h
 
-
 int last10(void) {
```
zcsahok, line 31: Not related to this PR: `last10()` considers only the hour+minute,  +/- 1 day. So opening an older log file sometimes shows last10 as if it they would be recent QSOs.

src/searchlog.c
```diff
@@ -373,37 +375,39 @@ int bandstr2line(char *buffer) {
...
 	}
 
-	g_strlcpy(s_inputbuffer, qsos[qso_index] + 29, 13); /* call */
-	if (strstr(s_inputbuffer, hiscall) == 0) {
+	if (strstr(qso->call, call) == 0) {
 	    continue;   // no match
 	}
 
-	g_strlcpy(searchresult[srch_index], qsos[qso_index], 81);
+	g_strlcpy(searchresult[srch_index], QSOS(qso_index), 81);
```
zcsahok, line 410: Just for the TODO list: rework searching to use parsed QSO data. Now it practically parses log line again in `extractData()`.

zcsahok (approved):
Fine with me. Just added 2 notes for future.

dl1jbe:
QSOS ist used in the following places:

1. As source for edit_last() function
2. Two times in genqtclist.c (one of it can use qso->call)
3. In write_qtclog_fm_cabr() which also can use the already parsed members
4. Rewriting the log in readcalls()
5. In scroll_log() for display of log lines and finally
6. above mentioned searchlog() function.

For 1., 4. and 5. we need the log line as text which could be generated from the parsed data. With that we could drop storing the line but would need to regenerate it later. Of that 5. is used quite often.

The other three places should be converted to using the parsed data.

src/last10.c
dl1jbe: Yes, adding the date to the calculation would be better.

PR #306, opened 2022-01-23
Use fields for editing last 5 QSOs

zcsahok:
As mentioned in #302, editing is restricted to the actual log fields. This improves "foolproofness" somewhat, but there are still issues if a space is entered where it is not expected.
image
image

A solution could be to move from token-based parsing to field-based, eventually sharing the same field definitions as in `edit_last.c`. Generally there is no full protection against such cases, but at least we can try to keep it localized.

This commit just shows the idea, some clean-up is still required.

One thing changed compared to the current behavior: tab'bing into the band/mode field lands on first character position (x=0). Previously for some reason it was x=1, the second position. The idea behind might have been that the first position is rarely used.

### DIFF
--- a/src/edit_last.c
+++ b/src/edit_last.c
@@ -50,21 +50,36 @@ typedef struct {
     int start;
     int end;
     bool tab;
+    const char *pattern;
 } field_t;
 
 static field_t fields[] = {
-    {.start = 0,  .end = 4,  .tab = true},  // band+mode
-    {.start = 7,  .end = 15},   // date
-    {.start = 17, .end = 21, .tab = true},  // time
-    {.start = 23, .end = 26},   // number
-    {.start = 29, .end = 29 + MAX_CALL_LENGTH - 1, .tab = true},   // call
-    {.start = 44, .end = 46},   // sent RST
-    {.start = 49, .end = 52},   // rcvd RST
+    {.start = 0,  .end = 5, .tab = true,    // band+mode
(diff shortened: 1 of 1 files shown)


dl1jbe:
> As mentioned in #302, editing is restricted to the actual log fields. 

Looks very good.

> This improves "foolproofness" somewhat, but there are still issues if a space is entered where it is not expected.

Maybe we could add some simple sanity checks before leaving a line (no embedded spaces, only numbers, ..). That could improve the robustness quite a lot and catch simple mistakes.

> A solution could be to move from token-based parsing to field-based, eventually sharing the same field definitions as in edit_last.c. Generally there is no full protection against such cases, but at least we can try to keep it localized.

I am not sure that it is a good idea in the long run. At the moment it is only a bad feeling back in the head. I would suggest to just keep parsing as it is for now, but keep your idea in memory.

> One thing changed compared to the current behavior: tab'bing into the band/mode field lands on first character position (x=0). Previously for some reason it was x=1, the second position. The idea behind might have been that the first position is rarely used.

I got the same impression, but left it untouched. In my opinion either way works.

zcsahok:
Added sanity check for the fields. The checks are not exact, but they catch most syntactical issues. Now one can't leave editing mode unless the line looks right.

Made some variables file-global to avoid high number of arguments to internal functions. Moved unhighlight+putback+get block into an own function.

dl1jbe:
Well done. Thanks. 
Will merge it in now.

PR #307, opened 2022-01-29
Use struct qso for scoring

dl1jbe:
Do no longer use global variables to provide qso information to scoring functions. Rely only on information from actual parameter 'struct qso_t *qso'.

### DIFF
--- a/src/getctydata.c
+++ b/src/getctydata.c
@@ -63,7 +63,7 @@ void change_area(char *call, char area) {
 /* prepare and check callsign and look it up in dxcc data base
  *
  * returns index in data base or -1 if not found
- * if normalized_call ptr is not NULL retruns a copy of the normalized call
+ * if normalized_call ptr is not NULL returns a copy of the normalized call
  * e.g. DL1XYZ/PA gives PA/DL1XYZ
  * caller has to free the copy after use
  */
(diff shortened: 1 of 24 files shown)


src/getctydata.c
```diff
@@ -157,14 +157,20 @@ int getpfxindex(char *checkcallptr, char **normalized_call) {
...
-int getctynr(char *checkcall) {
+int getctynr(char *call) {
     int w;
 
-    w = getpfxindex(checkcall, NULL);
+    w = getpfxindex(call, NULL);
 
     if (w >= 0)
-	return prefix_by_index(w)->dxcc_index;
+	return prefix_by_index(w)->dxcc_ctynr;
```
zcsahok, line 173: Renaming `dxcc_index` to `dxcc_ctynr` adds a slight asymmetry to the naming.

On the other hand `countrynr = pfx->dxcc_ctynr` (line 196) is improved. Not sure which is better.

zcsahok (approved):
Fine with me.
Just have a look at the index vs. nr comment.

src/getctydata.c
dl1jbe: I totally agree and see the discrepancy myself. 
While technically it is only an index into the dxcc array we use these index logically as country number. 
I opted for a naming according to the logical meaning. Maybe you have a better idea. Feel free to step in.

PR #308, opened 2022-01-29
Fix display of call input field length

zcsahok:
Found in #306: call input field length is shown too short. Input actually works correctly, just the background highlight was one character off.

Here the editing field shows the 13 character size:
image


dl1jbe:
Good catch. Thanks for fixing.

Issue #310, opened 2022-02-21
Point scoring for WAE on restart is broken 


dl1jbe:
After PR #292 merged in WAE point scoring ignores the QTC points on startup.

@zcsahok had the right fear already then:
> On startup `readqtccalls()` is called before `readcalls()`. Hopefully the order is not relevant.

As it turns out, the order is relevant. `readqtccalls()` counts the QTC points and `readcalls()` added to that number. New code resets `total` points to 0 during start of `readcalls()`.

Rescoring with `log_read_n_score()` is free from the problem as the order of evaluation is the other way round here. Maybe we should switch to using that function during startup. I will look into it.

dl1jbe:
Fixed.

PR #311, opened 2022-02-21
Fix scoring QTC points for WAE contest

dl1jbe:
- Calling `readqtccalls()` only after `readcalls()` adds QTC points to
  the regular QSO points which are already counted by the later
  function.

- Due to different use of the 'interactive' parameter during startup we
  should not reuse `log_read_n_score()` here.

Will fix issue #310.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -700,7 +700,6 @@ static int databases_load() {
 	    showmsg("QTCs giving up");
 	    return EXIT_FAILURE;
 	}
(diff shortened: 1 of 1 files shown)


zcsahok (approved):
OK, just added a comment on log_read_n_score.

dl1jbe:
Comment is fine and makes it more clear. Thanks

PR #312, opened 2022-02-22
Fix wpx scoring

dl1jbe:
Lately scoring wpx used actual bandindex value from rig instead of the
value from the evaluated qso. That led to errors during any time we did a
re scoring (on startup, after edit of last qso or :rescore command).

Furthermore fix also bandwidth handling which had the same problem.


zcsahok (approved):
Looks good, merging it.

PR #315, opened 2022-02-26
Minor doc fixes

zcsahok:
* `README.md`: added direct links to the manual and also an internal link to the Quick Start section. Also fixed man page reference at the bottom.
* `share/logcfg.dat`: removed explicit version number as the file is valid for the current one anyway. Ref to 1.1.0 was misleading, tnx to OZ1GNN for pointing this out. Also changed the reference to the non-existing _tlf.doc_ to the man page.


dl1jbe:
Thanks

Issue #316, opened 2022-02-28
Feature Request: Letter-by-Letter Voice Keying


alandove:
Hey, folks:

In setting up for ARRL DX SSB this coming weekend, I've been looking into voice keying from TLF. While the current voice keying option seems to work well with my setup, I also saw VK6FLAB's clever [ssbdaemon](https://github.com/vk6flab/ssbdaemon) project, which is a workaround allowing letter-by-letter voice keying. This is a major feature in N1MM+ that allows one to work an entire SSB contest without saying a word, a very appealing proposition for me as my wife somehow doesn't appreciate hearing me shouting random callsigns repeatedly from the basement.

Onno's script provides this function via a hack with cwdaemon's debug mode, which seems to work but has some side effects. TLF has to be in CW mode to do this, which means manually resetting the radio back to SSB after every band change. It also means TLF will log these SSB contacts as CW, and that cwdaemon can't switch to CW keying without restarting. Neither problem is a big deal in a single-mode SSB contest, but in a mixed-mode contest it could be very tricky to get both SSB and CW keying working, and to fix the log file to indicate the correct mode for each QSO afterward.

I'm not much of a C coder, but as I understand it, the current voice keyer system calls the play_vk script, passing a filename pulled from the rules configuration file. If it could instead pass the string from the keyer macro function directly, then putting Onno's file-search algorithm into play_vk would allow it to play files named after arbitrary characters. At least, I think something like that would work. 

Regardless of the specific mechanism, I think this would be a great feature that a lot of SSB contesters would appreciate.

zcsahok:
The voice keyer functionality supports only playing Fn messages configured by VKMn/VKSPM/VKCQM parameters. One option could be to expand the configured value as a message text (as if it'd be sent to the CW keyer in CW mode) before passing it to the script.

Example:
- current setup 
```
VKM3 = f3.wav  
# --> ./play_vk f3.wav
```
- with "playable" voice message
```
VKM3 = @ you are 59 #
# --> ./play_vk "EA1ABC you are 59 123"
```
Then it'd be up to the script to interpret its input and there would be also no interference with the CW keying setup. Expanding currently used wav file names has normally no effect on them.

alandove:
Yes, and I think as long as the voice keyer function feeds the VKM strings as expanded text, Onno's `ssbdaemon` code should work more or less as written. In practice, the macros would probably be more like:

```
VKM3 = @ EXC
# --> ./play_vk "EA1ABC EXC"
```
[...]

PR #317, opened 2022-03-02
Stop voice keyer messages

dl1jbe:
Start voice keying in an separate process so TLF can look for ESC key and send the play_vk script an TERM signal.

Make ´autocq()´ honor the play time of the message.

For testing make sure to copy the play_vk script either to our bin directory or into the actual working directory.

These code is an updated part of PR #276. For details and original discussion please have a look there.


dl1jbe:
> Looks good.

Thanks for checking. will merge in now.

> I'll check how #316 could be implemented.

Very well.

Issue #319, opened 2022-03-06
COUNTRYLIST parsing issues


zcsahok:
Specifying a long COUNTRYLIST can lead to buffer overflow. Additionally, long records are not parsed correctly.
(reported by Gerard ZL2GVA)

How to reproduce:

- create a country list file (countrylist.txt) with a record of 300+ chars
- set COUNTRYLIST=countrylist.txt
- the last country of the record (actually all starting from ~255 chars) will not be parsed

Causes:

1. in `cfg_countrylist` a buffer of 255 chars is copied to `country_list_raw` that is only 50 chars long. This is the same code as in 1.4.1.
2. config file lines are limited to 159 chars even though COUNTRYLIST can be as long as 255 times 5 characters.

Other country list related code parts also have to be reviewed for buffer overflow and functional issues.

PR #320, opened 2022-03-07
Fix countrylist parsing

zcsahok:
Added tests to illustrate #319.
Extended line length for config file parsing.


zcsahok:
The fix
- cleans `countrylist` on start
- does error checking and reporting
- the latest COUNTRYLIST wins. The original code behaved differently (first was effective) for some unknown reason.

airween (changes requested):
These changes are looks good to me.

My only note is that it would be fine to add this modification (I mean the `:` (colon) syntax of `COUNTRYLIST` disappears) to `ChangeLog`, `NEWS`, and the affected files under `rules/`.

zcsahok:
The _inline_ country list was borken: even the man page was not clearly stating if contest name is needed. Anyway, the example config
```
COUNTRYLIST=Scandinavia:SM,LA,OZ,OH
```
resulted in `Scandinavia`, `SM`, `LA`, `OZ` and `OH` added to the list. Luckily `Scandinavia` is not a valid prefix, but there was a potential clash.
For inline specification the contest name is anyway superfluous, so it's just ignored. Parsing starts after the optional colon.
The contest name in this case is not unsupported, only deprecated: existing configs keep working as-is.

In fact COUNTRYLIST is used the both with and without contest name in example rules:
```
rules/pacc_dx:COUNTRYLIST=PA
rules/okom:COUNTRYLIST=okom:OK,OM
```
I'll update such examples.

zcsahok:
Just a note: CONTINENTLIST has a similar code with similar issues. These were already identified during the rewrite of `parse_logcfg.c`, but as there were no complaints the fix was postponed.

airween:
> but as there were no complaints the fix was postponed.

Yeah, luckily there are only seven continents :).

airween (approved):

zcsahok:
fixed CONTINENTLIST too

PR #323, opened 2022-03-10
Manual Updates: SSB and Other Changes

alandove:
After getting voice keying to work during the ARRL-DX SSB contest, I've updated the relevant section of the manual. I've also made a few other minor edits for better clarity.


zcsahok:
It seems that the reduction of line length in a7d3e4d got somehow lost. Lines are now extended to 80 chars that makes the user (at least me) think whether the sentence continues on that line (need to scroll right) or the next one.

image

The line on the top ending in "All entries" is still fine, there is ample margin to guess that it continues on the next line.

(I see there were issues with git itself. My suggestion is to keep master branch in sync with the master of the original repo and do all PR work in separate branches stemming from master.)

alandove:
Oh, sorry. What line length should I set instead?

N0NB:
* On 2022 12 Mar 05:39 -0600, Alan Dove wrote:
> Oh, sorry. What line length should I set instead?

80 should be the absolute maximum.  Editors like Vim or Emacs will break
lines so that a word doesn't extend past that point.  They also have
commands to wrap a paragraph so that when editing and text is inserted
or deleted it is an easy matter of realignment (my life was much
improved upon learning those commands!).

There are suggestions to set the breaking column at 76 for source code
so that there is room for the extra "+ " or "- " prepended to a line in
unified diff files, like that Git generates so the lines don't exceed 80
characters.

Then there are suggestions that when editing emails to set the right
column to 72 characters to allow for quoting characters prepended at the
left margin.

I use Vim for editing and for mail I have its 'textwidth' variable set
to 72 and the three paragraphs above all have the longest line broken
after the 72nd character.

73, Nate

-- 
Web: https://www.n0nb.us

zcsahok:
Thanks, Nate. I believe things were still OK-ish after commit a7d3e4d. Current master (51fa18ddba) has mostly lines with 78 chars or less except for 7 special cases. 
For me 78 is the maximum for good readability on a 80-wide terminal.

As the master branch of your repo @alandove got too entangled with the one of the main repo, my suggestion is to start this PR from a fresh fork and in a separate branch (see also https://docs.github.com/en/pull-requests/collaborating-with-pull-requests).

The plan would be -

1. save all your local edits, rename your local directory (e.g. `tlf` -> `tlf_save`)
2. delete your fork and re-fork tlf on github GUI
4. clone your newly forked repo; `git describe` should give `tlf-1.4.1-252-g51fa18d`
5. make a branch for the PR; e.g. using CLI: `git checkout -b update_ssb_docs`
7. apply the edits; copy over only the changed parts from your saved version (not the whole document) and keep an eye on line length.
8. add+commit+push (push may require some extra options if using CLI; this will be clearly indicated, however)
9. create PR based on your branch

alandove:
Wow, is all that really necessary? I can fix the line length easily - I just had it set to 80, and can knock that down to 78 (or 72) in seconds, then commit and push the change, can't I? Or is there some other issue that necessitates deleting and re-forking?

In future, I can certainly make changes on branches.

zcsahok:
It's not just about line lengths. This was just the point where I started looking into this more thoroughly as I knew this has been fixed already.
This PR changes probably only a section of Manual.md, but the list of commits contains 20+ entries, most of them coming from the previous documentation update.
This already shows that there's an issue here.

Here's the current state of your master branch:
image

Normally it shall be in sync with the main repo's master. (or at most lagging by some commits)

It has to be fixed anyway, the sooner the better/easier.

zcsahok:
For the reference here's how my master looks like. "_Fetch upstream_" lets you sync with the main repo, but I doubt it would work if things have diverged. Hence the suggestion to re-fork it.
image

alandove:
"Fetch upstream" worked fine just now. Here's my current state:

82C380D9-5425-462B-8EB5-8AAC005E06A3

zcsahok:
Yes, but it's somehow 27 commits ahead of the main master. And most of those commits are in fact part of the main master.
I have no idea now to fix this apart from reforking.

alandove:
Right, I don't know why it's got all those previously merged commits in there. Okay, I'll nuke my fork and start fresh. Not today, though.

PR #324, opened 2022-03-14
Doc updates

alandove:
Let's try this again. I deleted my original fork, created a new fork, cloned it locally, created a new branch, merged in the edits to README.md and doc/Manual.md, and then set line lengths to break at 78 characters or less. Please let me know what I screwed up this time and I'll see if I can fix that too. Sorry for being such a noob with git.

### DIFF
--- a/doc/Manual.md
+++ b/doc/Manual.md
@@ -1,22 +1,22 @@
 # TLF, a Contest Logger for Linux
 
 There are two types of amateur radio logging programs: general-purpose loggers
-and contest loggers. A good general purpose logger has lots of features for
-lots of different activities within the hobby: chasing DX, keeping track of
+and contest loggers. A good general purpose logger has lots of features for lots
+of different activities within the hobby: chasing DX, keeping track of
 ragchewing buddies, managing incoming and outgoing QSLs, monitoring progress
-toward various awards, and managing a lifetime worth of amateur radio
-contacts. Ham Radio Deluxe on Windows, MacLoggerDX and RUMLogNG on Mac, and
+toward various awards, and managing a lifetime worth of amateur radio contacts.
+Ham Radio Deluxe on Windows, MacLoggerDX and RUMLogNG on Mac, and
 [CQRLog](https://www.cqrlog.com/) on Linux are excellent examples of
 general-purpose loggers.
 
-Contest loggers, in contrast, are laser-focused on making rapid-fire
-competitive radiosport contacts, either in a radio contest or on a DXpedition.
-A good contest logger maximizes operator efficiency and ergonomics for hams
-who want to make hundreds of QSOs per hour for hours on end. It will typically
-display just a minimal logging window for the essential contest exchange, the
-operator's current score, DX spots for Assisted or Multi-Operator competitors,
-and a few other tactically important bits of data. N1MM+ on Windows and
-Skookum Logger on Mac are excellent examples of contest loggers. 
+Contest loggers, in contrast, are laser-focused on making rapid-fire competitive
+radiosport contacts, either in a radio contest or on a DXpedition. A good
+contest logger maximizes operator efficiency and ergonomics for hams who want to
+make hundreds of QSOs per hour for hours on end. It will typically display just
+a minimal logging window for the essential contest exchange, the operator's
+current score, DX spots for Assisted or Multi-Operator competitors, and a few
+other tactically important bits of data. N1MM+ on Windows and Skookum Logger on
+Mac are excellent examples of contest loggers. 
 
 TLF is a contest logger for Linux.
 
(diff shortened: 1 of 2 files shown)


zcsahok:
We are getting there. :-)
Is there are a way to edit a file without reformatting it? (either explicitly or implicitly)
The issue with reformatting is that it introduces a lot of changes and it's impossible to figure out the actual change in the noise.
Let's take `README.md` as it's a smaller file: in the first commit it's perfectly clear that the part starting in "If you choose this route" has been extended. The text is reformatted in the second commit (d7c1525) making it hard to find the actual change. One has to comb through a bunch of diffs like this and spot the real one:
image
Line lengths were just OK in the master branch, that part has been fixed previously.

alandove:
Line length is the crux of this whole mess, so we need a consistent way of addressing it - other than "everyone must use Vim." In Visual Studio Code, I have a plugin that resets the line length. Since learning that we needed hard line breaks in the docs, my workflow has been: edit the file as usual, select all, apply the line length plugin, save.

From the diff example you included above, I'm not convinced that the previous version was fine. It looks like line lengths were quite inconsistent, and my update smoothed them out. Maybe if we're going to insist on formatted line breaks for plain text, we'll just have to tolerate a lot of diffs that fix only the formatted line breaks?

zcsahok:
The previous version was not using all available space, that's true. My concern is that if someone say changes a single word and uses a different line length or different plugin then the diff would be again enormous and hiding relevant stuff.

This is similar to what have on code parts: to unify indenting we use the `astyle` tool that normalizes code formatting thus keeping our focus on the changed parts. Would there something like this available for text files? (I'm not a docu-pro)

alandove:
I don't know. This has never come up in my professional writing, where someone else handles page layout and the only hard line breaks in my text are the ones separating paragraphs. Soft-wrap is the default. I think if we just specify "set maximum line length to 78 characters (or whatever number we pick), break at spaces," then future updates shouldn't be so bad. There shouldn't be much disagreement between different text editors on where to break the line if we've all set the same specifications, right? Then future edits that insert a few words will only break the immediately adjacent lines.

The problem here seems to be that someone used shorter breaks for many lines in a previous reformatting, presumably after I'd submitted my initial no-line-breaks version. Maybe they picked 72 character columns, or broke some lines manually?

zcsahok:
Re tooling this SX question provides some pointers: https://unix.stackexchange.com/questions/25173/how-can-i-wrap-text-at-a-certain-column-size . Particularly `fmt` looks promising. It's available as a part of Debian's coreutils package.

alandove:
Sure, there are a bunch of ways to do this. I think `fmt` or `wrap` can produce the same results as the plugin I'm using in VSCode, or a setting in Vim.

So have we agreed on maximum 78 character line length, breaking at spaces?

zcsahok:
Yes, 78 chars shall be the maximum line length for readability on an 80 wide terminal.

N0NB:
* On 2022 14 Mar 17:04 -0500, Zoltan Csahok wrote:
> Yes, 78 chars shall be the maximum line length for readability on an 80 wide terminal.

Sounds good to me.

73, Nate

-- 
Web: https://www.n0nb.us

doc/Manual.md
```diff
@@ -211,76 +212,84 @@ stopping and restarting it while trying different port configurations.
...
+for debug purposes. Please share any information you discover with the TLF
+community.
+
+A `rules` sub-directory within the local contest folder should contain a
+contest-specific rule file. As with `logcfg.dat`, this local copy will
+override the defaults, which you will want to do. This rule file must contain
+the paths to the audio files corresponding to each F-key to be used. An
+example might be: 
 
 ```
```
zcsahok, line 294: line 312: "f12" -> "F12"

"S+P" is referred to as "S&P" in other parts of the manual and the man page.

doc/Manual.md
```diff
@@ -952,7 +963,8 @@ CABRILLO=cwo
 is in the rules file for the CW Open contest. All QSOs will then be formatted
 as in the following example:  
 
-QSO: 28000 CW 2012-09-23 0815 DL1JBE        0001 Tom       HB9XXX        0023 Joe QSO: .....  
+QSO: 28000 CW 2012-09-23 0815 DL1JBE        0001 Tom       HB9XXX        0023
```
zcsahok, line 966: The second ellipsized QSO line can be dropped and the spaces in the first one reduced so that it fits one line.

doc/Manual.md
alandove: Fixed.

doc/Manual.md
alandove: Fixed.

PR #327, opened 2022-03-19
Fix SERIAL+GRID4 multiplier handling

zcsahok:
This case had a special handling that is now covered by the generic multiplier logic.

- grid loc is treated as a kind of section in `checkexchange_serial_section()`. A minor disadvantage of not introducing a specific function for grid4 is that it now allows non-grid like values ("01JN") provided they are separated with a space from the serial number. As the global max section length in 4, the locator is also truncated to this size. We could consider increasing section length to 6. This would also allow more exotic sections.
- the resulting `mult1_value` is used then in `addmult()`; there a default path was added for any future logic (e.g. plugin) that would set `mult1_value`. This mult is treated as per-band (`mult1_mode`), in future a config will be needed to control this.
- `makelogline()` handles the grid4 case also the same way as serial+section
- `getgrid()` has been dropped and `section[]` is restricted to `getexchange`.
- updated/added tests


src/addmult.c
```diff
@@ -42,6 +42,7 @@ GPtrArray *mults_possible;
 enum { ALL_BAND, PER_BAND };
 
 char mult1_value[40];
+int mult1_mode = PER_BAND;  //FIXME configure
```
airween, line 45: Is this a hard coded variable? I agree that we don't need to handle this one through config, but may be that wouldn't be a big work. Would I add it?

airween (approved):
Reviewed and tested, looks good to me.

Thank you Zoli!

src/addmult.c
zcsahok: Not a big work indeed, but for the time being I'd not commit to a config name as the general concept hasn't been worked out yet. (e.g. there shall be at least 2 mults, hence the name `mult1_...`)
IMO  for the moment it's enough to mark it as fixme.

zcsahok:
Thanks.
Added a bit more strict grid format checking.

Just realized that `locator2longlat.c` is quite a misnomer: it remained from the times when Hamlib was optional and now just contains the sole `check_qra()`.
Some files include `locator2longlat.h` but they actually use the Hamlib function.
I'm moving `check_qra()` to `util.c` and removing `locator2longlat.[ch]`.

zcsahok:
Cleaned up `prepare_fixed_part()` a bit: up to position 77 all except cqww/itumult/wazmult do the same.

PR #328, opened 2022-03-23
Use empty default proposed_exchange

zcsahok:
`get_proposed_exchange()` didn't have a default value for the no match case. Previously found value was used unless it was cleared via `cleanup_qso()`.
Ctrl-A or deleting using backspace does not call `cleanup_qso()`, hence the previous exchange value is kept.

### DIFF
--- a/src/recall_exchange.c
+++ b/src/recall_exchange.c
@@ -43,6 +43,8 @@ int get_proposed_exchange(void) {
     char *loc, *loc2;
     struct ie_list *current_ie;
 
+    proposed_exchange[0] = 0;   // default: empty (nothing found)
+
     if (strlen(hiscall) == 0)
 	return 0;
 


dl1jbe:
+1 

> Ctrl-A or deleting using backspace does not call `cleanup_qso()`, hence the previous exchange value is kept.

Maybe we should add that at least for Ctrl-A?

df7cb:
Thanks for tracking that down!

zcsahok:
Chris @df7cb could you confirm that it actually solves the issue? (just to be sure)

cleanup for Ctrl-A: I'm not even sure that clearing the call input field (as it is done now) is the right thing. The use case is to share the info, but once Ctrl-A is pressed one has to enter the call again if they want to work the station. Currently it fits only the 2nd SWL node case (i.e. no QSO logged).

df7cb:
It does fix my issue. Thanks!

I was actually wondering myself if Ctrl-A should remove the call from the input field, that seems like two independent actions. But then, it does fit my use case: in S&P mode, find a signal, type the call, realize I worked them before, hit Ctrl-A, move on. My vote would be on keeping the behavior. Maybe it should be configurable. But there are already so many options that this might not be too helpful.

Possibly related: I am missing a "grab spot *at this frequency*" button. Having not the best antenna, the "go to next spot in bandmap" button isn't too helpful, but when I tune to the next peak on the waterfall, there is often already an entry in the bandmap that I end up copy-pasting.

dl1jbe:
@Zoli > cleanup for Ctrl-A: I'm not even sure that clearing the call input field (as it is done now) is the right thing. The use case is to share the info, but once Ctrl-A is pressed one has to enter the call again if they want to work the station. 

No! Just work the station and it will sent to the other nodes (like Ctrl-A) automatically. 
Ctrl-A is only needed if you do NOT work the station, but like others to know about it or just use it to remember it for later (e.g. if you work without cluster support). For the latter maybe BMAUTOADD is more useful as it adds the spot if you just tune away from the frequency with the input field filled.

> Possibly related: I am missing a "grab spot at this frequency" button. Having not the best antenna, the "go to next spot in bandmap" button isn't too helpful, but when I tune to the next peak on the waterfall, there is often already an entry in the bandmap that I end up copy-pasting.

Just look for BMAUTOGRAB keyword in man page. As you tune in to a spot frequency TLF will copy the spot call into the input field. Maybe it is what your are looking for.

zcsahok:
Leaving Ctrl-A as-is is fine with me. We have anyway the Ctrl-A/Ctrl-B topic ("how to send a cluster spot") to be sorted out.

For automatic grabbing (if I got your use case correctly) you can use `BMAUTOGRAB`: it "grabs" the spot at the current frequency. If you tune away the call field is cleared.

zcsahok:
(sri, haven't seen your comment, Tom. we're on the same page.)

df7cb:
Ah right, @airween had even told me about BMAUTOGRAB before, I just forgot to try it. Sorry!

dl1jbe:
> Leaving Ctrl-A as-is is fine with me. We have anyway the Ctrl-A/Ctrl-B topic ("how to send a cluster spot") to be sorted out.

I will work on the Ctrl-B next.

PR #330, opened 2022-04-09
Extended score plugin with QRB calculation for Tesla Memorial

zcsahok:
Synced the branch with master (e.g. using qso_t). Removed some plugin functions that are not needed anymore due to improvements in master.
Added support for Tesla Memorial scoring.


airween:
Oh, if I'm right, this is a bigger collection of patches from the plugin's branch?

I need some time to check the functions, but I'll do that.

zcsahok:
Only the 5 commits after 61f4fb7 are new. The issue is that earlier development was done directly in the main repo (`dev_plugin` branch) but these new ones I did in my branch (`poc_plugin`) as it should be.
Generally this is work in progress, but the scoring part should be reasonably usable.

src/qrb.c
```diff
@@ -42,3 +42,14 @@ int get_qrb(double *range, double *bearing) {
...
 	       range, bearing);
 }
+
+bool get_qrb_for_locator(const char *locator, double *range, double *bearing) {
+    double dest_long, dest_lat;
+
+    if (RIG_OK != locator2longlat(&dest_long, &dest_lat, locator)) {
+	return false;   // invalid locator
+    }
+
```
dl1jbe, line 52: Comparing the use of 'qrb()' in 'get_qrb()' it seems we should have a look at our representation of the coordinates. The use of +/- prefix looks not intuitive.

zcsahok: Unfortunately `cty.dat` and Hamlib have different orientation of longitudes. It is also commented in `qrb.c` (also in master).

Example: (-19.28 vs. 19.00)
```
 $ grep Hungary cty.dat
Hungary:                  15:  28:  EU:   47.12:   -19.28:    -1.0:  HA:
$ rotctl loc2lonlat jn97
19.000000
47.500000
```

zcsahok:
The code is backwards compatible, no problem to merge it. I'd not rush with it as I have still the multiplier handling on my shortlist. (i.e. determining arbitrary mult; also using it for band map)

Hopefully the plugin interface (the functions) stabilizes by then and thus it can be taken over into the master.

airween:
> How do we plan to incorporate it into TLF in the long run?

As Zoli wrote the code is compatible with the current base (I didn't check yet so I believe him :)).

My question is how can we release the next version for package maintainers?

I mean if somebody uses Tlf from source, then the choice is available: use Python extension or not. If the answer is yes, then the user has to install the dependencies, and that's it. But what can do a package maintainer? He must to choose before makes the package. If he wants to add Tlf with Python extension, then the dependencies will come with it. May be we should find a solution to compile Tlf as standalone application, **and** beside a dynamic module (or something similar), which is optional.

@df7cb Christoph, do you have any idea, how does it work in (for eg.) Debian? I mean for example the Postfix packages: anyone can install the "native" Postfix package, but there are many modules for it, eg. `postfix-mysql`, `postfix-ldap`, and so on. Here I can imagine two packages: `tlf` and `tlf-python`.

Thoughts?

zcsahok:
`tlf-python` would then provide a shared object library (depending on python) to be used by `tlf`. Practically this means adding a layer of abstraction. Probably doable, but needs some more research.

df7cb:
Don't overcomplicate it. If tlf now requires python, so be it, and the tlf.deb package will grow the necessary dependencies. I don't think tlf is intended to run on embedded systems where the additional disk space required by python is a problem. Creating a separate tlf-python package doesn't seem worth the trouble (and extra user confusion).

airween:
Hi Christoph,

> Don't overcomplicate it. If tlf now requires python, so be it, and the tlf.deb package will grow the necessary dependencies. I don't think tlf is intended to run on embedded systems where the additional disk space required by python is a problem. Creating a separate tlf-python package doesn't seem worth the trouble (and extra user confusion).

right, thanks for the tip. In this case, I think we don't need to research how can we make a separated shared object.

src/plugin.c
```diff
@@ -281,43 +288,20 @@ int plugin_score(int band, const char *call, int mode,
...
-
-    if (NULL != PyErr_Occurred()) {
-	PyErr_Print();
-	//FIXME: action?
-    }
-    Py_XDECREF(pValue);
-#endif
-}
-
 bool plugin_is_multi(int band, const char *call, int mode) {
```
airween, line 291: Is it necessary the `const` modifier here? The `call` member of `qso_t` is [seems](https://github.com/Tlf/tlf/blob/master/src/tlf.h#L204) like just a single `char *`. Both `gcc` and `clang` generated a warning. I think we should leave it.

airween (changes requested):
I have only one note.

I think we can merge it now, I have few ideas to add to the plugin. With a common branch, we can work together.

zcsahok:
Thanks for the feedback.

PR #333, opened 2022-06-08
Improve editing

zcsahok:
Added flashing the edited field if the content is invalid. This is to remind the user that it's not TLF crashed (not possible to exit the field) but a correct value must be typed in. Could be handy when this happens at 3 a.m. ...
Fixed received RST field size and RST fields allow `---` to support the NO_RST case.

### DIFF
--- a/ToDo
+++ b/ToDo
@@ -36,7 +36,7 @@ Planned for later versions
 
 [ ] Do a complete rewrite of display related code in the long term to
     allow vertical AND horizontal resizing.
-
+[ ] Add undo functionality to editing last 5 QSOs (see PR #333)
 
 Internals
 ~~~~~~~~~
--- a/src/edit_last.c
+++ b/src/edit_last.c
@@ -62,7 +62,7 @@ static field_t fields[] = {
         .pattern = "^\\s*\\S+\\s*$"},
     {.start = 44, .end = 46,                // sent RST
         .pattern = "^\\s*\\d+\\s*$"},
-    {.start = 49, .end = 52,                // rcvd RST
+    {.start = 49, .end = 51,                // rcvd RST
         .pattern = "^\\s*\\d+\\s*$"},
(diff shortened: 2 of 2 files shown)


dl1jbe:
Looks good so far.

As a side idea - would it make sense to keep a copy of the original entry and restore it in case of error?

zcsahok:
Logically undo would sense for the current field only, else one would lose all edits on the line which may be not what the user expects. On the other hand the code considers the line as a single entity, so some extra handling would need to be added.
`Esc` is used already for exiting (this is somewhat misleading as one could expect that edits will be abandoned), so the next option could be the well known `Ctrl-Z`.
It's not a quick change and more likely a nice-to-have than a must-have feature, so let's put it on the backlog.

dl1jbe:
> Logically undo would sense for the current field only, else one would lose all edits on the line which may be not what the user expects. 

Yes, that is what I had in mind.

> On the other hand the code considers the line as a single entity, so some extra handling would need to be added. `Esc` is used already for exiting (this is somewhat misleading as one could expect that edits will be abandoned), so the next option could be the well known `Ctrl-Z`. It's not a quick change and more likely a nice-to-have than a must-have feature, so let's put it on the backlog.

+1

ha5se:
Well... Considering other common linux apps, like mcedit, undo often simply undoes keystrokes one by one, fully ignoring possible context interpretations (might perhaps please ignore TLF fields)...
In TLF, nervously pressing Esc more than once just to try to break my early sending, would delete all fields.
Backlog, up to the last logged (saved) QSO would perhaps be fine...

- Btw. for undo, Ctrl-u could perhaps be more common than Ctrl-z?
- Btw. in doc, Ctrl+? and Ctrl-? is inconsistently used... Please decide whether plus or minus? ;)

zcsahok:
Thanks for comments, added to the ToDo list.

Issue #334, opened 2022-08-01
Log truncated when exporting Cabrillo


ncw:
I used TLF (latest git version) for working the IOTA contest this weekend. TLF worked fine during the contest, but when I came to export my log to Cabrillo things started to go wrong.

The initial log looked like this

```
 40CW  30-Jul-22 13:02 0001  XXXX           599  599  145EU8                 0   7021.0
 20CW  30-Jul-22 13:12 0002  XXXX           599  599  087                    0  14003.6
 20CW  30-Jul-22 13:26 0003  XXXX           599  599  172EU002               0  14013.5
 20CW  30-Jul-22 13:42 0004  XXXXX          599  599  062EU029               0  14016.2
 20CW  30-Jul-22 13:47 0005  XXXX           599  599  076                    0  14018.4
```

But when I exported it to Cabrillo using the exchange "# EU-005" I got this

```
[snip header]
QSO:  7021 CW 2022-07-30 1302 M5NCW         599 0001 E XXXX          599 145EU8
QSO: 14003 CW 2022-07-30 1312 M5NCW         599 0002 E XXXX          599 087   
QSO: 14013 CW 2022-07-30 1326 M5NCW         599 0003 E XXXX          599 172EU0
QSO: 14016 CW 2022-07-30 1342 M5NCW         599 0004 E XXXXX         599 062EU0
QSO: 14018 CW 2022-07-30 1347 M5NCW         599 0005 E XXXX          599 076   
END-OF-LOG:
```

You can see here that my exchange has been truncated, but much more worringly the remote exchange has been truncated too. For instance "172EU0" should be "172EU002".

Have I misconfigured TLF somehow to cause this? Or is it a bug?

According to the RSGB site https://www.rsgbcc.org/hf/formats/templates.shtml the cabrillo log should look like this

```
QSO: 28000 CW 2002-07-27 1600 G3XTT         599 0001 EU-005 EA5BM         599 0031 ------
QSO: 28000 CW 2002-07-27 1601 G3XTT         599 0002 EU-005 PY1NX         599 0186 ------
QSO: 28000 CW 2002-07-27 1602 G3XTT         599 0003 EU-005 M6T           599 0451 EU-005
QSO: 28000 CW 2002-07-27 1611 G3XTT         599 0004 EU-005 9M6A          599 0310 OC-088
```

[...]

PR #335, opened 2022-08-03
Fix silently truncating Cabrillo fields

zcsahok:
Fixing #334:
When formatting a Cabrillo field (either right or left padded) the truncation is treated as an error. User is notified and exporting is stopped.
image

The same message is written into the Cabrillo file:
 ```
!ERROR: field too wide (0001 EU222)
END-OF-LOG:
```

QSO and QTC exporting has been moved into a common function `process_record()`. Length checking is done in `add_[rl]padded()` functions. They overwrite the line buffer with the error text.

### DIFF
--- a/share/cabrillo.fmt
+++ b/share/cabrillo.fmt
@@ -1,8 +1,9 @@
 # Cabrillo format descriptions for various contests
 
 # used for ARRL and CQ contests, RDXC, Oceania DX Contest, AP Sprint and others
+# (using default 14-char wide exchanges)
 [UNIVERSAL]
-QSO=FREQ,5;MODE,2;DATE,10;TIME,4;MYCALL,13;RST_S,3;EXC_S,6;HISCALL,13;RST_R,3;EXCH,6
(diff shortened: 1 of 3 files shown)


zcsahok:
Increased exchange width to 14 for the UNIVERSAL Cabrillo format.

dl1jbe:
Sorry for being late. Was away some days.

Fix looks good. Did we double check that the longer (14 chars) fields are accepted by cabrillo processing robots?

zcsahok:
I didn't check this explicitly, but as I have never had issues in this respect with Cabrillos my understanding is that parsers do not rely on exact field widths. As the specs page puts it: _All of the major contest sponsors will also accept the fields in any column as long as there is a space between each item_

With 14 chars we are at least internally consistent with our default (non-specific) exchange width.

dl1jbe:
Ok. So it should work as expected.

zcsahok:
@dl1jbe is there anything still open or it can be merged in?

dl1jbe:
As I wrote: it looks good and can merged in. I just asked to get some background as I was away and could not follow any discussion. Sorry for being not clear with that.

zcsahok:
No problem, I was just waiting for a review approval.

PR #336, opened 2022-08-04
Add a way to check for a new multiplier based on a spot

zcsahok:
This is going to be a longer PR with the goal to provide the possibility of marking new mults on bandmap for non built-in contests. Currently this feature is supported only for CQWW, WPX, ARRL DX, and the ones covered by `general_ismulti()`. There are a number of other contest that do not fit into these zone, prefix or country based rules.

Multiplier evaluation for a spot would go like this

- construct a `qso_t` object based on what is available (call, freq, optionally an guessed exchange)
- perform roughly the steps as when reading the log file in `readcalls()`, but instead of `addcall()` use `check_mult()` so that the QSO is not "logged" just the multiplier is evaluated
- based on the result mark spot as new multiplier

As most of the above functions involve manipulation of globals this has to be first ironed out. In particular, the current QSO data has to be moved into a `current_qso` object of type `qso_t`. This is the continuation of the earlier work to switch to consistently using `qso_t` and will be a multi-step process as there are many variables affected.

The first commit adds the basic multiplier checking functionality `check_mult()` by extending `remember_multi()`. The setting of global `new_mult` is factored out of `addmult()`.

### DIFF
--- a/src/parse_logcfg.c
+++ b/src/parse_logcfg.c
@@ -1010,22 +1010,28 @@ static int cfg_minitest(const cfg_arg_t arg) {
     return PARSE_OK;
 }
 
-static int cfg_unique_call_multi(const cfg_arg_t arg) {
+static int set_multi_mode(const cfg_arg_t arg, int *config) {
     char *str = g_ascii_strup(parameter, -1);
     g_strstrip(str);
 
     if (strcmp(str, "ALL") == 0) {
-	unique_call_multi = MULT_ALL;
+	*config = MULT_ALL;
     } else if (strcmp(str, "BAND") == 0) {
-	unique_call_multi = MULT_BAND;
+	*config = MULT_BAND;
+    } else if (strcmp(str, "NONE") == 0) {
+	*config = MULT_NONE;
     } else {
 	g_free(str);
-	error_details = g_strdup("must be ALL or BAND");
+	error_details = g_strdup("must be ALL, BAND or NONE");
 	return PARSE_WRONG_PARAMETER;
     }
 
     g_free(str);
     return PARSE_OK;
+}
+
+static int cfg_unique_call_multi(const cfg_arg_t arg) {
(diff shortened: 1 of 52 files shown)


zcsahok:
`comment` is taken now from a static `current_qso`. The functionality of `collect_qso_data()` will be actually creating a copy of `current_qso` for storage in the QSO array.

zcsahok:
moved `callupdate` to `current_qso`. This is used in `getexchange` only, needs not to be stored in QSO array.

zcsahok:
move `normalized_comment` to `current_qso`.

zcsahok:
move `section` to `current_qso`.

zcsahok:
This is already a usable state. I will post a PR for the plugin branch once this one has been merged.

zcsahok:
The code part is finished, let me know if there are any comments.

src/parse_logcfg.c
```diff
@@ -1252,6 +1262,7 @@ static config_t logcfg_configs[] = {
     {"DIGI_RIG_MODE",       NEED_PARAM, cfg_digi_rig_mode},
     {"CABRILLO-(.+)",       OPTIONAL_PARAM, cfg_cabrillo_field},
     {"RESEND_CALL",         NEED_PARAM, cfg_resend_call},
+    {"GENERIC_MULT",        NEED_PARAM, cfg_generic_mult},
```
airween, line 1265: It would be fine to add this new `GENERIC_MULT` directive to the manpage.

airween (approved):
This is a very fine PR, everything is traceable and clear.

I had only one note - see my comment above.

Anyway, looks good to me.

zcsahok:
Thanks for the review!

Added a fix for log corruption during start-up when lan mode is active caused by using global `lastqsonr`. Lan code needs a rework/redesign anyway.

PR #337, opened 2022-08-13
Fix Cabrillo input window background color

zcsahok:
A minor cosmetic fix: the background of the first Cabrillo input window is green whereas the subsequent ones have cyan background.
image
image

The fixed version:
image
image

### DIFF
--- a/src/nicebox.c
+++ b/src/nicebox.c
@@ -54,7 +54,7 @@ void nicebox(int y, int x, int height, int width, char *boxname) {
 void ask(char *buffer, char *what) {
 
     nicebox(14, 0, 1, 78, what);
-    attron(A_STANDOUT);
+    attron(COLOR_PAIR(C_LOG) | A_STANDOUT);
     mvaddstr(15, 1, spaces(78));
     move(15, 1);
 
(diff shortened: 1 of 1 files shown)


dl1jbe:
Looks good. 

Color and attribute handling in ncurses is always a nightmare. Unluckily the attribute settings are interpreted differently by different terminal emulators.

Tested it in xfmw4 terminal, xterm and urxvt.

N0NB:
Agreed, Tom, which is why the sample [Xresources](https://github.com/Tlf/tlf/blob/master/doc/Xresources) file is provided for Xterm and rxvt based emulators to get a console style VGA color pallet.  My experience with the desktop terminal apps is that they have a color pallet selection in the UI that often includes a VGA pallet.

Of course users are free to choose whatever color pallet they like but I think that if Tlf looks consistent in the VGA pallet then it should look consistent in other color pallets.

zcsahok:
After the first commit the cursor was not really visible initially as it had the same color as the box outline. Now the input field's color is set explicitly giving a distinct black cursor. It has the same color scheme as the keyer window, that BTW also uses `wnicebox()`, but in a more advanced way (own window, own panel, ...).

image

image

N0NB:
That looks good to me.

I see I need to try out the new Cabrillo template feature!  It has only been there for a while...

dl1jbe:
Seems to be finished. Lets pull it in.

PR #338, opened 2022-08-14
Use up/down arrows to increment/decrement serial number in exchange field

df7cb:
I've been meaning to implement this ever since I'm using tlf, but always forgot after the contest. Now finally took the time to do this during the WAEDC...

Now in S&P, you can enter the serial number some station got, and keep incrementing it by pressing "up" until the station finally logged you. (Works in LOG mode as well, but is probably less useful there.)

### DIFF
--- a/src/getexchange.c
+++ b/src/getexchange.c
@@ -64,6 +64,29 @@ static char section[MAX_SECTION_LENGTH + 1] = "";
 
 void exchange_edit(void);
 
+static void serial_up_down(char *exchange, int delta) {
+    /* length of serial part in "001" or "001 EU-001" */
+    int nr_len = strspn(exchange, "0123456789");
+    if (nr_len == 0 && nr_len > 5) {
+	return;
+    }
+    /* serial number, suffix ignored if any */
+    int nr = atoi(exchange);
+    nr += delta;
+    if (nr < 0 || nr > 99999) {
+	return;
+    }
+    /* preserve leading zeros, append old suffix */
+    char *buf = g_strdup_printf("%0*d%s", nr_len, nr, exchange + nr_len);
+    int len = strlen(buf);
+    /* length can change when overflowing 9 -> 10 */
+    if (len <= contest->exchange_width) {
+	strcpy(exchange, buf);
+    }
+    g_free(buf);
+}
+
+
 int getexchange(void) {
 
     int i;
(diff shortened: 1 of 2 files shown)


zcsahok:
A handy feature indeed.

Had some fun a on 64-bit system:
image
then after Up
image
At least it didn't overflow. :-)

df7cb:
> A handy feature indeed.

Thanks!

> At least it didn't overflow. :-)

Heh. Well, in some sense it did...

I changed the code to use atol() instead, and also limited the max length of the number to 20 so the compiler wouldn't complain that sprintf() wouldn't overflow comment[80].

zcsahok:
On a 32-bit system a down on all 9's (14 digits) still leads to 00002147483646 = 2^31 - 2.

How about these:

- limit length to 5. Normally it would be 4 but let's be more generous. Then `int` is more than enough for this. No need to take care of `exchange_width` as it is checked anyway.
- `g_strdup_printf` the updated value into a temp string
- if it fits into min(5, contest->exchange_width) chars then copy it into `comment`

I have in the queue a change to have a common code for aligning the serial number (partly controlled by  `SERIAL_EXCHANGE`). That could also be probably used for extracting the serial number even from an exchange like "12 EU3" (e.g. for IOTA).

df7cb:
Right, long is too much anyway, I went back to atoi+int now.

There is a limit up to 99999 now. No temp string involved, since printing integers has little surprises involved.

I thought about supporting serials hidden somewhere in the exchange but decided to go without first for simplicity.
Is the serial always first in these mixed exchanges? I forgot and can't think of a regexp to throw at my old logs now.

src/getexchange.c
```diff
@@ -262,6 +262,19 @@ int getexchange(void) {
...
 	    }
 
+	    case KEY_UP:	/* Up/Down--increase/decrease serial number */
+	    case KEY_DOWN: {
+		if (i > 0 && i <= 5 && strspn(comment, "0123456789") == i) { /* comment non-empty and only digits */
+		    int nr = atoi(comment);
+		    nr += (x == KEY_UP) ? 1 : -1;
+		    if (nr >= 0 && nr <= 99999) {
+			sprintf(comment, "%0*d", i, nr); /* preserve leading zeros */
+			i = strlen(comment); /* length can change when overflowing 9 -> 10 */
```
zcsahok, line 272: `i` must be `<= contest->exchange_width` even this is currently implicitly true. Exchange width may become configurable in the future and one may set it to <5.

zcsahok:
There is an issue after editing the exchange.
The steps:

- in a fresh exchange field press {space}{1}
- {left} to start editing
- {backspace} to remove the leading space
- {right} to stop editing
- now {up}/{down} do nothing
- if the remaining {1} is removed with {backspace} and re-typed then it works again

Serial comes first in all contests I know.

df7cb:
That was a separate bug, i (the comment length) wasn't updated after exchange_edit().

src/getexchange.c
df7cb: Updated.

zcsahok:
I see. Thanks for the correction.

I pushed the check for exchange_length what I meant in my comment.

(no need to force-push, it'll be squashed at the end anyway)

df7cb:
Hi Zoltan, thanks for the fix!

It occurred to me today that it is actually very easy to support "001 EU-001" since atoi() ignores the suffix anyway. I pushed an update to do just that.

zcsahok:
Nice!

Made two changes:

- switched to `g_strdup_printf()` as in the parallel PR #336 the static `char comment[]` is going to be changed to a dynamic `char *` type, thus `sizeof` won't be available.

- as the logic got a bit more complex it deserved an own function.

zcsahok (approved):

zcsahok:
@df7cb Chris, if there are no further updates then I'd merge this in.

df7cb:
@zcsahok looks good to me.

df7cb:
@zcsahok thanks!

PR #342, opened 2022-08-29
Use actual QSO mode for mode-based scoring

### LINKED ISSUE #341: Re-scoring does not consider QSO mode
Rescoring a log with mixed mode QSOs and using mode scoring (CWPOINTS/SSBPOINTS) leads to a wrong result. All QSOs are scored using the actually selected mode (normally CW).
For an example see #273.

zcsahok:
Fixes #341

On mid term we shall replace `trxmode` with `current_qso.mode` similarly to #336.

### DIFF
--- a/src/score.c
+++ b/src/score.c
@@ -135,9 +135,9 @@ int portable_doubles(int points, char *call) {
 
 
 /* apply points by mode */
-int scoreByMode() {
+int scoreByMode(struct qso_t *qso) {
 
-    switch (trxmode) {
+    switch (qso->mode) {
 	case CWMODE:
 	    return cwpoints;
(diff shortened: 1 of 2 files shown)


N0NB:
I made a quick glance at your commit and it appears reasonable.  I've not pulled and tested it yet.

N0NB (approved):
At first glance these changes look reasonable to me.  I have pulled and built it and the rescore on startup did not change the point values in the log.

N0NB:
I see that the QSO point score matches that of my Perl script although the multipliers are way too many but that is another bug.

zcsahok:
Thanks for checking!

Issue #273, opened 2021-08-29
Mult is added in scoring window when no new mult is worked


N0NB:
Yesterday I noticed that seemingly at random the score window would increment the mult count when no mult was worked and the log window correctly showed that no mult was worked.  Using `:REScore` or restarting Tlf resulted in the log file being read and the mult count and score being corrected,

From my rules file:

```
#################################
#      Scoring rules            #
#################################
#
MIXED
RECALL_MULTS
SSBPOINTS=2
CWPOINTS=3
SECTION_MULT_ONCE
### WYSIWYG_ONCE
MULT_LIST=ksqp.txt
```

I will note that I had erroneously had `SECTION_MULT` instead and after stopping Tlf, adding `_ONCE`, manually editing out the erroneous section mults, and restarting Tlf, the mult count in the scoring window was correct.

I did not observe that incrementing the mult count in the scoring window was related to `RECALL_MULTS` as it seemed to do it for both recalled callsigns and stations not worked before.

I've attached the section mults file.

[ksqp.txt](https://github.com/Tlf/tlf/files/7072206/ksqp.txt)

zcsahok:
I'm looking at multiplier handling at the moment, so may find something related to this issue.
What was your `CONTEST` setting?
By "log window" do you mean the log line showing/not showing the worked mult?

N0NB:
Hi Zoli.

Here is the contest section:

```
########################
# General contest mode #
########################
#
CONTEST=ksqp
LOGFILE=ksqp_2021.log
CONTEST_MODE
CABRILLO=UNIVERSAL
CALLMASTER=MASTER.SCP
```

I noticed that the issue doesn't seem to occur as often in SSB mode.

N0NB:
At the end of the party I ran a full hour of CW and the mult value in the scoring window was rather stable.

By "log window" I mean the mini-log that appears above the entry line.

zcsahok:
Did I get it right, that you have experienced that mults kept "randomly" added after logging a CW QSO?
Was the log line correctly showing or not showing the mult? I.e. only the total sum was wrong? Or both were off?

Could you please post your configs?

N0NB:
The log line was correct but the scoring window was way off.

Rules file `ksqp`:

```
########################
# General contest mode #
########################
#
CONTEST=ksqp
LOGFILE=ksqp_2022.log
CONTEST_MODE
CABRILLO=UNIVERSAL
CALLMASTER=MASTER.SCP
#
##################################
[...]

PR #343, opened 2022-08-30
Don't force per-band mults for MULT_LIST

zcsahok:
Fixing #273: when using MULT_LIST the condition
```
    } else if (multlist == 1 && !CONTEST_IS(ARRL_SS)) {

	return totalmults ;
```
selected effectively per-band counting of multipliers. (`totalmults` is the sum of mults worked per band)
The conditions involving `sectn_mult_*` came after this check and thus were not effective. This led to a seemingly random increase of the displayed mult count even if the log showed the actual multipliers correctly.

Solution: moved the check to the end of the chain. Probably we can even drop it altogether as it looks like some legacy code.

Also visually streamlined the evaluation of `total*` sums.

### DIFF
--- a/src/showscore.c
+++ b/src/showscore.c
@@ -128,26 +128,20 @@ int get_nr_of_points() {
 /* get total number of multis */
 int get_nr_of_mults() {
 
-    int n;
-    int totalzones;
-    int totalcountries;
-    int totalmults;
-
     if (!iscontest)
 	return 1;
 
-    /* precalculate summaries */
-    totalzones = 0;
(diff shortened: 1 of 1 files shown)


dl1jbe (approved):
No changes needed. I confirm the PR. Go on with it.

Let us hope that we do not find a possible configuration which requires to shuffel the lines around once more.

The whole set of keywords for mult handling got a little out of hand by adding special cases over special cases. I think in the long term a complete overhaul of the system is needed.

zcsahok:
Thanks.

PR #345, opened 2022-09-27
Make buffer long enough to accommodate long lines of aliases

VictorDenisov:
Long line of aliases for CQP is cut by short buffer.
I'm not sure about the exact buffer length though. Please advise.

CA:ALAM,ALPI,AMAD,BUTT,CALA,COLU,CCOS,DELN,ELDO,FRES,GLEN,HUMB,IMPE,INYO,KERN,KING,LAKE,LASS,LANG,MADE,MARN,MARP,MEND,MERC,MODO,MONO,MONT,NAPA,NEVA,ORAN,PLAC,PLUM,RIVE,SACR,SBEN,SBER,SDIE,SFRA,SJOA,SLUI,SMAT,SBAR,SCLA,SCRU,SHAS,SIER,SISK,SOLA,SONO,STAN,SUTT,TEHA,TRIN,TULA,TUOL,VENT,YOLO,YUBA

### DIFF
--- a/src/addmult.c
+++ b/src/addmult.c
@@ -336,7 +336,7 @@ void add_mult_line(char *line) {
 int init_and_load_multipliers(void) {
 
     FILE *cfp;
-    char s_inputbuffer[186] = "";
+    char s_inputbuffer[700] = "";
     char *mults_location;
 
     if (mults_possible) {
@@ -364,7 +364,7 @@ int init_and_load_multipliers(void) {
 	return 0;       // couldn't open file
     }
 
-    while (fgets(s_inputbuffer, 85, cfp) != NULL) {
+    while (fgets(s_inputbuffer, 300, cfp) != NULL) {
 
 	/* strip leading and trailing whitespace */
 	g_strstrip(s_inputbuffer);
--- a/src/addmult.c
+++ b/src/addmult.c
@@ -336,7 +336,7 @@ void add_mult_line(char *line) {
 int init_and_load_multipliers(void) {
 
     FILE *cfp;
-    char s_inputbuffer[700] = "";
+    char s_inputbuffer[2000] = "";
     char *mults_location;
 
     if (mults_possible) {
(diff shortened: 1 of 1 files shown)


zcsahok:
Thanks for the PR!

For reading multiplier file let's use the same line length (2000) as for parsing config files in `parse_configfile()`. As for some reason the buffer length and the size specified in `fgets` have drifted apart let's fix that one too:
```fgets(s_inputbuffer, sizeof(s_inputbuffer), cfp)```

The multiplier file can contain several lines per mult, so it is possible to split them like
```
CA:ALAM,ALPI,AMAD,BUTT,CALA,COLU
CA:CCOS,DELN,ELDO,FRES,GLEN
...
```
(man page: `You  can have more than one line for the same multiplier.`)

dl1jbe:
As the problem with long lines pops up from time to time, maybe we should check all 'fgets()' invocations for overrun of the supplied buffer. Reading the man page the only way seems to look for a trailing newline.

There is another function 'getline()' which seems easier to use.

zcsahok:
@dl1jbe Good point regarding `getline()`. `fgets()` is used in multiple places, so let's rework them in a separate PR. For this one a "classic" fix would be sufficient in my opinion.
@VictorDenisov could you check my comments?

VictorDenisov:
I'll address your comments. This weekend is the contest for which I need tlf. If not this weekend I'll address your comments shortly after.

On September 30, 2022 7:18:17 PM UTC, Zoltan Csahok ***@***.***> wrote:
***@***.*** Good point regarding `getline()`. `fgets()` is used in multiple places, so let's rework them in a separate PR. For this one a "classic" fix would be sufficient in my opinion.
***@***.*** could you check my comments?
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/Tlf/tlf/pull/345#issuecomment-1263930422
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

N0NB:
I know glib has some string handling functions.  Off-hand I'm not sure what it has for input handling but it likely has something safer than the C library.

dl1jbe:
> I know glib has some string handling functions. Off-hand I'm not sure what it has for input handling but it likely has something safer than the C library.

The relevant function in GLib would be 'g_io_channel_read_line()'.
But the above mentioned 'readline()' function is quite similar to use and also safe.

zcsahok:
Perfect, thanks!

Opened #348 to follow up switching to `readline()`.

PR #349, opened 2022-10-06
Add California QSO party rules

VictorDenisov:


dl1jbe:
Are these general rules and multipliers for all station? Or do we need different ones for stations from CA or similar?

VictorDenisov:
These rules are for CA stations. We would need slightly different rules for non CA stations.

dl1jbe:
Thanks for clarification. As you see in other rule files (in ~/rules) we have mostly added a suffix to the rule file indicating who the file is for (e.g. 'arrldx_usa' vs 'arrldx_dx'). I was wondering if we should rename 'cqp' into 'cqp_ca' to make it more clear.

dl1jbe:
Thanks Victor for the changes. Files looks good.

As we may not be able to do a thorough check of the files in short time I started a discussion how to handle them (see issue #353). Lets wait two or three days for possible feedback before pulling your PR in.

VictorDenisov:
In case it may help here are the rules of the contest: https://cqp.org/Rules.html

dl1jbe:
Thanks for the pointer to the rules. 

As you may have red in the above mentioned issue, can you please add two comment lines to the start of your files with author (callsign) and creation date?

dl1jbe:
Sorry or being late. Schedule is tight at the moment.

Thanks for the work. Will pull it in now.

PR #350, opened 2022-10-07
Add NO_LEADING_ZEROS_SERIAL config

VictorDenisov:
Some contests discourage sending leading zeros. It would be nice to have it configurable.


df7cb:
That option should be documented.

dl1jbe:
And the test code in test/test_parse_logcfg.c should also be extended.

dl1jbe:
Very well Victor. For the unit test it would be simpler to just add keyword and variable to the 'bool_trues[]' array starting at line 384 in test/test_parse_logcfg.c (sorry for not giving the hint in advance).

dl1jbe:
Many thanks. Merging in now.

PR #351, opened 2022-10-07
Drop nr_qsos variable

dl1jbe:
To complete the migration to 'qso_array' for in memory storing of log entries the commit drops the 'nr_qsos' variable and replaces its usage by the actual length of the array (see macro definition of 'NR_QSOS').

Some further minor changes are:

- Adapt test code to use of 'qso_array'
-  factor out writing a backup in 'readcalls()'
- add file of logfile as parameter for 'store_qso()' and
- reuse 'store_qso()' to write any notes into the logfile.

### DIFF
--- a/src/addspot.c
+++ b/src/addspot.c
@@ -100,7 +100,7 @@ void add_local_spot(void) {
 
 /* find last qso record in qso_array, return NULL if no one found */
 static struct qso_t *find_last_qso() {
-    int i = nr_qsos;
+    int i = NR_QSOS;
     while (i > 0) {
 	struct qso_t *last_qso = g_ptr_array_index(qso_array, i-1);
 	if (!last_qso->is_comment) {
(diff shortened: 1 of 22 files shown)


dl1jbe:
As the logfile does no only contain qso data but also comment lines (notes) maybe we should adapt the naming to be more consistent.  'entry' seems to be a better naming than 'qso' in that context. 

Any thoughts?

zcsahok (approved):

zcsahok:
I see the point regarding the naming. On the other hand it will in practice contain 99% real QSOs, I assume. I would keep it as-is. "entry" is just too generic, in my view.

dl1jbe:
Ok, lets keep it for the time being. 
Background: I am thinking about a set of macros/functions to encapsulate the different aspects of handling log entries:

NR_QSOS      - as the name tell
QSO(n)         - address of record
ADD_QSO   - add a new record to the array
DROP_LAST_QSO - for deleting last entry
REPLACE_QSO  - see calledit()

It would hide the real implementation of the in-memory log and make code more readable.

Something similar could be done for log file handling. It should concentrate all knowledge about the real log file format in one place only. Generating logline, storing it to disk and re-parsing into the qso_t record should be brought together in the long run. That would make a later switch to other logfile formats quite easy.

PR #358, opened 2022-10-26
Allow unsetting boolean configs

### LINKED ISSUE #357: Make boolean configs symmetric
Currently boolean config parameters can only be set to true.
Once a parameter is set in `logcfg.dat` there is no way to set it back in the rule file.

Config handling can be easily extended in a backward compatible way:
- PARAM : sets it to true; this is the current behavior
- PARAM=true (or yes) : also sets it to true (new)
- PARAM=false (or no) : sets it to false (new)
- PARAM=none of the above : error

This change would allow us to define configs that are true by default
thus avoiding the somewhat mentally hard to grab negative logic (NOxxx like NOAUTOCQ).
Maybe we can even change the recently added NO_LEADING_ZEROS_SERIAL to LEADING_ZEROS_SERIAL, if it is not too late.

I can provide a PR for this.

zcsahok:
See #357.
Minor extension: on/off are also treated as true/false.

Man page update will follow when code is OK. (will have to update a lot of params...)

### DIFF
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -241,7 +241,7 @@ extern bool mult_side;
 extern bool countrylist_only;
 extern bool mixedmode;
 extern bool qso_once;
-extern bool noleadingzeros;
+extern bool leading_zeros_serial;
 extern bool ignoredupe;
 extern bool continentlist_only;
 extern bool debugflag;
--- a/src/main.c
+++ b/src/main.c
@@ -141,7 +141,7 @@ bool serial_section_mult = false;
 bool serial_or_section = false;	/* exchange is serial OR section, like HA-DX */
 bool serial_grid4_mult = false;
 bool qso_once = false;
-bool noleadingzeros;
+bool leading_zeros_serial;
 bool ctcomp;
 bool nob4 = false;		// allow auto b4
 bool ignoredupe = false;
@@ -626,7 +626,7 @@ static void init_variables() {
     unique_call_multi = MULT_NONE;
     generic_mult = MULT_NONE;
 
-    noleadingzeros = false;
+    leading_zeros_serial = true;
     ctcomp = false;
     resend_call = RESEND_NOT_SET;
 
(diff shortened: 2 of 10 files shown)


zcsahok:
Updated man page. Also added some non-documented parameters.

@VictorDenisov would it be OK to rename `NO_LEADING_ZEROS_SERIAL` to `LEADING_ZEROS_SERIAL` with true default value?

VictorDenisov:
Yeah. Totally. I guess now the configuration will look like
LEADING_ZEROS_SERIAL off
?

zcsahok:
Yes. Any of these statements will suppress leading zeros.
```
LEADING_ZEROS_SERIAL=off
LEADING_ZEROS_SERIAL=NO
LEADING_ZEROS_SERIAL=False
```
(the argument is case-insensitive, so `no` does the same thing too)

dl1jbe:
Code looks good and should catch all needed use cases. I think we should merge it in.

To drop keywords with negative logic (e.g. NO_BANDSWITCH_ARROWKEYS) it seems we need to document the default value of the boolean values in the man page.

It seems further desirable to have a command to show the state of the boolean controls in the running program. 

But both may be later additions.

zcsahok:
I'd keep legacy negative logic parameters. It would be more work to document such a change than the actual benefit. But at least for new configs we shall stick to positive logic.

Showing (and editing + reloading) configs is related to #233: one idea could be move all globals and their initialization out of main into a data.c (like for the tests). But this is also more for the future.

dl1jbe:
> I'd keep legacy negative logic parameters. It would be more work to document such a change than the actual benefit. But at least for new configs we shall stick to positive logic.
> 
> Showing (and editing + reloading) configs is related to #233: one idea could be to move all globals and their initialization out of main into a data.c (like for the tests). But this is also more for the future.

I agree to both. They are ideas for the long run.

Issue #332, opened 2022-05-24
Re-send exchange should respect contest specific settings


dl1jbe:
After a qso gets logged you can re-send the exchange by pressing '_'. 

At the moment it is hard-coded  to send RST and last qso number. It should respect the the contest specific exchange settings instead.

VictorDenisov:
I'm wondering if there is any work happening around this topic. S&P mode in general seems like second class citizen in tlf. Sending repeats in S&P mode requires invoking manual keying. I would like to tackle this issue and would appreciate any directions.

dl1jbe:
From your discussion let me summarize the following needs:

- Use the last QSO nr sent instead of the actual one (At the moment '_' uses actual 'qsonum -1', we should keep a copy of last sent number for that).
- Have the ability for a custom reply. I fear using 'only the last word' or similar will fall short. Best maybe a new keyword RESEND_MSG (or similar) with an useful callback if not set.

zcsahok:
One more point:

- if callsign field is empty then F3 shall also use the stored previous QSO number. (i.e. make a "full" resend instead of the potentially shorter '_')

VictorDenisov:
I'm wondering if in S&P mode as long as the call sign field is empty we need to make all macro expansions with the previous QSO values. Some people like to put parts of qso exchange in alternative macros. This approach would allow for their macros to be useful without any custom extra variables.
What do you think?

dl1jbe:
Sounds reasonable. At the moment I am not sure if it would always be good, so let us think a little bit about it.

VictorDenisov:
As I was exploring code around this change I ran into this https://github.com/Tlf/tlf/blob/master/src/sendbuf.c#L178
Could you please help me understand the purpose of these variables:
- hiscall_sent
- sentcall
What is early start?
Why are first and subsequent @ signs are handled differently?

zcsahok:
`early_started` belongs to the auto-send feature, see `autosend()` in `callinput.c`.
Auto-send is controlled by ":CHAr"  command, it always starts disabled.
The use case is (assuming it was set to 3 chars)

- we want to work K1ABC
- as soon as we type the first 3 chars (K1A) keying starts
- we finish the call input (BC) and press Enter
- now K1A is already queued (even partially sent), so the initial @ in F3 shall only be expanded to "BC". Any eventual further @ references have to be replaced with the full call as usual.

[...]

PR #360, opened 2022-11-06
Resend macros draft

VictorDenisov:
@dl1jbe @zcsahok I'm working on a fix for https://github.com/Tlf/tlf/issues/332.
The work is in progress, but I thought maybe you could take an early look and say if you approve the overall direction.
Basically I want to make sendbuf accept qso_values_t type. It's going to use those values to expand macros in the buffer.
Depending on the situation we will be passing current_qso_values or last_qso_values into the sendbuf function.
If callsign field is empty we will be passing last_qso_values otherwise we will be passing current_qso_values.

What do you think?

### DIFF
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -217,7 +217,7 @@ void *background_process(void *ptr) {
 
 			if ((qsonum <= n) && (n > 0)) {
 			    qsonum = highqsonr + 1;
-			    qsonr_to_str(qsonrstr, qsonum);
+			    qsonr_to_str(current_qso_values.qsonrstr, qsonum);
 			}
 			lan_message[0] = '\0';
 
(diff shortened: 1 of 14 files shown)


VictorDenisov:
Ignore it for now. I think I have a better way to implement it.

zcsahok:
We could maybe drop `qsonrstr` altogether and use `qso_nr` of `qso_t`. `qsonrstr` seems to be a leftover from the times when even a `sprintf` was considered a performance hog. This would also save us tracking both values in parallel.

Then macro expansion could either use values from `current_qso` or from `previous_qso` pointing to the last QSO in `qso_array`, as you suggested.

VictorDenisov:
Closing this PR in favor of https://github.com/Tlf/tlf/pull/361

PR #361, opened 2022-11-08
Resend macros

VictorDenisov:
This is another approach to solving this problem: https://github.com/Tlf/tlf/issues/332.
Compared to https://github.com/Tlf/tlf/pull/360 this PR minimally touches existing code.
It's easier to prove that this code doesn't break any existing behavior.
We can make new behavior configurable if you are not certain whether it can suit everyone.


src/sendbuf.c
```diff
@@ -234,13 +235,61 @@ void ExpandMacro(void) {
 	replace_all(buffer, BUFSIZE, "|", "");	    /* drop it */
 }
 
+void ExpandMacro_PreviousQso(void) {
```
zcsahok, line 238: it would be nice to avoid code duplication

src/sendbuf.c
```diff
@@ -234,13 +235,61 @@ void ExpandMacro(void) {
...
+    int i;
+    static char qsonroutput[5] = "";
+    static char rst_out[4] = "";
+
+    struct qso_t *prev_qso = g_ptr_array_index(qso_array, NR_QSOS - 1);
```
zcsahok, line 243: what if there's no previous QSO?

in a networked setup our previous QSO is not necessarily the one before the last on the list.

src/getexchange.c
```diff
@@ -258,7 +258,7 @@ int getexchange(void) {
 
 	    case KEY_F(2) ... KEY_F(11): {
 		/* F2...F11 - F1 = 1...10 */
-		send_standard_message(x - KEY_F(1));
+                send_standard_message(x - KEY_F(1));
```
zcsahok, line 261: no check for empty call field here?

could the check be moved to a more central place?

zcsahok (changes requested):

src/sendbuf.c
VictorDenisov: What do you recommend to look at in the case of networked setup?

src/getexchange.c
VictorDenisov: There were concerns how applicable this feature will be for general public. Overall it seems reasonable to limit when this feature kicks in.
Usually this feature is useful right after you press enter in S&P mode at the end of the line. It means you are in the callsign field. If you moved outside of call sign field it means you are doing something else. We can always expand when this feature kicks in if people request it.

src/sendbuf.c
```diff
@@ -234,13 +235,61 @@ void ExpandMacro(void) {
...
+    rst_out[1] = short_number(last_rst[1]);
+    rst_out[2] = short_number(last_rst[2]);
+    rst_out[3] = '\0';
+
+    replace_all(buffer, BUFSIZE, "[", rst_out);   /* his RST */
+
+
+    if (NULL != strstr(buffer, "#")) {
+        char *prevnr = g_strdup_printf("%03d ", prev_qso->qso_nr);
+	int leading_zeros = 0;
```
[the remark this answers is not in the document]
VictorDenisov: If we decide to use printf in both current and previous qso. What number should we use when we are dealing the current qso?

src/getexchange.c
VictorDenisov: '_' is also implemented only in callinput file and not in getexchange.

src/sendbuf.c
VictorDenisov: Maybe I'm missing something, but lan seems to be broken in master. When I enter a qso in one node the other node is seg faulting.

zcsahok: Yes, network is broken now and needs a rework. So let's for the moment just add a function like `get_previous_qso()` which simply does the same as line 243. Later we can then update it to match the network code.

Probably `qso_t` will have to be extended with a `node` field to allow selection of own QSOs.

src/sendbuf.c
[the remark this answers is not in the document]
zcsahok: Good point. Let's postpone then this (i.e. dropping of `qsonrstr` and replacing it with a correctly updated `qso_nr`) to a second stage.

zcsahok:
Re-sent serial has one leading zero less:
image

src/getexchange.c
zcsahok: True. But ideally the message keys should work the same way in all input fields. This is unfortunately not the case as there is no common handling for them (yet). In my view we shall not increase this divergence.

As an extra field there is the keyer (Ctrl-K) window where it is quite possible that the user wants to mix F-keys with typed text.

src/sendbuf.h
```diff
@@ -22,9 +22,16 @@
...
 #define SENDBUF_H
 
+typedef void (*ExpandMacro_t) (void);
+
+void ExpandMacro_CurrentQso(void);
+void ExpandMacro_PreviousQso(void);
+
 char short_number(char c);
 void sendmessage(const char *msg);
+void sendmessage_with_macro_expand(const char *msg, ExpandMacro_t expandMacro);
```
zcsahok, line 32: I think `sendxxx` could expose less of its inner workings. The user just wants to send a message using the current QSO or the previous one. Whether this is implemented using a function pointer or somehow else is internal to `sendbuf`.

VictorDenisov: I'm not sure I understand what specific changes you want to see in this code. Do you mean introducing a function called:
sendmessage_previous_qso and wrap sendmessage_with_macro_expand in it?

src/getexchange.c
VictorDenisov: I made all three fields: exchange, call input, and keyer behave the same with respect to resend keys(F1..F11, '_'). I'm happy to work on combining all of them into a single function, but maybe in a different PR? They look very different and it will take time to make sure that nothing is missed.

src/sendbuf.h
zcsahok: Yes, practically.
That would also save us from mocking `ExpandMacro_xxx` in test sources not actually related to message sending.

src/keyer.c
```diff
@@ -246,50 +246,22 @@ void keyer(void) {
...
+		    } else {
+			send_standard_message(x - KEY_F(1));
+		    }
 		    break;
 		}
-		case KEY_F(12): {
-		    send_keyer_message(11);	/* F12 */
+
+		// Underscore, confirm last exchange.
+		case '_': {
```
zcsahok, line 259: This is kind of corner case here, but let's assume that no one wants to send an underscore via RTTY or a digital mode.
After all this is contest logger, not a general logger program.

src/sendbuf.c
```diff
@@ -234,13 +236,46 @@ void ExpandMacro(void) {
 	replace_all(buffer, BUFSIZE, "|", "");	    /* drop it */
 }
 
+struct qso_t * get_previous_qso() {
+    static struct qso_t empty_qso;
```
zcsahok, line 240: `empty_qso` could be also initialized as
```
static struct qso_t empty_qso = {
    .call = "",
    .qso_nr = 0
};

```

zcsahok:
Re-sent serial looks OK now.

src/getexchange.c
zcsahok: Yes, that will require another PR for sure. There is still surely a lot to be fixed or improved.

src/keyer.c
VictorDenisov: Does it mean you are ok with this piece of code or it needs to be changed somehow?

zcsahok: I'm ok with it. Was just thinking loud.

VictorDenisov:
I think I addressed all comments.

zcsahok:
Thanks, will make a final test tomorrow.

Two minor points:
Please check the man page if it needs an update.
Apply source formatting to the changed files: we use astyle with the setting in `tools/astylerc`

`sendbuf.c` has a merge conflict now due to #358, but I can merge it manually, no problem.

src/callinput.c
```diff
@@ -793,7 +803,7 @@ int callinput(void) {
 
 	    // Underscore, confirm last exchange.
 	    case '_': {
-		prev_qso();
+		send_standard_message_prev_qso(2);
```
zcsahok, line 806: In S&P mode SP_TU_MSG shall be sent.

In order to better see the difference I used this config
```
F3=@ [ #
S&P_TU_MSG=[ # XX
```
Working A1BB and the re-sending the exchange:
image

zcsahok:
Just for the record, there two minor issues related to the re-send functionality in general.
For normal contest operation these are not really relevant, but nevertheless shall be fixed to have a consistent behavior.

* an eventual correction of RST of the previous QSO is not honored: the value
of `last_rst` global is sent instead of the actual one from the last QSO.

* confirm call (`=`) does not respect correction of call. Same story as above:
global `lastcall` is used instead of a QSO field. In addition, its value
is initially empty so after start-up last call can't be confirmed.

Both globals have to be dropped and this will require a separate PR.

VictorDenisov:
I didn't quite understand the comment about confirm call. But I fixed rst field.

zcsahok:
Thanks for the RST fix! Please also remove the global variable `last_rst` as it is now written only.

S&P works now OK.

For confirming call I meant the message sent on pressing `=`. It is useful for CQ mode only and shall send the previous call. But let's check this separately.

zcsahok (approved):

zcsahok:
Thanks!

Merged in master to resolve the merge conflict.

Issue #362, opened 2022-11-25
Using HAMLIB_KEYER makes other hamlib calls fail


jothan:
I am using the latest git tlf. When adjusting the speed up and down, some hamlib calls fail sporadically (blue status bar on the bottom). I am using hamlib via rigctld (model 2).

Looking at the code, keyer calls are made on the background thread while other rig_* calls are made on the main thread. I suspect all hamlib calls should be protected by a mutex to avoid interleaving requests.

Let me know what you think. I can make a pull request with that change if you want.

73,
Jonathan VA2XZA

zcsahok:
Thanks for the report! Most likely the cause is what you described. I can't test it as I have no rig capable of CAT keying.
@df7cb could you check this?

A PR would be surely welcome.

df7cb:
Oh that explains a regression I'm seeing now, CW keying and speed adjustment works, but frequency readout stops working after a short while. Since it worked until ~2 weeks ago, I attributed it to a hamlib upgrade I did around that time, and/or RFI, but of course the problem might well be inside tlf, and only exposed by the hamlib upgrade.

I wasn't aware of threading in tlf, so I didn't pay attention to it while coding the HAMLIB_KEYER stuff. A PR is of course welcome, I can't promise when I'll have time to look at it myself.

PR #363, opened 2022-11-25
Synchronize all calls to hamlib with a mutex.

jothan:
With HAMLIB_KEYER, hamlib calls are made from two different threads, they need to be synchronized.

This is for fixing #362, I have not been able to reproduce it after this patch.

### DIFF
--- a/src/hamlib_keyer.c
+++ b/src/hamlib_keyer.c
@@ -24,11 +24,13 @@
 #include "hamlib_keyer.h"
 
 bool rig_has_send_morse() {
+   // SAFETY: should not be written after init
    return (my_rig->caps->send_morse != NULL);
 }
 
 bool rig_has_stop_morse() {
 #if HAMLIB_VERSION >= 400
+    // SAFETY: should not be written after init
     return (my_rig->caps->stop_morse != NULL);
 #else
     return false;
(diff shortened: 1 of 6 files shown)


src/hamlib_keyer.c
```diff
@@ -24,11 +24,13 @@
 #include "hamlib_keyer.h"
 
 bool rig_has_send_morse() {
+   // SAFETY: should not be written after init
    return (my_rig->caps->send_morse != NULL);
 }
 
 bool rig_has_stop_morse() {
 #if HAMLIB_VERSION >= 400
+    // SAFETY: should not be written after init
```
zcsahok, line 33: I'm not sure: what should not be written?

zcsahok (commented):
Looks OK.

src/hamlib_keyer.c
jothan: I am referencing the my_rig->caps field.

zcsahok: I see. Don't we only read it?

jothan: Yes, but technically, some other thread could update it while it is being read resulting in undefined behavior.

zcsahok: True. But we as a hamlib client shall not mess around rig caps in general. Whether hamlib updates them at runtime (highly unlikely) that is out of our control anyway.
Wouldn't a note reminding that these fields are evaluated on init be sufficient? Or should these be also read and cached in `init_tlf_rig()` a bit like `rigptt`?

jothan: I think having global flags for has_send_morse and has_stop_morse out of hamlib's reach is the best option.

jothan: I have pushed a patch with flags that are set on init.

df7cb:
Hmm, frequency readout is still broken here, and possibly even worse than without this patch. Instead of using the mutex, should all hamlib interaction be moved to a single thread instead?

Also, it's failing silently here, no error message at the bottom.

df7cb:
I downgraded hamlib to 4.3 and the problems are gone, even without recompiling tlf. Haven't had a single problem in CQWW today.

Not sure if that means I'm seeing a different problem.

zcsahok:
No issues here with hamlib 4.0 either, using cwdaemon keying.

zcsahok:
Merging in, thanks.
Let's keep an eye on hamlib 4.5.

PR #364, opened 2022-11-28
Update workflow for Node.js 16

zcsahok:
Due to github moving to Node.js 16 some actions has to be updated. TLF workflow had just warning, but mine did actually fail with

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2

So switching to checkout v3.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -16,7 +16,7 @@ jobs:
       AM_COLOR_TESTS: always
 
     steps:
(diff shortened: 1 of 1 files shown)


zcsahok:
The warning is now gone.

dl1jbe:
+1

PR #365, opened 2022-11-28
Fix critical Clang warnings

dl1jbe:
Coming Clang-16 converts some former warnings into strict errors. The patch provides fixes for the following Clang-16 settings:

# clang-16 defaults
-Werror=implicit-function-declaration
-Werror=implicit-int
-Werror=incompatible-function-pointer-types


zcsahok:
Fine with me.

PR #366, opened 2022-11-28
Fix ARRL Sweepstakes

zcsahok:
Fixing the issue reported by @N0NB: it was not possible to configure the sent exchange for Cabrillo as TLF was insisting on using the serial number.

Caused by the internal definition of SS setting `exchange_serial = true`.

The meaning of the rarely used `exchange_serial` flag probably has drifted over time from "exchange contains (or begins with?) a serial number" to "both sent and received exchanges are plain serial numbers". Updated man page to reflect this.

Also moved the globally defined, but only locally used flag to writecabrillo.c. Didn't make it static for testing sake.

### DIFF
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -222,7 +222,6 @@ extern char exchange_list[40];
 extern char rttyoutput[];
 extern char spot_ptr[MAX_SPOTS][82];
 extern char lastmsg[];
-extern char exchange[40];
 #ifdef HAVE_LIBXMLRPC
 extern char fldigi_url[50];
 #endif
--- a/src/main.c
+++ b/src/main.c
@@ -186,7 +186,6 @@ bool demode = false;		/* send DE  before s&p call  */
 
 int announcefilter = FILTER_ANN; /*  filter cluster  announcements */
 bool showscore_flag = false;	/* show  score window */
-char exchange[40];
 int defer_store = 0;
 mystation_t my;			/* all info about me */
 
--- a/src/setcontest.c
+++ b/src/setcontest.c
@@ -219,7 +219,6 @@ contest_config_t config_arrldx_dx = {
 contest_config_t config_arrl_ss = {
     .id = ARRL_SS,
     .name = "ARRL_SS",
-    .exchange_serial = true,
     .points = {
 	.type = FIXED,
 	.point = 2,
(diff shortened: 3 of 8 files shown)


zcsahok:
Minor note: we can't separately define sent and received exchange formats. This could make a difference in ADIF generation for contests with say fixed sent, but serial received exchanges as ADIF has (requires?) different tags for serial and text exchanges.

dl1jbe:
> Fixing the issue reported by @N0NB: it was not possible to configure the sent exchange for Cabrillo as TLF was insisting on using the serial number.
> 
> Caused by the internal definition of SS setting `exchange_serial = true`.

Just checked the oldest version I have access to (TLF-0.9.10). If contest was set to ARRL_SS 'setcontest()' function did set 'exchange_serial' to 1. I assume I just copied it over while moving to the config_t structs.

Maybe ARRL SS did use serial numbers around that time (2003). In actual contest rules serial number exchange is not mentioned anywhere.

> 
> The meaning of the rarely used `exchange_serial` flag probably has drifted over time from "exchange contains (or begins with?) a serial number" to "both sent and received exchanges are plain serial numbers". Updated man page to reflect this.
> 
No idea here.

> Also moved the globally defined, but only locally used flag to writecabrillo.c. Didn't make it static for testing sake.

+1 

Feel free to merge in. I would suggest adding comment 2 as an issue.

N0NB:
SS has long used serial numbers but always in conjunction with other information as a required part of the exchange.  In other words, serial numbers have always been used (since well before the use of computer logging) but not as the sole part of the exchange.  For as long as I've participated (1993 was the first year I did so) it has been the five part exchange of SN, Prec, call, Check, Section.  Other than the addition of sections and renaming some sections, it doesn't change much.

N0NB:
On another note, for SS no more than four digits are required for the serial number.  The all-time record is still less than 3000 QSOs for the 24 hour operating period.  Owing to working stations only once during either the CW or phone weekends, keeps the QSO tally low.

dl1jbe:
Thanks for the correction Nate. Somehow I skipped the serial number while reading the rules.

zcsahok:
The ADIF story is quite complex as it is not clearly specified what shall be written into SRX and what into SRX_STRING. E.g. if exchange contains serial+section (UBA DX), then shall the serial be written into SRX? Probably not, but it's not clarified. Another example: in-country stations send section and out-of-country stations send serial, but all can work all (e.g. HA DX). Then in theory it would require a per-QSO decision in which field to write the exchange.

So let's leave it until a concrete error report.
At least for the serial-only case we are fine.

PR #368, opened 2022-11-29
Fix pre-v4 Hamlib support

### LINKED ISSUE #367: Hamlib version depedency?
For some reason my repo's actions use Ubuntu 20 for CI build and they fail since #363.
Ubuntu 20 has Hamlib 3.3 and it provides no `rig_stop_morse` function.
Before #363 there was a warning about the missing function but the actual access was optimized away as `rig_has_stop_morse()` returned a contant false. So compilation was successful.
Now this optimization is not possible thus leading to a linking error.

We have two options -

1. Require hamlib >= 4.0
2. Guard `rig_stop_morse` access with `if HAMLIB_VERSION >= 400`

What would be your preference?

If 1) is preferred then all Hamlib version guards could be dropped.

zcsahok:
Added CI build for Ubuntu 20 to test with Hamlib 3.3

Added guard for `rig_stop_morse()`

Renamed Hamlib morse flags to avoid any clashes with `rig_stop_morse`. Applied astyle.

Fixes #367


dl1jbe:
Commit Number 2 looks good and will fix the problem.

I suggest to drop commit #3 and drop 'rigsendmorse' and 'rigstopmorse' at all. It makes no sense to have a private copy of public available information as additional global variables. We could instead check the rig capabilities direct or use a macro for it. There is nearly no overhead in doing so.

jothan:
> Commit Number 2 looks good and will fix the problem.
> 
> I suggest to drop commit #3 and drop 'rigsendmorse' and 'rigstopmorse' at all. It makes no sense to have a private copy of public available information as additional global variables. We could instead check the rig capabilities direct or use a macro for it. There is nearly no overhead in doing so.

Copying in the variables ensures no contention or potential synchronization issue with hamlib. Uncontended mutexes are fast, but it can be bypassed by grabbing the info on initilazation.

dl1jbe:
> > Commit Number 2 looks good and will fix the problem.
> > I suggest to drop commit #3 and drop 'rigsendmorse' and 'rigstopmorse' at all. It makes no sense to have a private copy of public available information as additional global variables. We could instead check the rig capabilities direct or use a macro for it. There is nearly no overhead in doing so.
> 
> Copying in the variables ensures no contention or potential synchronization issue with hamlib. Uncontended mutexes are fast, but it can be bypassed by grabbing the info on initialization.

Sri for being late.

These rig capabilities are constant data. Reading the data value should therefore make no problem wrt threading. The thread related problems came from calling Hamlib functions while other hamlib functions were running. I do not see the problem in our case. 

I would be glad to have copies of the configuration in TLF if we could make them read only, so that no one will change them by accident. If we rely on the data in rig capabilities we do not have to guard them against unintentional change.

But anyway it is just my 2 c.

zcsahok:
To address the above concerns made the flags private to sendqrg and added access functions. This ensures read-only accesses, avoiding any potential problems with threading. The flags themselves are evaluated at init time.

dl1jbe:
Looks like the best idea. +1

zcsahok:
Are we good to merge it in?

dl1jbe:
I think so. Sure.

Issue #344, opened 2022-09-01
Enhance _Escape_ to stop message TX on first press


N0NB:
In Log mode, Pressing _Escape_ while a message is being transmitted results in two actions happening--the TX is stopped and the field the cursor is in causes one of two things to happen.  If the cursor is in the Call field, both the Call and Exchange fields are cleared (bad news when the objective is to just stop the TX).  If the cursor is in the Exchange field then any entered text is erased and the cursor moved to the Call field.

In S&P mode if the cursor is in the Exchange field and there is text, _Escape_ stops the TX and clears the text but keeps the cursor in the Exchange field.  If the cursor is in the Call field the TX is stopped and all fields are cleared.

The actions appear to be the same whether the operating mode is CW or SSB.

Perhaps this is the desired action for _Escape_, or perhaps it mimics Tr Log (I don't know as I've never used Tr).

Ideally, I would like _Escape_ to only stop the TX if a message is being sent and not clear any text on first press.  However, if there is no message being sent, then if should clear the fields as it does now.  If this is at odds with the expectations of others that I'd propose this to be a configurable option.

On a couple of occasions the past weekend I had to ask for repeats because when I only intended to stop the TX I also had at least one input field cleared.

df7cb:
Fwiw there's already a note in the manpage about that:

```
       Escape Stop CW transmission, clears characters, returns to call input field, keyboard off (universal undo).
              Note:  Some  changes  are being considered to modify the behavior of Escape slightly.  The intent will be to have Escape
              only stop sending on first press and not clear any entered information.  It is planned that Escape will stop  the  play‐
              back of the voice keyer as well.

              As  an  example,  characters have been entered in both the call input and exchange fields, the cursor is in the exchange
              field, and the transmission of a CW message is in progress.

              The first press of Escape will stop the CW transmission and clear the exchange field and  position  the  cursor  to  the
              leftmost position of the exchange field.

              The second press of Escape will move the cursor to the right of the last character in the call input field.

              The third press of Escape will clear the call input field.
```

(4ffdce8 by you)

I think changing that would make sense, the current behavior is both overzelaous (stop and clear) and too timid (3 ESC needed to clear everything). Perhaps being smarter and still requiring less ESC hits is possible.

N0NB:
I know the man page does not match the behavior.  :)

If I recall, it was discussed on the mailing list but was never implemented, or at least HEAD is not matching the man page.  I wanted to get this on the list of issues for some additional discussion.

[...]

PR #369, opened 2022-11-30
Escape stop tx only

dl1jbe:
A discussed in Issue #344 this PR add the following features:

- New keyword 'ESC_STOP_TX_ONLY' modifies ESCAPE key behavior, so that it only stops an ongoing CW or voice  message. The gradual wipe off of information from exchange and  call input field are blocked.
- Add two new keys to wipe out (and restore afterward) the actual field the cursor is in - Ctrl-W - or the whole information line - Ctrl-U.


N0NB:
I ran from this branch during last weekend's ARRL 160m 'test.  It worked as expected so I think you can merge this, Tom.

dl1jbe:
Thanks or the test Nate. Will merge it in now.
At the moment I am experimenting to get the UNDO feature also used if you still use ESCAPE to wipe out the input fields stepwise.

PR #370, opened 2022-12-08
Fix leading zeros configuration in cqp rules

VictorDenisov:


dl1jbe:
Thanks for the correction

PR #371, opened 2022-12-18
prev_qso->call is NULL when the last entry is a comment.

jothan:
This prevents a segfault.


dl1jbe:
Good catch. 
But would it not be better in that case to search back in history to the last qso recorded before the comment (e.g. in 'get_previous_qso()').

jothan:
@dl1jbe I updated the patch to fix get_previous_qso instead.

dl1jbe:
Looks good. Thanks for the patch. will apply now.

PR #374, opened 2023-01-24
Fix mult count when working an existing one on new band

zcsahok:
When working an existing multiplier on a new band then it gets counted twice. Once in the `for` loop of `remember_multi()` and then at the end of the function when doing the actual update.

Extended the tests to check `multscore` too. The issue is seen in `remember_mult_same_2x` and `remember_mult_same_2x_newband`.


dl1jbe:
Correct. Old code did it only once - in the first loop if 'found' or in the second part if '!found'

+1

PR #375, opened 2023-01-24
Show per-band multipliers if they are used

zcsahok:
If per-band multipliers are configured then show them along with QSO numbers.

The condition for showing multipliers is now the same as the one used in `get_nr_of_mults()` to return `totalmults` (line 182). Made the mult-showing if's exclusive to avoid any future issues.


dl1jbe:
Very good. Now the displayed values are consistent with each other.

+1

PR #376, opened 2023-01-28
Use current time to calculate QSO rate

zcsahok:
Instead of using only the time data of QSOs use the full timestamp to determine how many minutes ago we had the 10th QSO on the current band.

This avoids showing bogus rate values when opening an older log or in a contest longer than 24 hours.

### DIFF
--- a/src/last10.c
+++ b/src/last10.c
@@ -19,6 +19,7 @@
 /* ------------------------------------------------------------
  *     last 10 - return time (in mins) for last 10 QSOs on
  *		 actual band
+ *		 in case of not enough QSOs -1 is returned
  *--------------------------------------------------------------*/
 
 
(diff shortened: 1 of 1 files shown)


src/last10.c
```diff
@@ -54,18 +52,9 @@ int last10(void) {
 	}
     }
```
dl1jbe, line 54: From the description in the header we want to return -1 if less than 10 QSOs are found on the band. So should we better check for 'qsocount <10' instead of 'index <0'?

src/last10.c
```diff
@@ -54,18 +52,9 @@ int last10(void) {
...
-
-    minsnow = get_minutes();
-
-    if ((minsnow - minsbefore) <= 0)
-	minsnow += 1440;
-
-    span = minsnow - minsbefore;
-
-    return span;
+    return (get_time() - qso->timestamp) / 60;  /* elapsed time in minutes */
```
dl1jbe, line 59: Good idea. I like it. Not only solves it the >24h problem but is also more efficient.

dl1jbe (approved):

src/last10.c
dl1jbe: You are right. I misread the code. Keep it as is.

PR #377, opened 2023-02-06
Fix parsing logline for FOCMARATHON

dl1jbe:
Add missing exchange_width field to contest definition in focm.c


dl1jbe:
Bug fix confirmed by reporter.

PR #378, opened 2023-02-18
Fix/typos

dforsi:
This PR fixes some small errors in the documentation and in comments.


zcsahok:
Thanks.
There is also at least one in `share/logcfg.dat`
```
# or for dots+calss
```

dforsi:
There were two more words in `share/logcfg.dat`.

dl1jbe:
Last correction should be 'calls' not 'class' ;-).

dforsi:
Ops :)
Now I squashed the last commit about `share/logcfg.dat` in the first commit.

dl1jbe:
So, seems no more hints at the moment. Merging it in.
Thanks for the PR.

PR #380, opened 2023-02-20
Hand over correct parameter for scoring

dl1jbe:
Second part of fixes.

Scoring functions need a 'struct qso_t *' as parameter and should evaluate based on that data.

### DIFF
--- a/src/focm.c
+++ b/src/focm.c
@@ -48,8 +48,8 @@ int cntry;
 int cont;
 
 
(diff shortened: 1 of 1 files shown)


dl1jbe:
P.S. Found thanks to testing PR #379. So it showed itself already as useful ;-)

dl1jbe:
No comments, merging it in

PR #382, opened 2023-03-11
Fix implicit function declarations with clang

dl1jbe:
Will be counted as errors starting with CLANG-16


zcsahok:
What are the offending functions? Maybe we could just replace them with the "correct" ones instead of adding a magic define.

dl1jbe:
Sorry, missed to give that information.
It's 'timegm()' in both cases. See man page.

zcsahok:
I see. Indeed GNU libc manual (https://www.gnu.org/software/libc/manual/html_node/Broken_002ddown-Time.html) says:

 _timegm is rather rare. For the most portable conversion from a UTC broken-down time to a simple time, set the TZ environment variable to UTC, call mktime, then set TZ back._ 

We have `timegm` in `log_utils.c` and `writecabrillo.c` too. Couldn't we just set `_DEFAULT_SOURCE` (and `_XOPEN_SOURCE`)  globally in `src/Makefile.am`? (When compiling with gcc-10.2.1 both are set automagically.)

dl1jbe:
> We have timegm in log_utils.c and writecabrillo.c too. Couldn't we just set _DEFAULT_SOURCE (and _XOPEN_SOURCE) globally in src/Makefile.am? 

Just looked why we get no problems from clang reported there: We have _GNU_SOURCE already defined in that files. 'man feature_test_macros' says about it:

"Since  glibc  2.19,  defining _GNU_SOURCE also has the effect of implicitly defining _DEFAULT_SOURCE.  
Before glibc 2.20,  defining  _GNU_SOURCE  also  had  the  effect  of implicitly defining  _BSD_SOURCE and _SVID_SOURCE."

One of the problems is that the meaning and use of the macros is slowly changing.

> (When compiling with gcc-10.2.1 both are set automagically.)

Unluckily that seems to be handled differently between different compilers. 

Wrt adding it to src/Makefile.am. It sure needs some test but i am open for it. But maybe we should add _GNU_SOURCE instead.

zcsahok:
Defining say _GNU_SOURCE globally would be more manageable in my opinion than various defines scattered around in sources files. Provided of course that this covers all cases.

zcsahok:
Thanks!

zcsahok (approved):

PR #384, opened 2023-03-19
Test MWC rules

zcsahok:
Added test for MWC rules.

Updated `run_tlf.py` to clean up `__pycache__` directory.


zcsahok:
Oops, there is no plugin for Ubuntu 20, have to somehow disable the test.

dl1jbe:
Is there a missing -enable-python-plugin in the ci-built-ubuntu-20.yml?

zcsahok:
Ubuntu 20 has Python 3.8 version, plugin code needs 3.10. The build has been added to detect any issues with legacy systems, but is compiles without plugin support.

zcsahok:
Now tests with a `*.py` in rules dir are skipped no plugins are supported.
The only thing is that one does not see from the action output which rules were actually tested.

dl1jbe:
As I read configure.ac it looks like 3.9 is the minimum required version for the plugin, or do I miss something?
It is a little bit weird that we require different minimum python versions for running the rules_test and for checking contest which use the plugin.

The automatic skip of *.py tests looks good. The tested rules should be readable from test/run_rules.log.

zcsahok:
Maybe it was 3.8 having some issues, I don't remember now  exactly. So, at least with 3.9 all works fine (probably I was mixing this up with 3.10, sri).
But for the testing script itself 3.8 is enough. The separate check has been added explicitly to support Ubuntu 20 build.

dl1jbe:
As the python plugin is now merged into master we should adapt our README.md to reflect the new configure switch and the Python and pexpect requirements. Shall I do?

zcsahok:
Yes, please.

PR #388, opened 2023-04-04
Add ksqp rules

VictorDenisov:


test/rules/ksqp/ksqp.log
N0NB, line 1: I will note that while this might not matter for testing, any call beginning with A requires a two letter prefix as the USA is only allotted AAA through ALZ in the international prefix table.

VictorDenisov: Hah, fair enough. I didn't realize it. I guess since it's only a test it won't matter, but point taken.

dl1jbe:
Hmm. Do we really need that '.bmdata.dat' file? Please drop if not.

VictorDenisov:
done

N0NB:
Perhaps it would be better for all references to N0NB and N0N to be replaced by N0CALL.  I would be very surprised to find myself working myself!

N0NB:
Thanks, Victor.

dl1jbe:
Seems stable now. Merging it in. 
Thanks, Victor.

PR #389, opened 2023-04-13
Changed MAX_SECTION_LENGTH to 5.

alandove:
This fixes the bug we discussed on the email list, which prevented exchanges in the New England QSO Party from being logged correctly when using a mult list. It seems to work fine for me with this one change.


alandove:
How do I go about creating the test files you mentioned, Zoli? I've looked at #388 but could use a little more guidance.

zcsahok:
The goal is to set up a minimal environment to be able to check that scoring/mults are applied correctly.
A short how-to:

- create the subdirs test/rules/neqp and test/rules/neqp/rules
- create a minimal logcfg.dat in test/rules/neqp (take the one from test/rules/ksqp)
- copy or link relevant rule file(s) from rules/neqp/ to test/rules/neqp/rules; also copy/link any other relevant file
- while in test/rules/neqp execute the locally compiled TLF as `../../../src/tlf -nrv`
- enter a couple of typical QSOs or provide a log file from some other source. try to somewhat obfuscate the callsigns. check that scores and mults are correct.
- re-executing TLF shall give no warnings
- check in logcfg.dat, the log file, any relevant files from there and from the rules dir

alandove:
Tests added and done. Thanks for the tutorial.

zcsahok:
Looks good.
Please provide attribution in rule file to be able find out the knowledgeable person in case of issues. See the header part of rules/ksqp/ksqp.

alandove:
Added attribution.

zcsahok:
Thanks!

Issue #286, opened 2021-11-01
Add keyword to disable rig mode setting 


dl1jbe:
By Onno VK6FLAB on the mailing list

For some use cases it is preferable to run contest mode independent of rig mode, e.g. keeping the rig in UBS while runiing a CW contest.

As discussed on the ML there should be a keyword which inhibits all mode settings via hamlib from TLF.

darvark:
Hi,  I would like to start working on that ticket.

zcsahok:
Here is the original mail for the record
https://lists.nongnu.org/archive/html/tlf-devel/2021-10/msg00009.html

TLF currently synchronizes with the rig practically only one way,  TLF->rig:

- user switches mode with :CW or :SSB (or :DIGI). `set_outfreq(SETxxMODE)` is called in `changepars.c` and then processed by `gettxinfo.c` in the background.
- in case of a band switch the rig's mode is updated in ` handle_trx_bandswitch()`.

There is no handling of the rig->TLF direction: mode changes on the rig are not reflected on TLF side. (could be even a bug)

As far as I understood, the idea is to totally decouple TLF's mode (`trxmode`) from the actual mode of the rig. Looks like suppressing execution of SETxxMODE operations and skipping `handle_trx_bandswitch()` would be enough.

Any thoughts?

darvark:
Thank you, sounds clear. 

In that case (synchronization tlf->rig) tlf knows that we changed freq/band on rig directly? I thought tlf synchronizes those informations.

zcsahok:
TLF polls the rig and if a band change is detected, then it sets the operating mode according to `trxmode` (so it's TLF->rig). There is no rig->TLF synchronization of the mode.

A quick check to see how it works:

1. start TLF in CW mode (default) on 40m
2. rig will be set to CW by TLF
3. staying on 40m change mode to LSB
4. nothing changes on TLF side
5. now switch on the rig to 80m, still in LSB
6. TLF will show the actual frequency, but will instruct the rig to switch to CW

zcsahok:
IMO before implementing disabling mode setting we shall improve the mode synchronization: in the step 4 above TLF should switch to SSB.
In general mode setting shall behave as frequency control: rig updated if it's changed on TLF side and TLF takes over the setting if changed on the rig.
This makes working in a mixed contest easier: no need to manually change mode using :SSB/:CW, a key press on rig suffices.

darvark:
[...]

PR #397, opened 2023-07-04
#286 Add keyword to disable rig mode setting

darvark:
Draft for changes adding configuration param to disable rig mode setting.  Disscussion is available : [https://github.com/Tlf/tlf/issues/286](https://github.com/Tlf/tlf/issues/286)

Not yet tested with any rig. Please ignore first commits to this branch for S&P mode. I'll clean this later


darvark:
Ok tested with FTdx3000, os: Elementary OS 7.
Test case: Rig setup for bands: 40,20,15m to SSB, for 80m for CW. FOLLOW_MODE enabled in logcfg.dat, tlf started on 80m with CW mode. Switching band on rig will cause to switch modulation on rig to CW. 
In case when FOLLOW_MODE is disabled, then changing bands will not cause switching mode on rig to CW, just new band which is visible in tlf.

Issue #348, opened 2022-10-05
Review usage of fgets in file reading


zcsahok:
Follow-up from #345:

As the problem with long lines pops up from time to time, maybe we should check all `fgets()` invocations for truncation of input due to limit buffer size.
There is another function `getline()` which seems better fit this use case.

dl1jbe:
Scanning thru the code there are quite some places where 'fgets()' is used. 

For start it may be best to begin with user provided files (multipliers, config.dat, rules file, initial_exchange, maybe cabrillo.fmt).

darvark:
I had some free time during my vacations and I've prepared code change that contains replacement of all fgets with getline function. Tests passes, tlf seems to work without problem. Soon I'll add pull-request

zcsahok:
Fixed by #403

PR #398, opened 2023-07-09
#348 Review usage of fgets in file reading

darvark:
Removed all occurrences of fgets and replaced with getline.
Additionally code re-formated with astyle


darvark:
Hi, can someone take a look to this pull request? Thanks

airween:
Seems you have replaced the fixed length char arrays by a pointer - is there any reason why is this better? You allocate and free the necessary memory, but never check the return of `calloc()`. Wouldn't it be better/more convenient to use arrays?

airween:
> as far as I read using pointer is faster than array

Hmm... this is new for me :)

> as a data structure it's the same.

Yes, exactly.

Moreover in terms of its use in these functions, there is no real difference.

> I agree that I forgot to check if calloc return correct address. Will add this now

I suggest you forget the pointers and dynamic memory allocations, and use static arrays. I think that's more than enough here.

darvark:
> > as far as I read using pointer is faster than array
> 
> Hmm... this is new for me :)
> 
Sorry maybe I misunderstood article. It's faster in some operations on strings. Sorry.

Hmm, ok I can try to use arrays. 

Thank you

darvark:
@airween sorry but looks like getline() will not work with static arrays. Probably due to internal mechanism reallocating memory if buffer is full. At least I couldn't find way how to make it working.

airween:
> @airween sorry but looks like getline() will not work with static arrays. Probably due to internal mechanism reallocating memory if buffer is full. At least I couldn't find way how to make it working.

Oh, I'm really sorry - now I see, and the reason is:

```
man 3 getline

...
If the buffer is not large enough to hold the line, getline() resizes it with
realloc(3), updating *lineptr and *n as necessary.
```

This won't work with static arrays.

Regarding the [original issue](https://github.com/Tlf/tlf/issues/348#issue-1397270005), the aim is to avoid the random string truncating, so we can't avoid to use the pointers.

The good news is `getline()` allocates the necessary memory if you pass a null pointer **AND** the length is 0, so you don't need to allocate memory, but you have to check the return values after calling it (`bufsize > 0`), and also the `errno` variable (`errno != ENOMEM`). See the manpage of `getline`. Eg. you need to change the bufflen values to 0, eg. [here](https://github.com/Tlf/tlf/pull/398/commits/717e2de537b8800c60e354b7ac98ee0ef003f198#diff-4db1de1ab4a1940544d152b06f56a1743acfebf6472b0cea394b8ad08495756aL334):
```C
    size_t s_inputbuffer_len = 0;
```

And you have to check the value of buffer, and if it's not `NULL` (possibly) then you have to free it - for example in [this loop](https://github.com/Tlf/tlf/pull/398/commits/717e2de537b8800c60e354b7ac98ee0ef003f198#diff-4db1de1ab4a1940544d152b06f56a1743acfebf6472b0cea394b8ad08495756aL363-L365) you need to check the result, and [this removed](https://github.com/Tlf/tlf/pull/398/commits/717e2de537b8800c60e354b7ac98ee0ef003f198#diff-4db1de1ab4a1940544d152b06f56a1743acfebf6472b0cea394b8ad08495756aL377) `free()` is required:

```C
    if (s_inputbuffer != NULL) {
        free(s_inputbuffer);
    }
```

Really sorry again for my mistake.

darvark:
Nothing happened. This was great lesson for me and occasion to dig deeper into documentation. 
Sure, I'll add those checks and clean-up.

src/checklogfile.c
```diff
@@ -115,165 +122,181 @@ int repair_log(char *filename) {
 
 
 int checklogfile_new(char *filename) {
-
     int lineno;
     int tooshort;
-    char buffer[160];
+	int read;
```
airween, line 127: Indentation is wrong. See the line 129 too.

src/addmult.c
```diff
@@ -336,57 +330,62 @@ void add_mult_line(char *line) {
...
+
+                /* drop comments starting with '#' and empty lines */
+                if (*s_inputbuffer == '#' || *s_inputbuffer == '\0') {
+                    continue;
+                }
+
+                add_mult_line(s_inputbuffer);
+            }
+            else {
+                perror("RuntimeError: ");
```
airween, line 375: This (and the other uses of `perror()`) is not too informative. If the (un)expected error occurs, you will get a message:

```
RuntimeError: 
```

You should consider to use `fprintf(stderr, "RuntimeError: %s:%ld", __FILE__, __LINE__)`

airween:
@darvark: thanks for the additional modifications. Please see my comments above.

src/addmult.c
```diff
@@ -336,57 +330,62 @@ void add_mult_line(char *line) {
...
+                exit(EXIT_FAILURE);
+            }
     }
 
     fclose(cfp);
-
+    if (s_inputbuffer_len > 0)
```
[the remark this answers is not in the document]
darvark: ahh my mistake, vscode automatically suggested that variable. Fixed in in comming commit

src/addmult.c
darvark: According to documentation message will be: for code example:
`int main()
{
    perror( NULL );
    FILE *fp = fopen( "nonexisting.file", "rb" );
    if( !fp )
         perror( "RuntimeError: " );
    else
         fclose( fp );
   
    return 0;
}`
 "RuntimeError: No such file or directory" but if fprintf is a standard I can change it. Eventually add__FILE__ and __LINE__ to perror message.

darvark:
> @darvark: thanks for the additional modifications. Please see my comments above.

Wondering if it wouldn't be too much to add checking if `errno != ENOMEM` to codition 
`if (inputbuffer_len > 0)`. Theoretically getline still can read some amount of data before exceed memory so input buffer will be non zero.

tools/astylerc
```diff
@@ -4,7 +4,7 @@
 style=attach
 
 # Indent with a width of 4 spaces. Use 8 space tabs if possible
-indent=force-tab-x=8
+# indent=force-tab-x=8
```
zcsahok, line 7: Please do not change astylerc.

src/addmult.c
```diff
@@ -336,57 +330,63 @@ void add_mult_line(char *line) {
...
-
-	/* drop comments starting with '#' and empty lines */
-	if (*s_inputbuffer == '#' || *s_inputbuffer == '\0') {
-	    continue;
-	}
-
-	add_mult_line(s_inputbuffer);
-
+    while ((read = getline(&s_inputbuffer, &s_inputbuffer_len, cfp)) != -1) {
+            if (s_inputbuffer_len > 0) {
```
zcsahok, line 363: Instead of moving the good case to an extra indented block
```
if (all_is_ok==true) {
  // the good case
   step1();
   step2();
   ...
} else {
  do_error_handling();
}
```

consider  bailing-out early and keep the good case minimally indented:
```
if (all_is_ok == false) {
   do_error_handling();
   break;  // or return or ...
}

// the good case
step1();
step2();
```

test/rules/cqww/rules/cqww
zcsahok, line 1: What happened to these files?

zcsahok:
Would it make sense to have a function that reads a file and return the lines as an array of pointers?
Just in order to split file reading and the actual processing.

tools/astylerc
darvark: Those two lines together causes strange formating. 
This is how example function looks like when both of them are enabled
obraz

and now how when one of them is disabled
obraz

Sorry but for me looks like something is not right with astyle or that setup.

test/rules/cqww/rules/cqww
darvark: wow... I didn't removed them intentionally. Will check what and restore them

darvark:
> Would it make sense to have a function that reads a file and return the lines as an array of pointers? Just in order to split file reading and the actual processing.

So you want to have new file dedicated only to read file and returning content of this file as array of `char *` ?

tools/astylerc
darvark: Both options works together only when `indent=force-tab-x` is set to number that is not a multiplayer of 4 (when  at the same time 'indent=force-tab` is set to 4

airween:
> Wondering if it wouldn't be too much to add checking if `errno != ENOMEM` to codition `if (inputbuffer_len > 0)`. Theoretically getline still can read some amount of data before exceed memory so input buffer will be non zero.

IMHO checking the result of memory allocation according to different aspects can never be too much. :)

src/addmult.c
airween: > According to documentation message will be: for code example:

Oh, that's new for me - thanks.

Then that would be fine for me.

> but if fprintf is a standard I can change it. 

I don't believe there is a "standard" way. In the (Tlf's) code above you check the result of memory allocation. Theoretically you will get an error message something like "Not enough memory" - without any context (eg. in which file and which function triggered the error). This is why I prefer the `fprintf()` with that way, but I do not consider it as "standard".

As I wrote, this would be fine for me. (With one note: this string is enough for argument: "RuntimeError" - no need the `:` and extra space)

zcsahok:
> So you want to have new file dedicated only to read file and returning content of this file as array of `char *` ?

The typical usage pattern is

1. look for the file and open it
2. read it line-by-line in a loop
    3. process each line

The loop in Step 2 is surely generic. So one could even make a function for it
```
int process_file(FILE *f, char *filename, char *(*func)(char *), char *comment_chars);
```
where `func` does the use case specific processing (Step 3). It returns a non-null message in case of an error.
Lines that start with one of the chars in `comment_chars` are treated as comments and skipped.

This would be even better than reading-in the whole file.

darvark:
@zcsahok I think it worth to consider such functionality, but probably it's outside of this issue scope. We should/can discuss this in separate thread and create new issue for it. Such change would be bigger than just switch from fgets to getline.

tools/astylerc
zcsahok: The issue is with the IDE you are using. Vim shows everything as expected:
image

Try with these VS Code settings:
```
$ cat .config/Code/User/settings.json
{
    "editor.tabSize": 8,
    "editor.insertSpaces": true,
    "editor.indentSize": 4,
    "editor.detectIndentation": false
}
```

image

src/addmult.c
```diff
@@ -364,29 +361,34 @@ int init_and_load_multipliers(void) {
...
+	    add_mult_line(s_inputbuffer);
 	}
-
-	add_mult_line(s_inputbuffer);
-
     }
 
     fclose(cfp);
-
+    if (s_inputbuffer > 0)
```
zcsahok, line 384: Comparing a pointer and 0. It works, but using the typical check
```
if (s_inputbuffer != NULL) 
```
would be clearer.

src/checklogfile.c
```diff
@@ -112,13 +120,12 @@ int repair_log(char *filename) {
...
-
-
 int checklogfile_new(char *filename) {
-
     int lineno;
     int tooshort;
-    char buffer[160];
+    int read;
+    char *buffer = NULL;
+    size_t buffer_len = 160;
```
zcsahok, line 128: Do not set `buffer_len`.

src/checklogfile.c
```diff
@@ -221,20 +237,20 @@ int checklogfile_new(char *filename) {
...
 void checklogfile(void) {
-
     extern char logfile[];
 
     int qsobytes;
+    int read;
     struct stat statbuf;
-    char inputbuffer[800];
+    char *inputbuffer = NULL;
+    size_t read_len = 160;
```
zcsahok, line 246: Do not set `read_len`.

src/dxcc.c
```diff
@@ -332,8 +333,10 @@ void dxcc_add(char *dxcc_line) {
 /** load cty database from filename */
 int load_ctydata(char *filename) {
     FILE *fd;
-    char buf[181] = "";
+    char *buf = NULL;
+    size_t buf_len = 181;
```
zcsahok, line 337: Do not set `buf_len`.

src/parse_logcfg.c
```diff
@@ -660,7 +669,9 @@ static int cfg_dx_n_sections(const cfg_arg_t arg) {
 }
 
 static int cfg_countrylist(const cfg_arg_t arg) {
-    char buffer[2000];
+    char *buffer = NULL;
+    size_t buffer_len = 2000;
```
zcsahok, line 673: Do not set `buffer_len`.

src/readcabrillo.c
```diff
@@ -391,16 +392,17 @@ void show_readcab_msg(int mode, char *msg) {
  */
 
 int readcabrillo(int mode) {
-
     struct cabrillo_desc *cabdesc;
     char input_logfile[24];
     char output_logfile[80], temp_logfile[80];
-    char logline[MAX_CABRILLO_LEN];
```
zcsahok, line 398: Remove unused MAX_CABRILLO_LEN from MAX_CABRILLO_LEN.

src/readqtccalls.c
```diff
@@ -21,31 +21,31 @@
...
 int qtcdirection = 0;
 int next_qtc_qso;
 int qsoflags_for_qtc[MAX_QSOS];
 int nr_qtcsent = 0;
 
 int readqtccalls() {
     int s = 0;
-    char inputbuffer[160];
+    char *inputbuffer = NULL;
+    size_t inputbuffer_len = 160;
```
zcsahok, line 43: Do not set length.

src/parse_logcfg.c
```diff
@@ -672,33 +683,43 @@ static int cfg_countrylist(const cfg_arg_t arg) {
...
 	g_free(prefix);
 	fclose(fp);
+
     } else {	/* not a file */
-	char *colon = index(buffer, ':');
-	if (colon != NULL) {
-	    country_list_raw = colon + 1;   // skip optional contest name
-	} else {
-	    country_list_raw = buffer;
+	if (buffer != NULL) {
```
zcsahok, line 716: buffer is never NULL at this point. In case the previous calloc fails, then the following g_strlcpy (line 687) will already cause a segmentation violation.

src/searchlog.c
```diff
@@ -747,7 +748,9 @@ int load_callmaster(void) {
 
     FILE *cfp;
     char *callmaster_location;
-    char s_inputbuffer[186] = "";
+    char *s_inputbuffer = NULL;
+    size_t s_inputbuffer_len = 186;
```
zcsahok, line 752: Do not set length.

src/writecabrillo.c
```diff
@@ -94,18 +95,26 @@ struct linedata_t *parse_logline(char *buffer) {
  * \return ptr to new qso record (or NULL if eof)
  */
 struct linedata_t *get_next_record(FILE *fp) {
-
-    char buffer[160];
+    char *buffer = NULL;
+    size_t buffer_len = 160;
```
zcsahok, line 99: Do no set length.

zcsahok:
My point was to avoid copy-pasting the same code to all affected source files. This can be done also later, no problem.

This PR contains 43 changed files including 3 file mode changes. On the other hand there are only 12 files in `src/` that use fgets. The extra changes are due to astyle reformatting as this was not applied previously in a consistent manner. Additionally, the comments in some files like `globalvars.h` or `qtcvars.h` get mangled when reformatted.

So, I suggest to consolidate this PR as follows
- take this as a successful prototyping and create a brand new branch from master head
- copy the 12 .c files from this branch considering the review comments
- apply astyle to the 12 files only, keep others unchanged
- open a new PR

In the meantime I am preparing a PR to allow unconditional application of astyle
and also actually reindent the code in a controlled way.

Then we check the feasibility of introducing a generic file reading function.

Does this sound like a plan?

darvark:
> My point was to avoid copy-pasting the same code to all affected source files. This can be done also later, no problem.
> 
> This PR contains 43 changed files including 3 file mode changes. On the other hand there are only 12 files in `src/` that use fgets. The extra changes are due to astyle reformatting as this was not applied previously in a consistent manner. Additionally, the comments in some files like `globalvars.h` or `qtcvars.h` get mangled when reformatted.
> 
> So, I suggest to consolidate this PR as follows
> 
>     * take this as a successful prototyping and create a brand new branch from master head
> 
>     * copy the 12 .c files from this branch considering the review comments
> 
>     * apply astyle to the 12 files only, keep others unchanged
> 
>     * open a new PR
> 
> 
> In the meantime I am preparing a PR to allow unconditional application of astyle and also actually reindent the code in a controlled way.
> 
> Then we check the feasibility of introducing a generic file reading function.
> 
> Does this sound like a plan?

Yes, thank you for all your comments.

PR #400, opened 2023-08-05
Autoreconf remove obsolete macros

darvark:
Some new distributions introduced new toolchain in which autoreconf drops support for some macros.
`AC_PROG_CC_C99` and `AC_HEADER_STDC` are not longer supported and were replaced by `AC_PROG_CC`
https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Obsolete-Macros.html


zcsahok:
Please check #399

PR #401, opened 2023-08-12
#348 Review usage of fgets in file reading 

darvark:
Rewrited  changes to replace fgets with getline.


zcsahok:
Let's first fix a git-technical issue: the master branch of your repo is not in sync the master of the TLF repo.
image
It should look like this
image

The yellow marked "Sync fork" syncs your repo with the upstream (TLF) repo. It may not work in your case as you're 12 commits ahead. As a fallback solution you could create another fork.

Only if your master is in sync, then create branches for PRs. Once they are merged into TLF, then sync your repo and you'll have your branch merged into your master.
Maybe check this: https://github.com/PointCloudLibrary/pcl/wiki/A-step-by-step-guide-on-preparing-and-submitting-a-pull-request

All this is not TLF specific, it's the normal git workflow.

darvark:
I can't create new fork, have to remove current one, or create new github repository. Sync fork as you noticed won't work in my case. So probably only recreating fork left. Will solve this issue soon (I had long weekend without access to pc)

PR #402, opened 2023-08-19
Audit KSQP mults lists

N0NB:
Remove 'DC' for states mults list as it is not a mult for this event.

Break lines of all mult lists at ten entities per line for ease of counting.

Add link to KSQP 2023 rules.


zcsahok:
Thanks for the update!

N0NB:
You're welcome.  I probably should have been a bit more proactive...

PR #404, opened 2023-08-20
Add keyword to disable rig mode setting

darvark:
Tested with FTdx3000, os: Elementary OS 7.
Test case: Rig setup for bands: 40,20,15m to SSB, for 80m for CW. FOLLOW_MODE enabled in logcfg.dat, tlf started on 80m with CW mode. Switching band on tlf will cause to switch modulation on rig to CW.
In case when FOLLOW_MODE is disabled, then changing bands will not cause switching mode on rig to CW, just new band which is visible in tlf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ sudo apt install sox xplanet
 
 Once the dependencies are installed, the easiest way to get TLF's source is by
 downloading the latest tarball (version 1.4.1) from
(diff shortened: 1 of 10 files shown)


darvark:
Any comments on that change?

dl1jbe:
Will have a look in next days. Thanks for reminder.

dl1jbe:
See comment on #286

zcsahok:
Superseded by #458

PR #405, opened 2023-08-23
Update California QSO party multipliers

VictorDenisov:
This year formerly combined entities
are expanded into individual multipliers:
https://cqp.org/pdf/cqp2023_rules_changes.pdf


zcsahok:
Thanks for the update!

PR #408, opened 2023-09-10
Apply astyle

zcsahok:
Follow-up from #398 - consistently applying astyle to all source files. (src/*.[ch])

Some specially formatted comments are protected using INDENT-OFF/ON.

Trailing comments in main.c moved before the definition they to.

### DIFF
--- a/.github/workflows/ci-build-ubuntu-20.yml
+++ b/.github/workflows/ci-build-ubuntu-20.yml
@@ -22,9 +22,13 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get -qq update
-        sudo apt-get install -y libhamlib-dev libxmlrpc-core-c3-dev libcmocka-dev python3-pexpect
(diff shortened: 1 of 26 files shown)


zcsahok:
The following checksum stays invariant after astyle:
```
$ cat src/*.[ch] | tr -d ' \t\n' | md5sum
4b07abb970a4164588fabb9723365bf2  -
```
(MD5 of all files with whitespace and newlines removed)

airween (approved):

airween:
Thanks Zoli, nice work.

airween:
I was wondering if we should force the use of astyle, eg. put a small script into the workflow what format the committed code and diff it with the original.

zcsahok:
A check step could be surely added to the actions. Thanks for the idea.

zcsahok:
The check fails if there are files to be formatted. The output provides the list of affected files.
image

airween:
LGTM, thanks!

PR #412, opened 2023-10-19
Provide a fallback for M_PI and M_PI_2

dl1jbe:
As C standards does not define M_PI or M_PI_2 provide a fallback. Some LIBC implementations (e.g. MUSL) miss these definitions.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -58,7 +58,7 @@ noinst_HEADERS = \
 	keyer.h keystroke_names.h \
 	lancode.h last10.h listmessages.h log_utils.h \
 	log_to_disk.h logit.h logview.h \
-	makelogline.h messagechange.h muf.h \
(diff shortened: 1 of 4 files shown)


zcsahok:
Could we just define one of these? For example drop `M_PI` as `M_PI_2` is used in more places and define `RADIAN` in terms of `M_PI_2`? Ideally even `M_PI` shall come from a single (.h) file.

dl1jbe:
Makes sense. I would also move the definition of RADIANS there.
I see two ways:

a) prepare a tlf_math.h (which includes math.h and defines M_PI_2 if needed) and include it everywhere instead of math.h.

b) define it in sunup.h as that is already needed by muf.c (but has to be added to sunup.c yet).

dl1jbe:
Thanks for the feedback.

PR #415, opened 2023-10-30
Fix 2-character prefix search

zcsahok:
In September 2018 cty.dat (https://www.country-files.com/cty-2810-04-september-2018/) added G8ERJ to US
```
G8ERJ, N8CL and N8RA are all United States, K in CQ zone 5, ITU zone 8
```

This causes all G8 calls considered as US calls. There was a missing logic to handle exact matches correctly.

First added a test for it.

### DIFF
--- a/src/dxcc.c
+++ b/src/dxcc.c
@@ -149,6 +149,7 @@ int find_best_match(const char *call) {
 	w = two_char_prefix_index[key];
 	if (w >= 0) {
 	    bool ok = true;
+	    // for an exact entry require the whole call to match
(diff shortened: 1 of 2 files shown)


zcsahok:
As expected test fails with _"G8ERJ" != "G"_

zcsahok:
Now fixed.

dl1jbe:
Nice problem. 

Maybe add to comment that having a unique prefix is only enough if call has not to be exact. Otherwise whole call should match.

Otherwise looks good and should go in.

PR #416, opened 2023-11-21
Fix startup if HAMLIB_KEYER selected and rig is off or disabled at start

dl1jbe:
Otherwise TLF will abort unconditionally.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -780,6 +780,7 @@ static void hamlib_init() {
 	    exit(1);
 	}
 	trx_control = false;
+	no_trx_control = true;
 	showmsg("Disabling rig control!");
 	sleep(1);
     }
@@ -872,6 +873,10 @@ static void keyer_init() {
 
     if (cwkeyer == HAMLIB_KEYER) {
 	showmsg("CW-Keyer is Hamlib");
+	if (no_trx_control) {
+	    showmsg("Radio control disabled - no keying!");
+	    return;
+	}
 	if (!trx_control) {
 	    showmsg("Radio control is not activated!!");
 	    sleep(1);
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -872,6 +873,10 @@ static void keyer_init() {
 
     if (cwkeyer == HAMLIB_KEYER) {
 	showmsg("CW-Keyer is Hamlib");
+	if (trx_control_disabled) {
+	    showmsg("Radio control disabled - no keying!");
+	    return;
+	}
 	if (!trx_control) {
 	    showmsg("Radio control is not activated!!");
```
zcsahok, line 881: Can we now reach this message? I.e. trx_control_disabled==false and trx_control==false

dl1jbe: Hmm, that state could only happen if you have RIGCONTROL disabled in logcfg.dat. Than having HAMLIB_KEYER activated seems to be an configuration error which is catched in the following if-statement. But I think that problem should be catched in parse_config() before.

zcsahok: Indeed. Then the second `if` ("Radio control is not activated") has to be kept, even if only this quite specific use case triggers it. I'd not bother adding any other check for it.

On the other hand it may be worth promoting the message of the first `if` to a warning by adding a sleep. Now this message just flashed for a split second unless using verbose mode. Unless of course if we rely on the previous question (if the rig was off) or assume the user knows what they they are doing (if started with -r option).

dl1jbe: Good points. Give me a day or two to think about.

dl1jbe: Just added a 2 second delay (see commit below). should be enough to see a problem and restart with -v if needed.

zcsahok (approved):
Thanks.

PR #417, opened 2023-12-01
Update arrlsections to the current list

N0NB:
Canada (RAC) made a number of changes to their sections which became effective sometime this year.  While this list will work going forward, I think there needs to be discussion of handling this list for backward compatibility in the future.

### DIFF
--- a/share/arrlsections
+++ b/share/arrlsections
@@ -13,7 +13,7 @@ ENY
 EPA
 EWA
 GA
-GTA
+GH
 IA
 ID
 IL
@@ -22,7 +22,6 @@ KS
 KY
 LA
 LAX
-MAR
 MB
 MDC
 ME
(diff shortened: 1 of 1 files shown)


N0NB:
As noted in the commit message, there should probably be some discussion on handling this list for backward compatibility.

zcsahok:
Thanks for the update. IMO users requiring backward compatibility (being able to open past logs using the then-current rules) shall store the rules/settings in separate directories. Something like arrldx2020, arrldx2021, ...
TLF has no explicit support for general backward compatibility.

dl1jbe:
Same goes for cty.dat database. Should be in the contest directory to avoid a re-scoring due to a newer database file.

dl1jbe:
Closing. See PR #447.

Issue #414, opened 2023-10-23
Use of 'getline' (PR 403) needs more work


dl1jbe:
There seems to be some problems in the switch to using getline() instead of fgets().

a) In some places the return value is checked against -1 (error) and in other cases against +1 (one char read).

b) Reading from man page if the function does *NOT* return -1 it means that one line got read from the file without errors. So checking for the length of the buffer being enough and checking for ENOMEM inside the while-loop seems unneeded. It should be checked after leaving the loop to decide between error and EOF condition.

As a further hint - the null pointer guard 'if (buffer) ..' before free() can be dropped as free works fine with null pointers.

@darvark Do you have a chance to work on  it? I would suggest starting with one instance and using it as a template for the others later.

darvark:
Hi,
Yes I can work on it.

dl1jbe:
Good. I would suggest starting with the one in cabrillo_utils.c.

zcsahok:
Replacing getline() was a longer process. In my comment https://github.com/Tlf/tlf/pull/398#issuecomment-1665115762 I proposed a generic line-by-line file reading function. This was then postponed to a later stage (https://github.com/Tlf/tlf/pull/398#issuecomment-1670100505). I can also take it.

dl1jbe:
Yes I saw it from the mails and comments on the PRs. 
I just did a check and saw the inconsistent checks of the return value and the wrong error handling.

Your idea of a generic processing function is a good one. I would suggest that we as a first step correct the problems and migrate afterwards. Let us give @darvark a chance to sort it it out with our help.

darvark:
Now I'm preparing my club station for cqww, so I have not enought free time, please be patient, I'll start working on that just after cqww ssb.

dl1jbe:
Take your time and good luck in the contest.

darvark:
Sorry for late reply, finally I found some time and I think I manage to unify all usage of getline(), but now I have problem with one test: run_rules, and as far as I know it uses python script run_tlf.py. As soon as I solve that issue I'll push changes for review

dl1jbe:
Commit got merged. If no further comments, I will close the issue in coming week.

zcsahok:
Fine with me. I'll look into the generic reading once I get to it.

PR #419, opened 2023-12-24
Fix getline usage

darvark:
Fixes according to issue https://github.com/Tlf/tlf/issues/414.

Still can't find solution for failing test_rules error.

### DIFF
--- a/ToDo
+++ b/ToDo
@@ -41,7 +41,7 @@ Planned for later versions
 Internals
 ~~~~~~~~~
 [x] Factor out helper functions to parse parts of logline (band, mode, ...)
-[ ] change internal QSO representation to structs (makes coding much
+[x] change internal QSO representation to structs (makes coding much
     simpler and will enable migration of log file to a different format
     later)
 
--- a/src/addmult.c
+++ b/src/addmult.c
@@ -363,11 +363,6 @@ int init_and_load_multipliers(void) {
 
     while ((read = getline(&s_inputbuffer, &s_inputbuffer_len, cfp)) != -1) {
 	if (s_inputbuffer_len > 0) {
-	    if (errno == ENOMEM) {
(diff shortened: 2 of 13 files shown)


zcsahok:
The tests seem to be all OK.

dl1jbe:
Looks much better and fixes most problems.

Just one question left: What do you want to check with the lines like

```
if (s_inputbuffer_len > 0) {
```
and similar. Do you want to make sure there is an allocated buffer? That should be guaranteed by readline() returning without an error.

darvark:
Yes I want to be sure that something was read. getline will return -1 only in case of error i other case returns number of characters read. So it can return in theory 0, and function will have to work on zero length buffer? In theory I can remove that check, but not sure if it's sully desired and safe

dl1jbe:
I see. But than you should test the value returned by getline() instead. That would be

'''
while ((read = getline(&s_inputbuffer, &s_inputbuffer_len, cfp)) != -1) {
  if (read > 0) {
   ....
  }
}
'''

If I read the man page right the inputbuffer will be expanded as needed but not shrink-ed if it is to large, so its size from 's_inputbuffer_len' may not represent the line length.

darvark:
Right, Ok, switching condition. Meanwhile one question (not related to that PR). Is there any documentation about usage python plugins? We can enable support or python plugins, bu I didn't found anything about usage or how to write them

dl1jbe:
> Meanwhile one question (not related to that PR). Is there any documentation about usage python plugins? We can enable support or python plugins, bu I didn't found anything about usage or how to write them

Please see TLFs man page section `PYTHON PLUGIN'.

As the plugin was made by Zoli, he can give more informations if needed.

dl1jbe:
Your code looks good now. There are some places were a check for empty lines is already in place. But we will sort that out later.

Thanks for the work on it. Will merge it in now.

PR #420, opened 2024-01-04
Marconi Club QSO Party Day rules

zcsahok:
(International Marconi Day is in April and there is also a Marconi Memorial in July, so this could be somewhat confusing)


zcsahok:
The `bandindex` field was not set in qso struct created from a spot. As a result the multiplier could not be shown correctly on bandmap (as all spots were treated as being a 160 m (= bandindex 0) spot).

Added deriving multiplier info from initial exchange file. The result is that new mults show up on the bandmap.
image

zcsahok:
The feedback from @ok1rr was positive, so if there are no issues with the code change then I'd merge this in.

dl1jbe:
> The feedback from @ok1rr was positive, so if there are no issues with the code change then I'd merge this in.

Yes. please go ahead with it.

PR #421, opened 2024-01-08
Fix frequency input logic

zcsahok:
Changing frequency by entering it in the call input field is a handy feature, but it has some rough edges.

1. if an out-of-band frequency (all digits) is entered then it is treated erroneously as a callsign
image

2. similarly, when there is no rig control then all numeric values are taken for a callsign
image

Reworked `sendqrg()` to use `plain_number()` to check if input is a number. Originally `atof()` was used to parse it, presumably to support fractional kHz values. As dot is used to configure bandmap this part of the feature was broken anyway, so I replaced it with a simple `atoi()`.
In `logit.c` moved the clearing in case of freq up to the check itself and thus reduced the indenting of the code actually doing relevant stuff.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -1059,10 +1059,14 @@ bool valid_call_char(int ch) {
  * Check if string contains only digits
  * \param str    the string to check
  * \return true  if only digits inside
- *         false at least one none digit
+ *         false contains at least one non-digit or it's an empty string
  */
 bool plain_number(char *str) {
     int i;
+
+    if (strlen(str) == 0) {
+        return false;
+    }
 
     for (i = 0; i < strlen(str); i++) {
 	if (!isdigit(str[i])) {
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -1065,7 +1065,7 @@ bool plain_number(char *str) {
     int i;
 
     if (strlen(str) == 0) {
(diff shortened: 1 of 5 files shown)


dl1jbe:
Looks good from my side.

While looking up the changes I found an error in callinput.c line 1257. send_bandswitch() expects an frequency as parameter but gets an bandindex. Maybe you can add a fix for it too.

zcsahok:
Fixed the argument for bandswitch. Probably nobody uses the parallel(!) port info cwdaemon can provide, so this went unnoticed.

Also fixed `plain_number()`: it returned true for an empty string. When called from `callinput.c` it is always ensured that the input is not empty, that's why this wasn't a problem until now.
The effect of treating empty string as a number broke keyer window if the input field was empty.

dl1jbe:
> Fixed the argument for bandswitch. Probably nobody uses the parallel(!) port info cwdaemon can provide, so this went unnoticed.
> 
Thanks.

> Also fixed `plain_number()`: it returned true for an empty string. When called from `callinput.c` it is always ensured that the input is not empty, that's why this wasn't a problem until now. The effect of treating empty string as a number broke keyer window if the input field was empty.

Fine too.

Whole PR looks good. Should be merged in.

PR #424, opened 2024-01-27
Terminate background thread gracefully

### LINKED ISSUE #422: Sporadic lock-up on exit
Sometimes TLF freezes on exit. It happens quite rarely, but the other day I had a chance to look into it using gdb. The process sits in `close_tlf_rig()` waiting for `tlf_rig_mutex` (see #363) .

What happens is the following

- `background_task` performs a Hamlib operation with `tlf_rig_mutex` locked
- user exits TLF and `background_task`'s thread is instructed to terminate using `pthread_cancel()` in `tlf_cleanup()`
- the thread exits according its default setting `PTHREAD_CANCEL_DEFERRED`, that is roughly after next syscall, irrespective of the application state, including mutexes.
- so `tlf_rig_mutex` stays locked
- once the thread had exited, `close_tlf_rig()` tries to lock `tlf_rig_mutex`, but never manages it
- process has to be killed manually

As code spends just some milliseconds in Hamlib calls all this happens only occasionally, but it's still quite annoying.

The preferred way of existing `background_task` would be gracefully signaling via a shared variable. Something like
```
  while (!stop_backgrnd_process) {
       // loop body
  }
```
Current `stop_backgrnd_process` should be renamed to `pause_backgrnd_process` as it effectively only pauses/suspends it.

zcsahok:
Fixes #422

### DIFF
--- a/src/background_process.c
+++ b/src/background_process.c
@@ -43,39 +43,39 @@
 
 static bool go = true;
 // don't start until we know what we are doing
-static bool stop_backgrnd_process = true;
+static bool pause_backgrnd_process = true;
 
-static pthread_mutex_t stop_backgrnd_process_mutex = PTHREAD_MUTEX_INITIALIZER;
-static pthread_cond_t start_backgrnd_process_cond = PTHREAD_COND_INITIALIZER;
-static pthread_cond_t backgrnd_process_stopped_cond = PTHREAD_COND_INITIALIZER;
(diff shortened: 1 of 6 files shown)


zcsahok:
Renamed actions to match what they are doing

- stop -> pause
- start -> resume

dl1jbe:
Looks good. +1

PR #425, opened 2024-02-13
Fixed LOGFILE casing in pacc_pa rules

blikjeham:
The LOGFILE keyword needs to be in uppercase, or else the program will not start correctly.


airween (approved):
LGTM - thanks.

PR #426, opened 2024-02-20
Fix bare script commands

N0NB:
This patch set cleans up an ambiguous example in the man page that resulted in my actual `cron` entry killing the `gnome-keyring-daemon` on my system!  By using the `which` command in the scripts and man page example, I hope others can avoid my annoying experience.

### DIFF
--- a/scripts/soundlog
+++ b/scripts/soundlog
@@ -8,7 +8,7 @@ trap 'kill -s INT $PID' TERM USR1
 while [ -f $HOME/.VRlock ]
 do
 		# store recorded files in actual directory
-		filename=`eval date +%d%H%M`".au"
+		filename=$(eval date +%d%H%M)".au"
 		if test -f $filename
 		then
 			sleep 10


dl1jbe:
Oops. Good catch.
Just two questions:
- Why also use full path to start the programs?
- Did you experiment with dropping the -f parameter to pkill (as that seems the main culprit) or substitute it by -x (exact match)?

N0NB:
* On 2024 21 Feb 09:38 -0600, Thomas Beierlein wrote:
> Oops. Good catch.
> Just two questions:
> - Why also use full path to start the programs?

To make the pattern as unique as possible.  Besides, it is good practice
to do so.  Even if there is a locally compiled package that is early in
the user's PATH, `which` should find it.

OTOH, as `which` searches the user's PATH just like the bare `rec` as
originally written, there really is no difference.

> - Did you experiment with dropping the -f parameter to pkill (as that
> seems the main culprit) or substitute it by -x (exact match)?

I did experiment with dropping `-f`, but I guess I got focused on the way
I submitted the patch.  Without `-f` the pattern is limited to 14
characters as I recall which is quite sufficient for `rec`.

Thanks for pointing out `-x` as I missed that one.  A quick check with
`pgrep` shows that it does not match the substring in 'directory'.  I
suppose that changing `-f` to `-x` in the man page template is all that
is required to spare others the same pain.  Also, `-fx` appears to
prevent the substring match as well.

73, Nate

-- 
Web: https://www.n0nb.us

dl1jbe:
> To make the pattern as unique as possible. Besides, it is good practice to do so.

Hmm, I know these practice with complete path to NOT allow bash to pick up the first matching binary with that name in PATH - especially for security reasons..
As 'which()' according to main page uses the same search order as bash the suggested construct would not help here.

> Even if there is a locally compiled package that is early in the user's PATH, `which` should find it.

Yes and the shell would do also.
I am not sure if we really get something useful out of that added complexity. 


But back to the main point - the -f parameter:

> I did experiment with dropping `-f`, but I guess I got focused on the way I submitted the patch. Without `-f` the pattern is limited to 14 characters as I recall which is quite sufficient for `rec`. Thanks for pointing out `-x` as I missed that one. A quick check with `pgrep` shows that it does not match the substring in 'directory'. I suppose that changing `-f` to `-x` in the man page template is all that is required to spare others the same pain. 

I did some search in gits history. The section in the man page dates back to Rein's work before 2010. Seems that nobody (me included) did a thorough check of that part. Further in all other cases we use 'pkill()' in the source code we have no -f around.

So I agree that changing -f to -x in the man page is the way to go.

N0NB:
* On 2024 21 Feb 15:01 -0600, Thomas Beierlein wrote:
> So I agree that changing -f to -x in the man page is the way to go.

Go ahead and do that, Tom.

73, Nate

-- 
Web: https://www.n0nb.us

dl1jbe:
See PR #427

PR #427, opened 2024-02-22
Fix pkill man page

dl1jbe:
Cleans up the ambiguous -f parameter from the pkill example in man page. Additionally replace backticks by $() as suggested by N0NB. 

Replaces PR #426.


N0NB (commented):
Common English on this side of the pond would use the word "caught".

N0NB (approved):

N0NB (approved):

PR #428, opened 2024-03-26
test/test_addpfx.c: Fix declaration of pfxmultab

df7cb:
This was causing test failures on Debian's s390x architecture.

Since the last release was so long ago, I uploaded a git snapshot to Debian. On big-endian s390x, one test was actually failing. (On other architectures this is a just a compiler warning.)


zcsahok:
Thanks!

PR #431, opened 2024-04-30
Improve keyer: unify key handling and add graceful closing of keyer window

zcsahok:
(this is still work in progress, comments are welcome)

My motivation was twofold:

1) provide a single-key alternative to close keyer window without stopping keying. This allows faster typing of additional messages. No need to press Ctrl-K for a graceful exit. The new closing key is ` (grave accent or back tick) that is located below Esc on the US keyboard.
image

2) allow use of all F-key, Alt-key messages, plus keyer related other keys (speed up/down, weight, tuning) uniformly in all input fields: call input, exchange input and keyer window. This allows also removing code duplication.

"Keyboard" mode had to defined as an additional extension to CQ/S&P (as opposed to an standalone state), as some keys behave differently if one was in CQ or S&P mode.

tune() was moved to keyer() as the common handling is located there.

There is one issue: the `SEND_DE` (=F1 sends "DE <mycall>" in S&P mode) seems to affect call input field only. In exchange input field the plain "<mycall>" is sent. Not sure if this was deliberate or was simply forgotten.
With the common handling it's technically much easier to send in both cases "DE <mycall>".
If the user still wants to send their bare callsign it is always possible using F6.
My suggestion is take this change of `SEND_DE` handling as a bug fix.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -270,6 +270,8 @@ int callinput(void) {
 	    x = KEY_F(1);   // continue as F1
 	}
 
+	x = handle_common_key(x);
+
 	switch (x) {
 	    // Ctrl-V: toggle grab direction
 	    case CTRL_V:
@@ -361,43 +363,6 @@ int callinput(void) {
 		break;
 	    }
 
-	    // Alt-w (M-w), set Morse weight.
-	    case ALT_W: {
-		char weightbuf[5] = "";
-		char *end;
-
-		mvaddstr(12, 29, "Wght: -50..50");
-
-		nicebox(1, 1, 2, 12, "Cw");
-		attron(COLOR_PAIR(C_LOG) | A_STANDOUT);
-		mvprintw(2, 2, "Speed:   %2u ", GetCWSpeed());
-		mvprintw(3, 2, "Weight: %3d ", weight);
-		refreshp();
-
-		usleep(800000);
-		mvaddstr(3, 10, "   ");
-
-		echo();
-		mvgetnstr(3, 10, weightbuf, 3);
-		noecho();
-
-		g_strchomp(weightbuf);
-
-		int tmp = strtol(weightbuf, &end, 10);
-
-		if ((weightbuf[0] != '\0') && (*end == '\0')) {
-		    /* successful conversion */
-
-		    if (tmp > -51 && tmp < 51) {
-			weight = tmp;
-			netkeyer(K_WEIGHT, weightbuf);
-		    }
-		}
-		clear_display();
-
-		break;
-	    }
-
 	    // Alt-v (M-v), change Morse speed in CW mode, else band down.
 	    case ALT_V: {
 		if (ctcomp) {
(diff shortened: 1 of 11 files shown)


df7cb:
Wouldn't it be more intuitive to have <enter> close the keying window?

zcsahok:
Which key do you mean?

df7cb:
Sorry, markdown/html ate it... `<enter>`

zcsahok:
Would be fine for CW, but it does have a meaning for digimodes.
I assume nobody uses back tick even in digimode. And it's just "almost" Esc on the keyboard.

df7cb:
Ok, understood - perhaps it could be both `enter` and `backtick` in CW and just `backtick` in digi?

"Type message and hit enter to send" would be quite natural I'd think

zcsahok:
Checked the code: enter is mapped to space (newline is send by | (pipe)).
We could change this mapping to happen for digimode only.

dl1jbe:
Just some comments:

>     1. provide a single-key alternative to close keyer window without stopping keying. This allows faster typing of additional messages. No need to press Ctrl-K for a graceful exit. The new closing key is ` (grave accent or back tick) that is located below Esc on the US keyboard.

Position of special keys like ` are not consistent between keyboards for different languages, so the decision may be counterproductive. E.g. on normal german keyboard ` needs to use shift key in addition, while ESC is a key of its own. Laptop keyboards are also a different story.

In contrast the suggested '<enter>' looks quite standardized.

If not applicable how about something like 'KEY_DOWN' arrow - like 'bring the keyer window down'?

 
> There is one issue: the `SEND_DE` (=F1 sends "DE " in S&P mode) seems to affect call input field only. In exchange input field the plain "" is sent. Not sure if this was deliberate or was simply forgotten.

Was kept from Rein's old TLf versions (just had a look into 0.9.10).

> With the common handling it's technically much easier to send in both cases "DE ". If the user still wants to send their bare callsign it is always possible using F6. 

Could work. F1 mostly used to just call the other station in S&P, so F6 means you have to change your standard routine (I know it can be done by simply pressing Enter-key, but I remember discussions with hams using only the F-keys).

> My suggestion is take this change of `SEND_DE` handling as a bug fix.

Which one? a) having SEND_DE respected in both fields or b) hardcoding F1 to use F1 and F6 to not use it?

If b) we could drop the SEND_DE keyword completely.

dl1jbe:
Code looks good. 

Only question: What do we gain by using separate 'keyboard_mode'?

zcsahok:
Added closing on enter in CW mode.
Not sure why was the enter->space mapping introduced, as enter makes completely sense in digimode.

zcsahok:
Re SEND_DE: my preference in a), so F1 would send "DE <call>" in both input fields. But b) would be also fine: stations calling with DE are very rare.

zcsahok:
Re keyboard_mode: originally keyboard mode (=using keyer window) was a standalone mode with it's own key handling. The mode was temporarily overwritten when in keyer and then restored. This caused a minor discrepancy as F1 behaved differently from the normal input mode (CQ vs. S&P).
The solution is to define keyboard mode as "sub-mode" of the input mode. This allows keyer window to evaluate the original input mode and thus F1 behaves the same way as in other input fields.

dl1jbe:
> Not sure why was the enter->space mapping introduced, as enter makes completely sense in digimode.
> 
Without that mapping the display of the keyer windows gets garbled (cursor switches to next line to border of the window). We can drop the mapping for the TX but should keep it for display (or use some more obvious character to show the line break).

> Re SEND_DE: my preference in a), so F1 would send "DE " in both input fields. But b) would be also fine: 
> stations calling with DE are very rare.

Ok, I would say lets stay with a). That way input and exchange field have the same behaviour and prepending DE stays optional.

Wrt keyboard_mode: Thanks for explanations. Makes sense.

zcsahok:
Added closing on cursor down. Now we have a lot of options... ;-)

I see the point regarding mapping enter key. For the time being I'd leave it as is. It needs a more thorough testing with Fldigi.

Will update the docs as soon as code is stable.

zcsahok:
Documented Enter too. And trying to avoid line break on the hyphen in Ctrl-K/Alt-K.

PR #434, opened 2024-05-22
Adding His Majesty King of Spain CW/SSB contest

KHoos:
I created this plugin and ruleset to make tlf work for the His Majesty King of Spain CW contest.
Verification with the rules of the His Majesty King of Spain SSB contest show they should work there.

This code has been tested in the contest. The resulting scoring has not yet been verified with the scoring by the URE, which I hope to do in a number of weeks when they publish the first results.

### DIFF
--- a/rules/eakingofspain/eakingofspain
+++ b/rules/eakingofspain/eakingofspain
@@ -1,0 +1,77 @@
+#
+# ruleset for EA King of spain contest CW and SSB
+#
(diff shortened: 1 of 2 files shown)


KHoos:
Scoring by the code compared to the log report made available to me by the URE shows no differences in scoring

zcsahok:
Thanks!

PR #435, opened 2024-05-28
Fix mult info when on a WARC band

zcsahok:
When switching to a WARC band the score summary keeps showing the multiplier count from the corresponding contest band. QSO counts are displayed correctly.
Caused by not using `bi_warc` while showing mult info.

Example: 
score as visible when on 40 m -
image
after switching to 30 m: QSO count is zero, but mults are kept from the normal display -
image
fixed display, with 60 m in place of 40 m when on a WARC band  -
image

Of course due to the limited amount of columns the displayed score does not add up when on a WARC band.


dl1jbe:
Looks good and should go in.

PR #436, opened 2024-08-18
Add IOTA contest rules

zcsahok:
As per-mode (CW/SSB) multipliers are not supported yet it has been worked around by appending `c`/`s` to the multiplier value.

Island stations shall specify their reference via PLUGIN_CONFIG
e.g.
```
PLUGIN_CONFIG=EU115
```


dl1jbe:
LGTM.
Only question: Did you get a chance to sent in the cabrillo and got it accepted? Example cabrillos state IOTA index as 'EU-nnnn' while we provide 'EUnnnn'.

zcsahok:
No, I didn't participate in the contest. I'm clarifying the Cabrillo format with the organizers. EU-nnn is definitely needed, will fix it.

zcsahok:
Now references are normalized into he format expected in the Cabrillo file (EU-nnn). For non-island stations 6 dashes are used instead.
Examples:
4NA5 --> 004 NA-005
123 --> 123 ------

Also added the correct CABRILLO_EXCHANGE setting into the logcfg.dat of the tests.

dl1jbe:
Fine. Lets merge it in.

PR #439, opened 2024-09-11
Fix handling of range of allowed TLF nodes

dl1jbe:
According to man page only nodes A..H are allowed. That corresponds with the defined number of nodes MAXNODES = 8.

The patch fixes the check for allowed NODE numbers to respect this setting.


airween (approved):
LGTM

PR #442, opened 2024-10-01
Support several multipliers for a QSO

### LINKED ISSUE #411: Submitting contact with several multipliers
California QSO party has this concept of county lines. Contact with county line stations gives two multipliers.
Are there any ways for recording several multipliers in one QSO?

zcsahok:
Allow generating several multipliers for a single QSO. These are generated by the corresponding contest plugin as a space separated list. Usages include both explicit multipliers as for California QSO party and also derived ones as for EU-DX. In the latter case both DXCC and the EU region code count as multipliers.

Added rules for EU-DX.

Resolves #411.

### DIFF
--- a/src/addmult.c
+++ b/src/addmult.c
@@ -40,6 +40,49 @@
 GPtrArray *mults_possible;
 
 /*
+ * process a space separated list of multipliers from qso->mult1_value.
+ * qso->mult1_value is updated with the list of actually applied new multipliers.
+ * if none could be applied then it contains an empty string.
+ */
+static void remember_generic_mult(struct qso_t *qso, bool check_only) {
+    static GRegex *regex = NULL;
+    if (regex == NULL) {
+	// words with trailing space(s) or at the end of string
+	regex = g_regex_new("(\\S+(\\s+|$))", 0, 0, NULL);
+    }
+
+    GMatchInfo *match_info;
+    g_regex_match(regex, qso->mult1_value, 0, &match_info);
+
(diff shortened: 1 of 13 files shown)


src/addmult.c
```diff
@@ -39,6 +39,49 @@
 
 GPtrArray *mults_possible;
```
dl1jbe, line 41: Maybe locate function near related remember_multi()?

dl1jbe (approved):
Sorry for missing that one before.

The solution will certainly work for now. Would be nice if the single mults were vertically aligned, but that looks difficult in the actual case.

In the long run a solution with a separate handling of the different multipliers would  be desirable. (Same goes for the separate scoring of multipliers per txmode.)

Otherwise LGTM.

dl1jbe:
Btw, we can give us some more room for multipliers if we change the 'displayed' date from 12-Aug-24 to 12Aug24. 

Similar, but not as readable would be to drop the ':' from 17:09. In summary that are 2..3 characters more for the information on the right side.

The displayed line would than read as

` 40CW  20Feb22 1709 0004  VE3KL          599  599  23            VE3         1  `

Be aware that I am NOT talking about changing the line format in log file, but only on display. Downside: We need separate functions to format both (But that may not be a bad decision.).

zcsahok:
Yes, the on-screen format could be tuned. We still have to store something in the log file, so there it will be truncated. As this is quite a special case, I'd postpone any changes for later.

PR #443, opened 2024-10-01
Fix storing of lan log segfault

airween:
This is a modified version of #441, based on [this](https://github.com/Tlf/tlf/pull/441#issuecomment-2385232412) discussion.

### DIFF
--- a/src/log_to_disk.c
+++ b/src/log_to_disk.c
@@ -118,7 +118,13 @@ void log_to_disk(int from_lan) {
 
 	total = total + score2(lan_logline);
 
-	struct qso_t *qso = parse_qso(lan_logline);
+	gchar *tlogline = g_strdup(lan_logline);
+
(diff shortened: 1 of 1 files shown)


src/log_to_disk.c
```diff
@@ -118,7 +118,13 @@ void log_to_disk(int from_lan) {
 
 	total = total + score2(lan_logline);
 
-	struct qso_t *qso = parse_qso(lan_logline);
+	gchar *tlogline = g_strdup(lan_logline);
+
+	struct qso_t *qso = parse_qso(tlogline);
+
+	if (tlogline != NULL) {
```
zcsahok, line 125: Check for null can be dropped: _If mem is NULL it simply returns, so there is no need to check mem against NULL before calling this function._
https://docs.gtk.org/glib/func.free.html

dl1jbe (approved):

PR #445, opened 2024-10-03
Add freq to lanentry

dl1jbe:
Fix handling of missing QRG for log entry sent via LAN.

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -284,8 +284,7 @@ int send_lan_message(int opcode, char *message) {
     }
 
     if (opcode == LOGENTRY) {
-	sendbuffer[82] = '\0';
-
+	strcat(sendbuffer, "\n");
 	lan_send(sendbuffer);
     }
 
@@ -298,7 +297,7 @@ int send_lan_message(int opcode, char *message) {
 	lan_send(sendbuffer);
     }
     if (opcode == FREQMSG) {
-	sendbuffer[10] = '\0';
+	strcat(sendbuffer, "\n");
 	lan_send(sendbuffer);
     }
     if (opcode == INCQSONUM) {
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -45,7 +45,7 @@ char lan_message[256];
 //--------------------------------------
 int bc_socket_descriptor[MAXNODES];
 ssize_t bc_sendto_rc;
-int cl_send_inhibit = 0;
+bool cl_send_inhibit = false;
 struct sockaddr_in bc_address[MAXNODES];
 /* host names and UDP ports to send notifications to */
 char bc_hostaddress[MAXNODES][16];
@@ -167,7 +167,7 @@ int lan_recv(void) {
     errno = 0;			/* clear the error */
 
     if (lan_recv_message[1] == CLUSTERMSG)
-	cl_send_inhibit = 1;	// this node does not send cluster info
+	cl_send_inhibit = true;	// this node does not send cluster info
 
     if (lan_recv_rc > 0)
 	recv_packets++;
@@ -277,7 +277,7 @@ int send_lan_message(int opcode, char *message) {
     sendbuffer[2] = '\0';
     strncat(sendbuffer, message, 98);
     if (opcode == CLUSTERMSG) {
-	if (cl_send_inhibit == 0) {
+	if (!cl_send_inhibit) {
 	    strcat(sendbuffer, "\n");
 	    lan_send(sendbuffer);
 	}
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -269,7 +269,7 @@ static int lan_send(char *lanbuffer) {
 
 /* ----------------- send lan message ----------*/
 
-int send_lan_message(int opcode, char *message) {
+void send_lan_message(int opcode, char *message) {
     char sendbuffer[102];
 
     sendbuffer[0] = thisnode;
@@ -323,7 +323,7 @@ int send_lan_message(int opcode, char *message) {
 	lan_send(sendbuffer);
     }
 
-    return 0;
+    return;
 }
 
 /* ----------------- send talk message ----------*/
--- a/src/log_to_disk.c
+++ b/src/log_to_disk.c
@@ -46,7 +46,7 @@
 
 pthread_mutex_t disk_mutex = PTHREAD_MUTEX_INITIALIZER;
 
-char lan_logline[81];
+char lan_logline[LOGLINELEN + 1];
 
 
 /* restart band timer if in wpx and qso on new band */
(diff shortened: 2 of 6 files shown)


zcsahok:
Frequency is correctly stored, but see the comments.

src/lancode.c
```diff
@@ -284,8 +284,7 @@ int send_lan_message(int opcode, char *message) {
     }
 
     if (opcode == LOGENTRY) {
-	sendbuffer[82] = '\0';
-
+	strcat(sendbuffer, "\n");
```
zcsahok, line 287: sendbuffer seems to contain a correctly formatted log line. I tested it OK without adding a new line.

src/lancode.c
```diff
@@ -298,7 +297,7 @@ int send_lan_message(int opcode, char *message) {
 	lan_send(sendbuffer);
     }
     if (opcode == FREQMSG) {
-	sendbuffer[10] = '\0';
+	strcat(sendbuffer, "\n");
```
zcsahok, line 300: Is new line needed?

src/log_to_disk.c
```diff
@@ -46,7 +46,7 @@
 
 pthread_mutex_t disk_mutex = PTHREAD_MUTEX_INITIALIZER;
 
-char lan_logline[81];
+char lan_logline[89];
```
zcsahok, line 49: LOGLINELEN + 1 ? or would LOGLINELEN suffice?

src/log_to_disk.c
```diff
@@ -106,10 +106,10 @@ void log_to_disk(int from_lan) {
...
     } else {			/* qso from lan */
 
-	/* LOGENTRY contains 82 characters (node,command and logline */
-	g_strlcpy(lan_logline, lan_message + 2, 81);
-	char *fill = g_strnfill(80 - strlen(lan_logline), ' ');
-	g_strlcat(lan_logline, fill, 81);    /* fill with spaces if needed */
+	/* LOGENTRY contains 89 characters (node,command and logline */
+	g_strlcpy(lan_logline, lan_message + 2, LOGLINELEN);
+	char *fill = g_strnfill(LOGLINELEN - 1 - strlen(lan_logline), ' ');
+	g_strlcat(lan_logline, fill, LOGLINELEN);    /* fill with spaces if needed */
```
zcsahok, line 112: I believe that this would cut away the NL added in the lan code. It is correct, as NL is appended anyway in `store_qso()`.

src/log_to_disk.c
```diff
@@ -106,10 +106,10 @@ void log_to_disk(int from_lan) {
...
 
 	if (cqwwm2) {	    /* mark as coming from other station */
```
zcsahok, line 114: How about removing this dubious check and have QSOs coming from other nodes marked with an asterisk?
image


Or even better: align the working with the man page - 
 CQWW_M2
              Put  the node ID into the logline (just after the QSO number) to
              support Multi/2 operation where the station logging the QSO must
              be in the Cabrillo file.  This can also be used for M/1 and M/M,
              to enable post-contest analysis of the nodes.

dl1jbe:
Thanks for review. See answers inline.

src/lancode.c
dl1jbe: The reason for adding the newline is on the receiving end. Lines 137-139 in background_process.c always drop the last character of the received string. That is the normal code used to drop appended newlines from strings. 

As all LAN-messages are sent as normal text strings it seems appropriate to terminate any of them with a newline. That also makes the debuglog more readable.

I think we should harmonize all LAN messages to that behavior. That would also allow us to simplify the send_lan_message() code.

At the moment (with the suggested changes) we add a newline to almost all messages. Only exception is send_time() where a space is added to the time string to avoid truncation. That should be change too.

src/lancode.c
dl1jbe: send_freq() formats the QRG as xxxx.y with one decimal. Without the newline the decimal gets dropped.

See also above.

src/log_to_disk.c
dl1jbe: Hmm, comments say LOGLINELEN (87) includes all logline information including linefeed. That would mean LOGLINELEN+1 for buffer.
But as the newline is already stripped away one character less would be enough. 
Anyway we should use symbol instead of the hard coded number.

src/log_to_disk.c
dl1jbe: As discussed above the newline is already cut off here.

src/log_to_disk.c
dl1jbe: Err, LOGLINELEN is defined as 88.

src/log_to_disk.c
dl1jbe: That is a good idea. But I would like to check first which other places needs to be changed to make it work and change that afterwards.

dl1jbe: Just reread the CQWW rules about logging the second stations contacts.
<The MULTIPLIER SIGNAL will be designated with a “1” in the transmitter column of the Cabrillo log. >

src/lancode.c
zcsahok: We are processing our own messages, so any preprocessing is at least suspicious. Blindly dropping the last char is even worse. This is a legacy code originating from af43a4d. It must have been a quick, but incorrect fix for some unknown problem.

Logging shall cope with any message and in fact it does this OK in background_process.c:
```
    148                     fputs(lan_message, fp);
    149                     fputs("\n", fp);
```
So I see no reason adding NL on the sending side.

src/log_to_disk.c
zcsahok: Yes, log lines in the log file are 88 chars long including the NL. Internally it's the buffer length is the same as there is a NUL instead of NL.

src/log_to_disk.c
zcsahok: Aha, thanks. Then this will need some work to sort out.

zcsahok:
How about this plan?

1. in this PR keep only the minimal changes to get freq into the log and also remove the code for dropping the last char of the LAN message.
2. document LAN messages in the manual
3. test to see if messages work as required and fix the arising issues. 

I can help with 2) and 3).

src/lancode.c
dl1jbe: Yes, the whole code (besides the QTCxxx messages) are ancient code (It is also in the very early 0.9.10 version from 2003). There was neither time nor need to sort it out. But let us do it now.

I do not think that it was a fix - I assume Rein had NL terminated strings in mind when writing these lines. But the whole artificial fixes of line length (lan_loglin[xx] = '\0') looks really like hot fixes, especially as all these xx values are mostly wrong.

In meantime I did some work to harmonize the LAN message format to EVERY time use of at least ONE NL (which would be dropped by the two lines in question) - see last commit. The generated messages sometimes contains two NL (even without my change). So that mess should really be sorted out.

See my second message as comment to your plan below.

dl1jbe:
See the last commit to see a more developed state after harmonizing the sent code.

But, let us discus your plan.

>     1. in this PR keep only the minimal changes to get freq into the log and also remove the code for dropping the last char of the LAN message.

Ok, rework of the commits can be made easily. I would make a new PR for it. (Maybe not before friday).
But I would still recommend to send ALL message NL terminated in the end (makes debugging more easy). So no need to drop the last char.

>     2. document LAN messages in the manual
Or maybe better in the code, e.g. in lancode.h. I think the message format is not relevant for normal user.
>     3. test to see if messages work as required and fix the arising issues.
>
> I can help with 2) and 3).

Fine by me.

zcsahok:
Thanks for the updates. I still see no need for the NL, but can live with it. You can use this PR, no separate one is needed IMO.

Re 2): lancode.h is also a perfect place to document the messages.

src/lancode.c
zcsahok: Could the removal of last char be made more relaxed using `g_strchomp()` instead?

dl1jbe:
That should be step 1. Using g_strchomp is a good idea.

zcsahok:
Thanks for the update.

zcsahok (approved):

PR #451, opened 2024-12-22
Fix point calculation for RAEM contest

dl1jbe:
Round own lattitude and longitude to whole degree as stated by contest rules

### DIFF
--- a/rules/raem.py
+++ b/rules/raem.py
@@ -27,8 +27,8 @@ def score(qso):
     lon = int(m.group(3))
     if m.group(4) == 'O':
         lon = -lon
-    points = points + abs(lat - int(tlf.MY_LAT))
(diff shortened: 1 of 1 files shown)


airween (approved):
LGTM

rules/raem.py
```diff
@@ -27,8 +27,8 @@ def score(qso):
     lon = int(m.group(3))
     if m.group(4) == 'O':
         lon = -lon
-    points = points + abs(lat - int(tlf.MY_LAT))
-    diff_long = abs(lon - int(tlf.MY_LONG))
+    points = points + abs(lat - int(round(tlf.MY_LAT)))
```
zcsahok, line 30: int() could then be dropped

dl1jbe: thanks

zcsahok (approved):

PR #452, opened 2024-12-28
Show autosend configuration after change

zcsahok:
After changing autosend configuration (`cwstart`) the usual 1 sec delay was missing, causing the information immediately cleared and thus not actually showing.

Extracted the functionality from the switch statement into a separate function.

Pressing ESC immediately exits the function without showing the current setting. Note: the original intention was to show the setting even in this case. I find it more logical not to bug the user with the 1 sec delay if they wanted to abort the change.

### DIFF
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -73,34 +73,35 @@ void wipe_display();
 static void change_autosend() {
     mvaddstr(13, 29, "Autosend: (0, 2..5, m)?");
     refreshp();
-    int x = 1;
+    int x;
 
-    /* wait for correct input or ESC */
-    while ((x != 0) && !((x >= 2) && (x <= 5)) && !(x == 'm' - '0')) {
+    /* wait for valid input or ESC */
+    do {
 	x = key_get();
 	if (x == ESCAPE)
-	    break;
-	x = x - '0';
-    }
+	    return;
+    } while (!(x == '0' || (x >= '2' && x <= '5') || x == 'm'));
 
     /* remember new setting */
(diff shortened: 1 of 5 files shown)


zcsahok:
Note2: autosend configuration is missing from logcfg.

dl1jbe (approved):

dl1jbe:
Good catch.

Wrt Note2: Not sure if it is worth to add it to logcfg.dat. But just go ahead.

zcsahok:
As it was easy I have added AUTOSEND configuration.

dl1jbe:
May I ask for the reason to switch from G_REGEX_DEFAULT/G_REGEX_MATCH_DEFAULT to 0?

Other wise looks good. 

Should we think about renaming :CHar to something like :AUTosend? Would be consistent. Maybe a as second command name for the moment and drop :CHAr in the long run?

zcsahok:
The default flags are available only from GLib 2.74 (https://docs.gtk.org/glib/flags.RegexCompileFlags.html), but even Ubuntu 22 used by our CI build has 2.72 only. The current Ubuntu version (24) ships with GLib 2.82.
We are now requiring GLib 2.40. It could be bumped to 2.74, but then the build will fail on older systems.

dl1jbe:
I see. Thanks for explanation.

zcsahok:
Regading :AUTosend - it would be more logical for sure. But as there was no actual issue reported for this PR, I'd vote for keeping it as-is for the moment.

dl1jbe (approved):

Issue #454, opened 2025-01-10
Kenwood TS-480 rig control issue.


MikePavol:
Hello,
 Let me explain a problem with connection TS-480SAT to tlf I faced.
I have two laptops with Linux on them: AntiX and Linux Light.
There are two transceivers: Kenwood TS-480SAT and Xeigu G90, as well.
There are also usb rs232 adapters of various types: FTDI, Prolific, CP2102.
G90 works without any problem with tlf, CQRLOG and rigctl on both systems.
TS-480 works OK with CQRLOG and rigctl , but when I try to change rig from 3088 to 2028 in tlf with any combinations of port speed, have no luck.
With CQRLOG TS-480 works on 9600 port speed only.
One of the laptops has both loggers installed, the other one has only tlf.
Of course, I've checked port numbers and so on every connection.
Please, help me with at least a direction to go.
Thank you in advance,
Mike, VA3CY.

dl1jbe:
Hi Mike,

I have a TS480SAT myself and did some experiments to use it with TLF
some weeks ago. If I remember correctly besides some problems about
finding the right serial cable it worked flawlessly.

It needed a straight serial line cable with RTS/CTS connected besides
RXD/TXD. But as soon as I got rigctl working with it it worked also
with TLF.

Can you give us some feedback how you configured rigctl to work with
the rig and please also the relevant menu configurations for your TS480?

I will try to retest it on my side over the next days.

73, de Tom 



 Am Fri, 10 Jan 2025 05:32:41 -0800
schrieb MikePavol ***@***.***>:

> Hello,
>  Let me explain a problem with connection TS-480SAT to tlf I faced.
> I have two laptops with Linux on them: AntiX and Linux Light.
> There are two transceivers: Kenwood TS-480SAT and Xeigu G90, as well.
> There are also usb rs232 adapters of various types: FTDI, Prolific,
> CP2102. G90 works without any problem with tlf, CQRLOG and rigctl on
> both systems. TS-480 works OK with CQRLOG and rigctl , but when I try
> to change rig from 3088 to 2028 in tlf with any combinations of port
> speed, have no luck. With CQRLOG TS-480 works on 9600 port speed
> only. One of the laptops has both loggers installed, the other one
> has only tlf. Of course, I've checked port numbers and so on every
> connection. Please, help me with at least a direction to go. Thank
> you in advance, Mike, VA3CY.
> 
> 



-- 
"Do what is needful!"
Ursula LeGuin: Earthsea
--

[...]

PR #457, opened 2025-01-22
Allow logging of hamlib debug output

dl1jbe:
* Fix display of errors during rig initialization and cleanup init_tlf_rig() function.
* Allow control of debug output verbosity via '-d lvl' switch during startup.
* Record selected Hamlib debug output into 'riglog.txt' file.

### DIFF
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -252,7 +252,7 @@ extern bool qso_once;
 extern bool leading_zeros_serial;
 extern bool ignoredupe;
 extern bool continentlist_only;
-extern bool debugflag;
+extern int debuglevel;
 extern bool trx_control;
 extern bool nopacket;
 extern bool verbose;
--- a/src/main.c
+++ b/src/main.c
@@ -88,7 +88,7 @@ int tlfcolors[8][2] = { {COLOR_BLACK, COLOR_WHITE},
     {COLOR_BLUE, COLOR_YELLOW},
     {COLOR_WHITE, COLOR_BLACK}
 };
-bool debugflag = false;
+int debuglevel = 0;
 char *editor_cmd = NULL;
 int tune_val = 0;
 int use_bandoutput = 0;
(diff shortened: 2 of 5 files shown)


zcsahok:
It would maybe useful to have a general debug log, covering not only hamlib comms, but usable also for say cluster connection or internal networking, etc.

dl1jbe:
The actual PR was triggered by the discussion in Issue #454 some days ago. Some debug output could have helped there.

But I  totally agree that a general debug logging feature is very useful. Let us use the PR as a base for discussion:

* At the moment there are two debug mechanisms: '-d' for starting rig debug and 'LANDEBUG' keyword to start recording the LAN messages between TLF nodes. Missing is a way for general debug logging. The log should optionally include also the messages from TLF_ERR/WARN/INFO output.

* Logging should be controlled by a command line switch, but there shoudl also be a possibility to start logging while TLF is already running (maybe by something like ':debug' via changeparms().

* It would be good to have a control about the verbosity level it records in the debuglog. The three levels ERROR/WARNING/INFO seems enough. General ones include all more sincere ones (so INFO would include WARN and ERROR).

* It seems to be good to have different debug levels for different message sources, e.g. it may be needed to log TLFs own mesages in detail while ignoring hamlibs output completely or log only errors there. Maybe some syntax similar to other programs may be appropriate (e.g. '-d rig:1,tlf:3,lan:0' or similar).

* I am not sure if there shoudl be a common log file for all output sources or different ones. Recording the source in the debuglog could allow an easy filtering via `grep` or similar tools.

airween:
> I would want a trace of the fldigi communication when RTTY is active. (I think sometimes tlf+fldigi get stuck in an infinite loop fighting over the frequency trying to account for the tone offset. Sometimes I can get it working, but more often I have to kill one of programs hard.)

Could you open an issue about this? And please put some details about that - eg. I suggest this issue happens in RX mode, but I'm not sure about that.

zcsahok:
Lan operations use `syslog()`:

```
2025-01-24T22:39:34.424359+00:00 xxxx tlf: LAN: gethostbyname failed
2025-01-24T22:44:20.466117+00:00 xxxx tlf: open socket: to 127.0.0.1:7302
```

It's nice, but - since an unprivileged user can't read /var/log/syslog - near to useless.

dl1jbe:
In course of the discussed debug interface that `syslog()` usage should be migrated to the new debug interface.

dl1jbe:
I have split out the fixes for hamlib error display into PR #460. Lets concentrate onto the debug log here.

dl1jbe:
See PR #468for the debug log part.

dl1jbe:
Closing as PR #468 got merged in.

PR #458, opened 2025-01-22
Follow rig mode (cont'd)

zcsahok:
Continuing from #404 that seems to get stuck.

### DIFF
--- a/share/logcfg.dat
+++ b/share/logcfg.dat
@@ -129,14 +129,13 @@ CWTONE=800
 #           for ten tec OMNI VI #
 #################################
 #
-#SYNC_RIG_MODE
-#
 #RADIO_CONTROL
 #RIGMODEL=351
 #RIGSPEED=2400
 #RIGPORT=/dev/ttyS0
 #RIGPORT=/dev/ttyUSB2
 #RIGPTT
+#RIG_MODE_SYNC=OFF
 #
 #SSBMODE
 #
--- a/share/logcfg.dat
+++ b/share/logcfg.dat
@@ -129,6 +129,8 @@ CWTONE=800
 #           for ten tec OMNI VI #
 #################################
 #
+#FOLLOW_MODE
+#
 #RADIO_CONTROL
 #RIGMODEL=351
 #RIGSPEED=2400
--- a/share/logcfg.dat
+++ b/share/logcfg.dat
@@ -129,7 +129,7 @@ CWTONE=800
 #           for ten tec OMNI VI #
 #################################
 #
-#FOLLOW_MODE
+#SYNC_RIG_MODE
 #
 #RADIO_CONTROL
 #RIGMODEL=351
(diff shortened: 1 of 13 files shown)


share/logcfg.dat
```diff
@@ -129,6 +129,8 @@ CWTONE=800
 #           for ten tec OMNI VI #
 #################################
 #
+#FOLLOW_MODE
```
airween, line 132: I suggest a different keyword here, eg. `RIG_FOLLOW_MODE` - what do you think?

zcsahok: Let's wait till the end. ;-)

zcsahok:
`trxmode` is now checked and shown by the background process, the same way it is done for rig freq.
(I was just wondering if a mutex would be needed for curses operations too, but as it works now, let's thinks about this later)

`poll_rig_state()` polls rig mode also if needed. Then comes the tricky part: setting of `trxmode` based on rig mode is not always possible as for DIG the rig can be practically in any mode, maybe except CW/CWR.
So the logic is:
* if TLF is in CW and the rig is in LSB/USB then switch TLF to SSB
* if TLF in _not_ in CW but the rig is in CW then switch TLF to CW.

Other combinations result in no action.

...to be contd

zcsahok:
Now TLF<->rig mode sync works both ways.

Next is to configurably disable it.

zcsahok:
For the configuration my naming suggestion is `SYNC_RIG_MODE`. By default it's on: TLF and rig mode are kept in sync. If it's disabled then the two modes can be set completely independently.

share/logcfg.dat
```diff
@@ -129,6 +129,8 @@ CWTONE=800
 #           for ten tec OMNI VI #
 #################################
 #
+#SYNC_RIG_MODE
```
dl1jbe, line 132: Maybe we should do it as
`#SYNC_RIG_MODE=OFF`

`ON` is default state. So why should we remove the leading `#` to just turn it on a second time.

dl1jbe: A second though: How about calling it RIG_SYNC_MODE just to have it similar to other rig related commends.

airween: > A second though: How about calling it RIG_SYNC_MODE just to have it similar to other rig related commends.

Agree (see my [previous](https://github.com/Tlf/tlf/pull/458#discussion_r1925862247) comment).

zcsahok: How about `RIG_MODE_SYNC` ? As it is more about synchronizing rig mode than the mode of "rig sync" as `RIG_SYNC_MODE` would suggest.

dl1jbe: Both sound good for me.

airween (approved):

dl1jbe (approved):

dl1jbe:
Just saw, that by accident we picked up a change in README.md which names github as main download source.
We should discuss that and if we want to keep that change, we would need to drop the project pages from savannah.

zcsahok:
I have no preference.
But looking at http://savannah.nongnu.org/projects/tlf/ I see that the latest news are from 2003 and it points a CVS repo. There is even a bug from 2004 (I hope it has been fixed in the meantime...). At least Group Homepage leads to github.

PR #460, opened 2025-01-25
Fix display of hamlib errors during initialization

dl1jbe:
Move output of hamlib error messages from footer into flow of text output. Fix problems with changed behavior of hamlibs `rigerror()` function.

Furthermore cleanup some useless code


zcsahok (approved):
Looks good

PR #461, opened 2025-02-08
Add NODE_X syntax to specify network nodes

zcsahok:
As discussed in #444 this PR introduces a new way of specifying network nodes. 
Originally one had to ensure that own node is _not_ added using ADDNODE. As nodes were anonymous there was no easy way to tell which node has issues when checking the :INFo screen.

The NODE_X syntax is an alternative to ADDNODE. If fact the two are exclusive: either only ADDNODEs (the old way) or only NODE_X statements can be used. Mixing them results in an error message.

A typical setup would look like
```
NODE_A=biggun1
NODE_B=biggun2
NODE_C=rxonly
THISNODE=B
```

An example :INFo screen with all 8 nodes (A-H) defined. Note that to packets are send to our own node (C).
 image

The value of `using_named_nodes` controls whether the original (false) or the new approach (true) is used.
It is derived during loading logcf.dat.
The named nodes occupy an entry in `bc_hostaddress[]`corresponding to their node name. The node ids doesn't have to be continuous, there can be missing nodes.

Although ADDNODE still works as before the man page documents NODE_X only in order not to create too much confusion and to favor the usage of the new syntax.

### DIFF
--- a/share/logcfg.dat
+++ b/share/logcfg.dat
@@ -68,8 +68,8 @@ TIME_MASTER
 #################################
 #  use 'addnode' only for OTHER nodes !!
 #
-#ADDNODE=10.0.0.115
-#ADDNODE=192.168.1.2
+#NODE_B=10.0.0.115
+#NODE_C=192.168.1.2
 #
 THISNODE=A
 #
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -689,38 +689,56 @@ void networkinfo(void) {
 
     wipe_display();
 
-    if (lan_active)
+    int n_nodes = 0;
+
+    if (lan_active) {
 	mvprintw(1, 10, "Network status: on   Node: %c", thisnode);
-    else
+	mvprintw(3, 32, "Total packets rcvd: %d | %d", recv_packets, recv_error);
+
+	for (int i = 0; i < nodes; i++) {
+	    if (*bc_hostaddress[i] == 0) {
+		continue;
+	    }
+	    GString *info = g_string_new(NULL);
+	    int column = 10;
+	    if (using_named_nodes) {
+		char *id = g_strdup_printf("%c%c) ",
+					   (i == thisnode - 'A' ? '*' : ' '),
+					   'A' + i
+					  );
+		g_string_append(info, id);
+		g_free(id);
+		column -= 4;
+	    }
+	    g_string_append(info, bc_hostaddress[i]);
+	    if (*bc_hostservice[i]) {
+		g_string_append(info, ":");
+		g_string_append(info, bc_hostservice[i]);
+	    }
+	    char *s = g_string_free(info, FALSE);
+	    mvaddstr(4 + n_nodes, column, s);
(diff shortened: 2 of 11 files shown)


zcsahok:
Removed support of service names. Original code allowed port to be looked up from /etc/services. To achieve this there was some extra juggling needed: parallel tracking of `default_lan_service` and `lan_port` and converting them in both directions.
As it is highly unlikely that one ever will specify a service name instead a numeric port, we can drop this code.

share/logcfg.dat
```diff
@@ -68,8 +68,8 @@ TIME_MASTER
 #################################
 #  use 'addnode' only for OTHER nodes !!
 #
-#ADDNODE=10.0.0.115
-#ADDNODE=192.168.1.2
+#NODE_B=10.0.0.115
+#NODE_C=192.168.1.2
```
dl1jbe, line 72: Do we need a NODE_A entry for our-self?

src/changepars.c
```diff
@@ -689,38 +689,57 @@ void networkinfo(void) {
...
+    mvprintw(6 + n_nodes, 10, "Config file: %s", config_file);
+    mvprintw(7 + n_nodes, 10, "Contest    : %s", whichcontest);
+    mvprintw(8 + n_nodes, 10, "Logfile    : %s", logfile);
 
-    mvprintw(13 + nodes, 10, "callmaster : %s",
+    mvprintw(9 + n_nodes, 10, "Cluster    : %s", pr_hostaddress);
+    mvprintw(10 + n_nodes, 10, "TNCport    : %s", tncportname);
+    mvprintw(11 + n_nodes, 10, "RIGport    : %s", rigportname);
+    mvprintw(12 + n_nodes, 10, "Band output: %s",
+	     (use_bandoutput ? "on" : "off"));
```
dl1jbe, line 739: `use_bandoutput` could be moved to bool

dl1jbe (commented):
Otherwise looks good.

share/logcfg.dat
dl1jbe: My concerns were about the display of network data for our own node in `networkinfo`. As far as I see not having a node for our-self leads to the according bc_hostaddress not set and that leads to `networkinfo`  not showing data for our node. 
(Not sure if I am correct here, just glancing over the code).

dl1jbe: Did some more reading. Network data for our own node stays always 0. So no real need to show them. Feel free to ignore my comment above.

zcsahok: Yes, we show the sent packets per node, so our will always be zero.
Normally all others shall be equal unless there were errors, as we broadcast to all nodes.
Regarding the errors: in a typical setup (i.e. hosts could be looked up) I see no real possibility of UDP sending failing to a particular node. Sending works even if the node is down. So maybe only if the network/interface down. Anyway, these stats doesn't harm even if they are not too useful.

dl1jbe (approved):

PR #469, opened 2025-02-16
Fix crash at startup on terminal resize

zcsahok:
To reproduce the issue

1.  start TLF in a resizable terminal with size smaller than 22x80
2. at the "Continue anyway?" prompt resize the terminal
3. get a segfault
image

The issue is caused by `resize_layout()` trying to access not yet initialized variables (rst arrays).

Solved by introducing variable `tlf_initialized` that is set to true only when all has been set up, just before calling `logit()`.
While `tlf_initialized` is false, resizing now just stores the terminal size.

A related issue has also popped up: resizing "keypress" is treated as an actual one, causing misfunction at some "Y/N" or "press any key" prompts.
Terminal resize is now not propagated to the caller of `key_get()`/`key_poll()`.

### DIFF
--- a/src/globalvars.h
+++ b/src/globalvars.h
@@ -169,6 +169,7 @@ extern bool demode;
 extern bool ctcomp;
 extern bool show_time;
 extern bool use_rxvt;
+extern bool tlf_initialized;
 extern bool time_master;
 extern bool noautocq;
 extern bool no_arrows;
--- a/src/main.c
+++ b/src/main.c
@@ -78,6 +78,7 @@ int portnum = 0;
 
 bool use_rxvt = false;
 bool use_xterm = false;
+bool tlf_initialized = false;   // true: everything has been set up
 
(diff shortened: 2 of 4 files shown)


dl1jbe:
Makes sense.
But it seems to someway conflict with the packet screen, especially with the login procedure.
To reproduce unset `CLUSTERLOGIN` in logcfg.dat an start with packet support. Before entering your password resize the window. Login fails and cluster interface gets deactivated.

zcsahok:
I got no issue with resize at the cluster login prompt. All worked as without resize.

zcsahok:
Even the original code can cope with resize at cluster login.

dl1jbe:
> I got no issue with resize at the cluster login prompt. All worked as without resize.

Ok, I will dig in what happens here. I would say feel free to merge your PR in in meantime. If its reproducible here I will add another PR later.

> Even the original code can cope with resize at cluster login. Interesting.

Yes that worked all the time.

PR #470, opened 2025-02-19
fix: set correct argument of prototypes

airween:
There is an [FTBFS](https://wiki.debian.org/FTBFS) bug in Debian, see [#1098006](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098006) with [gcc-15](https://gcc.gnu.org/gcc-15/porting_to.html).

This PR solves the problems.

### DIFF
--- a/src/focm.c
+++ b/src/focm.c
@@ -30,10 +30,8 @@
 #include "tlf_curses.h"
 #include "ui_utils.h"
 #include "bands.h"
+#include "setcontest.h"
 
-bool no_multi(spot *s) {
-    return false;
-}
 
 struct pos {
     int column;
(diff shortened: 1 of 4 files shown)


src/focm.c
```diff
@@ -31,7 +31,7 @@
 #include "ui_utils.h"
 #include "bands.h"
 
-bool no_multi() {
+bool no_multi(spot *s) {
```
zcsahok, line 34: `setcontest.c` has a `static bool no_multi(spot *data)`, let make it accessible and use here too. just to avoid duplication.

airween: Did you think something is like in 0ee226e6?

zcsahok: exactly

zcsahok (approved):

PR #459, opened 2025-01-24
Minor fixes

dl1jbe:
Running scan_build from LLVM showed a number of problems. The PR provides fixes for a relevant part of it.

PR #472, opened 2025-02-23
Fix search panel size

zcsahok:
#459 brought up a hidden issue: in order to everything work correctly search panel was initialized twice. First to ensure that the startup messages are shown and then according to the actual CONTEST_MODE configuration.

The second initialization has been suppressed, so we were left with the 9-band non-contest search panel.
image
And there is no way to get it reduced to 6. :CONtest command merely switches `iscontest`, doesn't resize the panel.

The solution is to drop the early initialization and call ncurses' plain `refresh()` in showmsg.c. There are no panels at this point yet, so no use calling out `refreshp()` (=update_panel() + co).

Getting back to :CONtest - it still can't switch panel size according to `iscontest`.
My view is that this feature is not needed: one shall not switch contest mode while already running TLF. This is a task to be done when configuring a contest or simple QSO setup, _before_ starting TLF.

Taking it one step further, :CFG/:SET shall also be removed (see the long runner #233). As I see we have practically no chance to cleanly restart the program as there is a number of allocated data structures, TCP and serial connections, etc. to be (really) correctly freed and set up anew. User shall simply restart TLF on such configuration changes. It takes just 1 second. (ok, plus the 3 seconds cluster wait, but that's another story...)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -559,9 +559,6 @@ static void ui_init() {
 	exit(EXIT_FAILURE);
     }
 
-    InitSearchPanel();	/* at least one panel has to be defined
-				   for refreshp() to work */
-
     getmaxyx(stdscr, ymax, xmax);
     if ((ymax < 22) || (xmax < 80)) {
 	char c;
--- a/src/showmsg.c
+++ b/src/showmsg.c
@@ -30,7 +30,7 @@ static int linectr = 0;
 void clearmsg() {
     clear();
     linectr = 0;
-    refreshp();
+    refresh();
 }
 
 
@@ -41,7 +41,6 @@ void clearmsg_wait(void) {
 	mvaddstr(LINES - 2, 0, "Press any key to continue!");
 	move(LINES - 1, 0);
 	clrtoeol();
-	refreshp();
 	IGNORE(key_get());
     } else {
 	sleep(1);
@@ -50,11 +49,11 @@ void clearmsg_wait(void) {
 }
 
 
-static int has_room_for_message() {
+static bool has_room_for_message() {
     if (linectr < LINES - 3)
-	return 1;
+	return true;
     else
-	return 0;
+	return false;
 }
 
 void show_formatted(char *fmt, ...) {
@@ -68,7 +67,7 @@ void show_formatted(char *fmt, ...) {
     va_end(args);
     mvaddstr(linectr, 0, str);
     g_free(str);
-    refreshp();
+    refresh();
     linectr++;
 }
 
--- a/src/showmsg.h
+++ b/src/showmsg.h
@@ -21,6 +21,11 @@
 #ifndef STARTMSG_H
 #define STARTMSG_H
 
+/*
+ * These functions provide progress and error reporting at start-up phase
+ * !!! Do not call them after switching to the main screen !!!
+ * For runtime notifications use TLF_SHOW_* macros instead
+ */
 
 void clearmsg(void);
 void clearmsg_wait(void);


dl1jbe:
> #459 brought up a hidden issue: in order to everything work correctly search panel was initialized twice. First to ensure that the startup messages are shown and then according to the actual CONTEST_MODE configuration.
> 
> The second initialization has been suppressed, so we were left with the 9-band non-contest search panel. 
And there is no way to get it reduced to 6. :CONtest command merely switches `iscontest`, doesn't resize the panel.
> 
I see the problem. 

> The solution is to drop the early initialization and call ncurses' plain `refresh()` in showmsg.c. There are no panels at this point yet, so no use calling out `refreshp()` (=update_panel() + co).
> 
That will work as long as anybody uses `showmsg()` and relatives only during initialization. Maybe we should enforce that or add at least a comment on the `showmsg` code stating the problem.

Other ways would be to allow resizing of the searchwindow, having a dummy panel or using an early init for the keyer panel. But I assume it is not worth the work here.


> Getting back to :CONtest - it still can't switch panel size according to `iscontest`. 
>
That would need a `resize_panel()` for the new window size. But I agree, we should drop the :CONtest command at all.

---
 
> Taking it one step further, :CFG/:SET shall also be removed (see the long runner #233). As I see we have practically no chance to cleanly restart the program as there is a number of allocated data structures, TCP and serial connections, etc. to be (really) correctly freed and set up anew. User shall simply restart TLF on such configuration changes. It takes just 1 second. (ok, plus the 3 seconds cluster wait, but that's another story...)

+1 from my side.

zcsahok:
I will add a comment to `showmsg.c` that these functions shall be using in the setup phase only. They are used like this, so all is fine.

As far as I understand `showmsg` functions are meant to output to the default screen that is then replaced by a new window in `main.c`.
What is kind of interesting is that `mainscreen` is created in `ui_init()`, but actually activated via `set_term` in the seemingly unrelated `packet_init()` function. I have the strong suspicion that the code originated a long time ago from a telnet cluster client...

`changepars.c` needs a rework anyway (getting rid of the mega-switch with hardcoded indexes), will do the dropping in that round.

dl1jbe:
> I will add a comment to `showmsg.c` that these functions shall be using in the setup phase only. They are used like this, so all is fine.
> 
Just merge it in afterwards.

> As far as I understand `showmsg` functions are meant to output to the default screen that is then replaced by a new window in `main.c`. What is kind of interesting is that `mainscreen` is created in `ui_init()`, but actually activated via `set_term` in the seemingly unrelated `packet_init()` function. I have the strong suspicion that the code originated a long time ago from a telnet cluster client...
> 
Yes that seems to be the case - just look at the first lines of `splitscreen.c`. `VERSIONSPLIT` tells you some of the ancestry. See also the complete handling of the packet interface (addtext, history navigation, ....).

At the time I picked up TLF (around '09) there were three terminals - the default one used during startup, a 'packetscreen` for packet interface  and a second `mainscreen` used for TLFs main ui. All overlay windows (search, score, keyer, ...) were made programmatically by redrawing hard coded parts of the main ui. That resulted in a lot of problems:

- difficult to change layout
- no chance to resize the screens during runtime.

One of the reasons were that 'newterm' was not designed to handle different screen layouts on the same physical display but to switch between different physical consoles. 

So I dropped the second `packetscreen` and switched packet interface and other overlaying windows to ncurses panels. Maybe it is possible to drop the `mainscreen` also - needs some experiments.

> `changepars.c` needs a rework anyway (getting rid of the mega-switch with hardcoded indexes), will do the dropping in that round.

Ok. Sounds sensible.

PR #474, opened 2025-06-18
Fix editing note (comment) line

zcsahok:
Editing a note (or comment) line is not really possible as editing assumes and enforces QSO-line structure.

Already entering a note can lead to strange results.
image
The two ^[ symbols are the escape keys the desperate user was trying to use to close the note input window.
The first chars of the input line are covered due the longer line.

When trying to fix the note one runs into a deadlock. The only way out is to kill TLF.
image

The fix introduces specific field definition for comment lines. When switching between different types of lines the cursor is moved to a valid position.

Also fixed a minor issue: Ctrl-A/E was not updating `field_index`, only the cursor position.


zcsahok:
Note input uses curses' own `getnstr()` function, so handling esc and other non-printables can't be changed.
What we can do is to clean up the user input: replace non-printables with space and trim leading/trailing whitespace.
As a result if user types just spaces (or escapes) no note will be included.

zcsahok:
Looks like Ubuntu 20 support is removed: 

> Ubuntu 20.04 LTS runner will be removed on 2025-04-15.

https://github.com/Tlf/tlf/actions/runs/15743352324/job/44374380014?pr=474

zcsahok:
How about this: on a new line the cursor is moved left to the rightmost non-space position. If there are chars right from the original cursor position then cursor is not touched.

Some paths going up and down between the lines:
image

dl1jbe:
That looks good and should do for most cases. Thanks.
+1

dl1jbe (approved):

dl1jbe:
:+1:

PR #475, opened 2025-06-19
Switch from Ubuntu 20 to 22

zcsahok:
Ubuntu 20 support has been removed. Current "latest" version is 24, so use 22 as "old" version for CI build.

https://github.com/actions/runner-images


zcsahok:
removed unused Travis file

PR #476, opened 2025-08-13
Fix memory leak and accessing unitialized data

zcsahok:
Had been playing around with valgrind.
Fixing memory leak in remember_generic_mult(): `word` and `mult` were not free'd due to early exit from the loop.

### DIFF
--- a/src/lancode.c
+++ b/src/lancode.c
@@ -150,7 +150,7 @@ int lan_recv(void) {
 
     errno = 0;			/* clear the error */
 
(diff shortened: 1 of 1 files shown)


zcsahok:
The amount of received data was not checked before evaluating the 2nd byte in `lan_recv()`.

dl1jbe (approved):

zcsahok:
Thanks for the feedback. I find keeping free'ing localized somewhat better maintainable.

PR #477, opened 2025-08-13
Fix python regexps

zcsahok:
Current python version (3.13) is more strict about escape codes. The _SyntaxWarning: invalid escape sequence ..._ message is fixed by properly defining regexps as raw strings (r"").

To trigger the warning:
```
$ python3
Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print('\sa\db\w')
<python-input-0>:1: SyntaxWarning: invalid escape sequence '\s'
\sa\db\w
```
The string is interpreted correctly (at least for the moment).
see also https://stackoverflow.com/questions/79275524/invalid-escape-sequences-in-python-3-12-in-a-docker-image

### DIFF
--- a/rules/cqww160/cqww160.py
+++ b/rules/cqww160/cqww160.py
@@ -28,7 +28,7 @@ def init(cfg):
 
     # read initial exchange file
     global K_VE_EXCHANGES
-    comment = re.compile("(^#|^$)") # starts with a hash or empty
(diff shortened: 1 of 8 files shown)


dl1jbe (approved):
+1

airween (approved):

PR #478, opened 2025-08-29
Remove INSTALL file from Git

df7cb:
The autoconf toolchain really likes to replace INSTALL with its own generic installation instructions, causing noise in the `git diff` output. Since the current file doesn't really have content, just remove it and add INSTALL to .gitignore.

This has been driving me crazy ever since Debian started automatic autoreconf'ing during package builds.

Plain `autoreconf -i` doesn't trigger it, but I do run into it all the time... please consider accepting this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ config.status
 config.sub
 configure
 depcomp
+INSTALL
 install-sh
 missing
 src/.deps
--- a/INSTALL
+++ b/INSTALL
@@ -1,3 +1,0 @@
-INSTALL
-
-Please see README.md for installation instructions.


airween:
I'm fine with this, but we can consider to add a `build.sh` script which runs all necessary `auto*` commands, something like this:
```bash

rm -rf autom4te.cache
rm -f aclocal.m4

case `uname` in Darwin*) glibtoolize --force --copy ;;
  *) libtoolize --force --copy ;; esac

autoreconf --install
autoheader
automake --add-missing --foreign --copy --force-missing
autoconf --force
rm -rf autom4te.cache
```

df7cb:
Afaict, `autoreconf -i` should be enough, so making a shell script for it might be overkill.

That's also what's documented in README (perhaps in a weird order, though).

N0NB:
If INSTALL doesn't exist, then one of the Autotools will complain, as I recall as having it is a requirement of the GNU standard, if that matters to anyone.

Christoph, do you have the same issue with Hamlib and its INSTALL file?

zcsahok:
I guess this is the issue -- on a fresh clone 
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean

$ autoreconf --install --force
configure.ac:17: installing './compile'
configure.ac:48: installing './config.guess'
configure.ac:48: installing './config.sub'
configure.ac:7: installing './install-sh'
configure.ac:7: installing './missing'
src/Makefile.am: installing './depcomp'
parallel-tests: installing './test-driver'

$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   INSTALL

no changes added to commit (use "git add" and/or "git commit -a")

$ git diff
diff --git a/INSTALL b/INSTALL
index cd9fb18..d3b6e4e 100644
--- a/INSTALL
+++ b/INSTALL
@@ -1,3 +1,365 @@
-INSTALL
+Installation Instructions
+*************************
 
-Please see README.md for installation instructions.
\ No newline at end of file
+Basic Installation
+==================
+
+   The following shell commands:
...
```

zcsahok:
I don't recall why I added --force to my local build script. Maybe it was needed when adding test files, not sure.
A normal build in fact works fine with just --install, as described in README.md.
If INSTALL would add extra information regarding installation, then I'd suggest renaming it to INSTALL.md. As it does not, I could live without it.
(in my build script I use `git checkout INSTALL` to get back the original file, so never had to bother with it)

N0NB:
Another option is to add a `bootsrap` script (borrow the one from Hamlib!) that simply checks for the presence of Autoconf and Automake and then runs `autoreconf -i`.

I wonder if the Debian tooling, in the absence of a `bootstrap[.sh]` or `autogen[.sh]` (the less preferred name), simply runs `autoreconf --install --force` as a means to make packages that comply with the GNU packaging standard and thus lessen the number of packaging bugs generated.  There are probably a number of upstream packages that lack all the requisite files.

zcsahok:
@dl1jbe would renaming INSTALL into say INSTALLING be an option?

dl1jbe:
> @dl1jbe would renaming INSTALL into say INSTALLING be an option?

In that case running autoreconf will generate an INSTALL file with a standard text. There is a lot into these text that is not really related to TLF (and maybe misleading) but at least it redirects to a README file for more information. While we do not have such a file I think the README.md will be recognized as similar.

If we can accept that standard INSTALL as not misleading we can drop the INSTALL file without renaming.
Please have a critical look at the generated file.

zcsahok:
The content of automake's INSTALL file is generic (it originates from /usr/share/automake-1.xx/INSTALL), but could serve as a starting point:
```
Installation Instructions
*************************

Basic Installation
==================

   The following shell commands:

     test -f configure || ./bootstrap
     ./configure
     make
     make install

should configure, build, and install this package.  The first line,
which bootstraps, is intended for developers; when building from
distribution tarballs it does nothing and can be skipped.

   The following more-detailed instructions are generic; see the
‘README’ file for instructions specific to this package.  Some packages
provide this ‘INSTALL’ file but do not implement all of the features
documented below.  The lack of an optional feature in a given package is
not necessarily a bug.  More recommendations for GNU packages can be
found in the GNU Coding Standards.
...
```

Execution of the listed shell commands actually works and produces an executable. Of course `make install` fails, but it is fine as a starting point.
The text says, that it's generic and refers to the `README` file. We have a `README.md`, but that should be OK in my view. So the first couple sentences lead to user to the right place. Then comes a lengthy description of the tooling, that hardly anybody reads. But that's also OK, as it stand for generic reference.

dl1jbe:
I agree that the provided text covers most of our needs. What stays open are build dependencies and additional configure switches. That gets discussed in README.md later but there may be some failed build attempts in meantime. 

> Execution of the listed shell commands actually works and produces an executable. Of course make install > fails, but it is fine as a starting point.

The need to have root privileges for 'make install' is discussed little bit further down the text in point 6.

PR #480, opened 2025-09-17
Remove references to internal Hamlib data

GeoBaltz:
Hamlib 5.0 (late 2026) will rearrange its internal data layout, removing the structures-within-structures that have caused many past problems.

This removes the references to the internal port, and substitutes calls to the preferred methods of setting config parameters.

The new code also verifies that the speed is valid.

### DIFF
--- a/src/sendqrg.c
+++ b/src/sendqrg.c
@@ -120,8 +120,8 @@ int init_tlf_rig(void) {
 			   rigportname);
 
     if (retcode != RIG_OK) {
-        showmsg("Pathname not accepted!");
-        return -1;
+	showmsg("Pathname not accepted!");
+	return -1;
     }
 
     snprintf(speed_string, sizeof speed_string, "%d", serial_rate);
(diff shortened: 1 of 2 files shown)


src/sendqrg.c
```diff
@@ -115,10 +116,22 @@ int init_tlf_rig(void) {
     }
 
     g_strchomp(rigportname);	// remove trailing '\n'
-    strncpy(my_rig->state.rigport.pathname, rigportname,
-	    TLFFILPATHLEN - 1);
```
zcsahok, line 119: TLFFILPATHLEN can be dropped as it was only used here

zcsahok:
A question: are other constructs like
```
my_rig->caps->level_gran[...]
```
also affected?

GeoBaltz:
I have a test branch of hamlib that deprecates everything that is moving (as of now).  With that installed, tlf with this patch compiles cleanly.

73 n3gb

zcsahok:
Thanks!

PR #482, opened 2025-09-26
Fix tests for >=cmocka-1.1.8

dl1jbe:
In 1.1.8 CMocka changed the list of required includes for tests. Without the patch it will complain about missing definitions for 'uintptr_t'.

### DIFF
--- a/test/test.h
+++ b/test/test.h
@@ -9,6 +9,7 @@
 #include <unistd.h>
 #include <stdarg.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <setjmp.h>
 #include <string.h>
 #include <stdbool.h>


airween (approved):

airween:
Good catch. Probably we should upgrade the workflow environment to Ubuntu 24.04 (even the cmocka is still 1.1.7 [there](https://packages.ubuntu.com/search?suite=noble&section=all&arch=any&keywords=libcmocka&searchon=names)).

zcsahok:
Thanks.
The default CI build runs now on Ubuntu 24.04.3 LTS.

airween:
> The default CI build runs now on Ubuntu 24.04.3 LTS.

Ah, sorry, now I see the other workflow file. Thanks!

PR #483, opened 2025-09-26
Disable man-db update triggers in workflows

zcsahok:
apt update in workflows triggers a potentially lengthy and completely useless man.db update.

A good run -- update was quick:
image

A bad run -- took twice as much time
image

The fix disables man-db update upon installing packages.
(see also https://github.com/actions/runner-images/issues/10977)

### DIFF
--- a/.github/workflows/ci-build-ubuntu-22.yml
+++ b/.github/workflows/ci-build-ubuntu-22.yml
@@ -21,6 +21,10 @@ jobs:
       run: cat /etc/os-release
     - name: Install dependencies
       run: |
+        # first disable man-db update triggers
+        echo "set man-db/auto-update false" | sudo debconf-communicate
+        sudo dpkg-reconfigure man-db
+        #
         sudo apt-get -qq update
         sudo apt-get install -y libhamlib-dev libxmlrpc-core-c3-dev libglib2.0-dev libcmocka-dev python3-pexpect python3-dev astyle
     - name: Set up datadir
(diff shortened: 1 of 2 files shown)


airween (approved):

N0NB:
Good idea!  I think I'll borrow that for the Hamlib CI workflow.

zcsahok:
@N0NB I had a similar requirement in TLF: show test failures but only if tests were actually failed.
Here's what I did ('id+'continue-on-error' and the 'if' line)
```
    - name: Run tests
      id: run_tests
      continue-on-error: true
      run: make check
    - name: Show failed test results
      if: ${{ steps.run_tests.outcome == 'failure' }}
      run: |
        for f in test/run_*.log; do
            grep -q FAILED $f && echo -e "\n>>> $f:\n" && cat $f
        done
        exit 1
```

For Hamlib it would be something like
```
    - name: make distcheck
      id: make_distcheck
      continue-on-error: true
      run: make distcheck ...
    - name: 'tar files'
      if: ${{ steps.make_distcheck.outcome == 'failure' }}
      run: tar -cvf ubuntu_latest_features.tar $(pwd)
    - name: 'Upload failed Build Artifact'
      if: ${{ steps.make_distcheck.outcome == 'failure' }}
      uses: actions/upload-artifact@v4.6.2
      with:
        name: 'Ubuntu_latest_features'
        path: ubuntu_latest_features.tar
```

PR #485, opened 2025-10-16
add RECALL_NO_SERIALS to disable recalling serial numbers

### LINKED ISSUE #356: INITIAL_EXCHANGE handling
INITIAL_EXCHANGE is a nice thing for contests with a predefined exchange. It is tied together with RECALL_MULTS. This is a problem in a contest like WAG (DL side), because there are not only DOK's from the multi file are exchanged but also serial numbers from non-DL stations which should not get recalled. So both commands should not be coupled together.

zcsahok:
Aiming to resolve #356 

Added **RECALL_NO_SERIALS** to control whether serial numbers from previous QSOs should be recalled.
Recalling is still enabled by RECALL_MULTS.

The logic is (provided RECALL_MULTS is active):

- RECALL_NO_SERIALS unset/false (default): exchanges are looked up first in the log, then in initial exchange (IE) list. No checks are done.
- RECALL_NO_SERIALS set/true: exchanges from the log that look like a serial number (all digits) are not considered. Values from IE are taken regardless if they are numbers or not.

This way errors from IE list are corrected in both cases as over-the-air value from the log takes precedence. The basic assumption is that contests doesn't use all-digit section exchanges.
Will update man page if this solution is generally OK.

(contest booleans have false default value, hence the double negation)

### DIFF
--- a/tlf.1.in
+++ b/tlf.1.in
@@ -3075,7 +3075,20 @@ Multiplier is DXCC country or section from multiplier file.
 .B RECALL_MULTS
 Exchange can be recycled, will be filled into exchange field when it is known
 (see also
-.BR INITIAL_EXCHANGE ).
+.BR INITIAL_EXCHANGE
+and
+.BR RECALL_NUMERIC_EXCHANGES ).
+.
+.TP
+\fBRECALL_NUMERIC_EXCHANGES\fR[=<\fION\fR|\fIOFF\fR>]
+Controls whether all-numeric exchanges (like serial or zone number) shall be
+recalled when using
+.BR RECALL_MULTS .
+Default is \fION\fR, meaning that numeric exchanges are also recalled.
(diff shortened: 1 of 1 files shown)


df7cb:
Sorry for the continued bikeshedding. :slightly_smiling_face: 

It feels weird to have to set `RECALL_SERIALS=on` for a contest like CQWW where the exchange isn't a serial. (Even when the default is on and you'd only disable it for other contests.)

Could we name the option `RECALL_NUMERIC_EXCH` or similar? Then it's immediately clear what it does.

zcsahok:
Ok, let's take the more explicit `RECALL_NUMERIC_EXCHANGES`.
So for WAG the configuration would be

INITIAL_EXCHANGE = wag.txt
RECALL_MULTS
RECALL_NUMERIC_EXCHANGES = NO

Issue #463, opened 2025-02-11
TLF and fldigi fighting over frequency in RTTY mode


df7cb:
Over the past year, I've been trying to do some RTTY contests with TLF. I got it working occasionally, but most often, things wouldn't initialize correctly and the fldigi waterfall (and sometimes the whole UI) would freeze. Tuning the VFO around would show that the actual rig frequency is always jumping back to the original value after a fraction of a second. At one point I had a sequence of "start X, start Y, switch mode, ..." that made it work, but it was very fragile and any problem during operation (rig communication hickup something) would trigger the problem again. Once the problem happens, even quitting TLF will leave fldigi frozen for some more time. I put in some printf to see the communication between the two programs and it looks like fldigi is flooded with requests that it doesn't have the time to digest.

What I think happens around the rig frequency being locked is that both TLF and fldigi are connected to the rig via rigctld, so when RTTY mode is initially activated and TLF tries to communicate the frequency offset to both the rig and fldigi, fldigi will also see a frequency change in the rig which TLF can then read back and communicate to the rig again... etc. If you manage to get past that initialization, it works, until things get out of equilibrium again. (I can't really disconnect fldigi from rigctld because it needs to key PTT.)

Last Friday, just before the WPX RTTY contest I spent the evening reading the TLF source to see if I can fix the problem. Originally I thought disabling the "when RTTY is activated, change the frequency by CENTER_FREQ" logic in fldigixmlrpc.c, but this had little or no effect.

What made things work was to remove the two `rig.set_mode` and `rig.set_frequency` calls at the end of the fldigi_xmlrpc_get_carrier() function. With this, starting TLF and fldigi and entering RTTY mode worked and I could to the contest.

In summary, I think there's more than one bug or missing feature here.

* if both TLF and fldigi are connected to the rig, they shouldn't fight over the frequency
* CENTER_FREQ is hard-coded to 2210, it should be configurable (fldigi and myself prefer 1500)
* `rig.set_mode` and `rig.set_frequency` every few ms might be too much for fldigi to ingest. (The problem might be elsewhere, but removing these helped a lot)
* the code flow in fldigixmlrpc.c is a mess. The confusingly-simillarly-named variables `fldigi_var_carrier` and `fldigi_var_shift_freq` are set in non-trivial ways without any comments and are used at the very other end of the file to communicate with other parts of the program.
[...]

PR #487, opened 2025-11-01
Improve freq handling when using Fldigi

zcsahok:
Addressing the issue of TLF and Fldigi both controlling the rig frequency (#463).

The `fldigi_xmlrpc_get_carrier()` function and `gettxinfo` contained a specific logic to control the rig frequency that could interfere with what Fldigi was doing. This has been restructured in a way that frequency control in digimode-via-Fldigi case goes through Fldigi.

It's like considering Fldigi as a virtual digital rig: in CW/SSB modes we are not interested how the physical rig internally combines the frequencies, we just tell it to set a frequency and it does all required. Same here with Fldigi: we shall not care about the internals (like what rig mode is being used), just instruct it to tune to a frequency. The frequency we are interested is the _log_frequency_ in Fldigit terms.
image

Unfortunately Fldigi's interface is not symmetric for log_frequency. It can be read out, but can't be set. This is circumvented by calculating the offset on TLF side. Not the nicest thing, plus needs two more roundtrips to Fldigi, but still way cleaner and more localized that the current logic.

`fldigi_xmlrpc_get_carrier()` has been completely dropped further reducing calls to Fldigi and `gettxinfo` handles the digimode-via-Fldigi case separately.

All this of course requires that Fldigi is configured to control the rig via _rigctld_ sharing it with TLF. I believe this is just how it normally shall be used.

I'm not sure about GMFSK support. This software hasn't been updated to 10 years. I'd remove the condition marked with FIXME at getting rig mode in gettxinfo.

### DIFF
--- a/tlf.1.in
+++ b/tlf.1.in
@@ -2539,6 +2539,17 @@ If you run Fldigi's XMLRPC server on an different port use
 \fBFLDIGI\fR=\fIhttp://localhost:port_#/RPC2\fR
 .
 .TP
+\fBFLDIGI_RTTY_SWEET_SPOT\fR=\fIhz\fR
+Sets the preferred RTTY audio center frequency
+used for auto-QSY in FSK RTTY mode.
+.
+.IP
+Default is 2210, which is in the middle
+of the standard mark (2125 Hz) and space (2295 Hz) tones.
+.
+To disable auto-QSY set it to 0.
+.
+.TP
 \fBMINITEST\fR[=\fINNN\fR]
 Use this option when the contest is a minitest like contest.
 .


df7cb:
Thanks for working on this! I just gave this PR a try. I haven't yet tried to understand the new code, so the comments here are based just on the behavior observed.

* The frequency readout from fldigi is whole kHz only. The modem offset readout seems to work correctly, but the frequency displayed in tlf is always `xxxxx.0`.
* Setting the frequency from tlf (memory recall, bandmap grab) crashes the fldigi connection (`Fldigi: lost connection!`) and the frequency in tlf jumps to `0.0`. Fldigi does not change the frequency. (Fortunately, `:fldigi` reenables the connection without the need for a restart.)
* Everything else seems to work. :smiley_cat:

zcsahok:
@df7cb could you please try reading log_freqency from a terminal:
```
XMLRPC_TRACE_XML=1 xmlrpc http://localhost:7362/RPC2 log.get_frequency
```

Here is what I get
```
XML-RPC CALL:

<?xml version="1.0" encoding="UTF-8"?>\r\n
<methodCall>\r\n
<methodName>log.get_frequency</methodName>\r\n
<params>\r\n
</params>\r\n
</methodCall>\r\n

XML-RPC RESPONSE:

<?xml version="1.0"?>\r\n
<methodResponse><params><param>\r\n
\t<value>3534.765</value>\r\n
</param></params></methodResponse>\r\n

Result:

String: '3534.765'
```

that matches what I see in Fldigi
image

zcsahok:
It's the _log.get_frequency_ method that delivers the value of the field (a string) as shown to the user. On one hand it's correct, on the other hand a machine readable value (double) would be preferred.
Pushing a fix to replace comma with dot.

The old code didn't use this method. It was calling other methods (like AF carrier and rig mode) to calculate the offset.

zcsahok:
How do you start Fldigi with a different locale? I tried LC_ALL, but it didn't really work: some error messages were translated, but the UI remained in English.

df7cb:
The UI is also mostly English here even when running in de_DE.UTF-8. You should see the comma in `rig.get_frequency` to check.

zcsahok:
With de_DE.utf8 I still get a dot, but with the DE menu. Eigenartig...
image

df7cb:
> With de_DE.utf8 I still get a dot, but with the DE menu. Eigenartig... 

It's only using comma in the XML api :clown_face:

zcsahok:
@airween added "auto QSY" for hardware RTTY mode. The sweet spot can be later made configurable to match Fldigi setup
image

airween:
> added "auto QSY" for hardware RTTY mode. The sweet spot can be later made configurable to match Fldigi setup 

Probably that would be enough - unfortunately my Fldig does not have this item under that sub-tree:

kép

Which Fldigi version you have? I have 4.1.20.

airween (approved):

airween:
Looks good to me, thank you.

The only one thing that I'd like to add that we need to update the documentation. I think it's an important change that user needs to set up `rigctld` instead of Tlf "direct" built-in rig control. Also we need to mention the minimal required Fldigi version.

Thank you again Zoli!

Issue #466, opened 2025-02-12
AUTO CQ msg modification


9a3jh:
First keyboard key press will stop AUTO CQ message(F12). Now, if someone like to change keyer speed(Pg up/Pg dwn) or AUTO CQ delay(ALt Pg up/Pg dwn), this message will be interrupted.

df7cb:
This has also annoyed me in the past... I think AUTO_CQ should even keep sending while I'm trying to edit the last QSO. (Perhaps only the timer should be cancelled, but not the current message being sent.)

N0NB:
* On 2025 12 Feb 08:12 -0600, 9a3jh wrote:
> 9a3jh created an issue (Tlf/tlf#466)
> 
> First keyboard key press will stop AUTO CQ message(F12). Now, if
> someone like to change keyer speed(Pg up/Pg dwn) or AUTO CQ delay(ALt
> Pg up/Pg dwn), this message will be interrupted.

It is something I've learned to live with.  Sending other macros are not
affected by a speed change.  Perhaps the F12 behavior could be modified
to only be canceled by key characters that are valid for the call input
field (letters, numerals, and /) and <Escape>.

73, Nate, N0NB

-- 
"The optimist proclaims that we live in the best of all
possible worlds.  The pessimist fears this is true."
Web: https://www.n0nb.us
Projects: https://github.com/N0NB
GPG fingerprint: 82D6 4F6B 0E67 CD41 F689 BBA6 FB2C 5130 D55A 8819

dl1jbe:
I see the point of changing CW speed and Auto CQ-delay. That should be enabled. It goes along the idea named by Nate.

The immediate stop of transmit was made intentionally on request of some high rate contesters. Together with early start feature it allows to just start typing in a heard call to start a qso. That way there are two key presses saved (ESC to stop Auto CQ and Enter to start sending the others call sign).

PR #489, opened 2025-11-07
Improve auto CQ

zcsahok:
Auto CQ came up in two issues recently (#463 and #466).

Let's see if we can improve here.

Stop auto CQ on callsign grab (from #463): normally the callsign field is empty during CQ, so check if it has been set by a grab (clicking on Fldigi RX window). In addition to key presses consider such events also as CQ-stopper, but don't report them to the caller.

### DIFF
--- a/src/autocq.c
+++ b/src/autocq.c
@@ -162,8 +162,6 @@ int auto_cq(void) {
     cqmode = AUTO_CQ;
     show_header_line();
 
-    const long message_time = get_autocq_time();
-
     int key = NO_KEY;
 
     // any unhandled key press terminates auto CQ loop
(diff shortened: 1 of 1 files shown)


zcsahok:
Handle PgUp/Dn (keyer speed up/dn) internally (from #466): make use of `handle_common_key()` after filtering.
UI experience is not the best due to the delayed update, but one can live with it.

zcsahok:
Moved message length calculation into the loop.

We have no definite info how much of the message is left when speed changes. It could be possible to estimate it, but it's not worth the effort IMO. For small speed  changes (+/- 2 wpm) the difference is hard to notice.

PR #495, opened 2025-11-19
Make Enter and similar keys keep working when editing call and exchange

df7cb:
When editing the call field, Enter didn't anything. When editing the exchange field, Enter at least exited the edit mode, but did not log the QSO.

Make Enter, Backslash, Tab and similar keys exit editing, and perform their normal function so QSOs can be logged faster.

This also fixes the bug that editing the call did not terminate on unhandled keys (like End in that context).

### DIFF
--- a/src/calledit.c
+++ b/src/calledit.c
@@ -37,11 +37,11 @@
 #include "ui_utils.h"
 
 
-void calledit(void) {
+int calledit(void) {
 
     int i = 0, l, b;
     int j = 0;
-    int x = 0;
+    int x = -1;
     int cnt = 0, insertflg = 0;
     char call1[30], call2[10];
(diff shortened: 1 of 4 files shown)


df7cb:
Fwiw, the actual bug here was https://github.com/Tlf/tlf/pull/495/files#diff-701366522334b17251de991f054e1cb7e0d18baf25788c8e8d0ce0ab637e6216L180-R187 (plus a bit more x/i confusion), if we just merged that part, that would already fix 80% of the problem. Should I split the PR into two parts?

zcsahok:
Call input editing is addressed in #516

zcsahok:
Exchange editing is covered in #532.

PR #506, opened 2026-03-16
Prefer Country during lookup of partialcall in spot list

dl1jbe:
Alternative lookup for spot to grab. See discussion on PR #505.

This version searches first for calls starting with the searchstring giving precedence for a search_by_country. If nothing found look for the searchstring inside the spot calls.


airween:
I like the idea itself (both solution), but I think it would be great to configure which method the OP prefers. What do you think guys? Or it could be controlled through commands, like `:PARTCTY`, `:PARTBM`, or `:PARTNORM`?

dl1jbe:
Did not solve the problem, therefore I am closing it.

PR #507, opened 2026-03-18
Limit bandmap_lookup to the list of displayed spots

dl1jbe:


dl1jbe:
A quick hack to limit the search in 'bandmap_lookup()' to the displayed list of spots. May need some more love.

zcsahok:
Looks good, thanks. Not sure whether the start/stop variables need to be protected by the mutex as grabbing happens in the main thread. Also keeping 'i < spots->len' would not hurt.

dl1jbe:
> Not sure whether the start/stop variables need to be protected by the mutex as grabbing happens in the main thread. 

Yes, 'spots' array and the variables needs to stay in sync. But we can not just wrap the whole bandmap_show() by the mutex as it is already used in filter_spots() and the mutex can not be used recursive. 

I will look into the problem - just give me time till Sunday. There are some more minor changes needed in bandmap.c anyway.

dl1jbe:
Besides some refactoring of the range calculation and the final spot display it should be correct now.
The mutex covers now the built of the spots array AND the calculation of the display range.

zcsahok:
Looks good. There are just 2 typos (frequencyi, capsi; using vi?).

When we at it, could we maybe have a look at the selection of bandmap's current frequency? Not sure how it shall work, but it's asymmetric now. 

Example: 

freq=3543.5, it is marked as different from the spot's frequency (3543.7), fine:
image

freq=3543.6, we are 100 Hz off and it is "rounded up" to the spot:
image

freq=3543.7, we are spot on:
image

freq=3543.8, we are again 100 Hz off, but no rounding takes place:
image

dl1jbe:
> Looks good. There are just 2 typos (frequencyi, capsi; using vi?).

Yes, thanks - easy guess :-) .

> When we at it, could we maybe have a look at the selection of bandmap's current frequency? Not sure how it shall work, but it's asymmetric now.

Here is the situation:

- The frequency gets read (and stored) from TRX with full resolution (in Hz).
- 'TRX:' display is rounded by +/-50 Hz (3543.651...3543.750 results in TRX: 3543.7)
- Spot frequencies are stored in Hz as reported from cluster/own spotting, but displayed with one decimal (truncated).
  Cluster messages are reported with 100 Hz granularity, self spotted (Crtl-A) with 1Hz granularity.
- The criteria for 'on spot' is spot frequency +/- 100Hz.

The following drawing shows the situation

   .......|--------X--------|.....          
   ..................|------------------Y-----------------|......... 

X - displayed frequency of rig, |----X----| range the 'real' trx frequency may be in
Y - spot frequency, |-----Y-----| range accepted as 'on spot'

So depending on trx frequency rounded up or down a displayed difference from .1 kHz between TRX: and spot frequency may lead to a display 'beside spot' or 'on spot'.


The situation is more complicated for the self spotted entries. We should harmonize the situation by using also 100 Hz granularity here.

zcsahok:
I see the logic, thanks. Nevertheless there is still an issue here. The internal freq values are stored in Hz precision, so normally no rounding is involved when evaluating the bandmap.

I think here is my point:
```
    /* calc number of spots below your current QRG */
    for (i = 0; i < spots->len; i++) {
	data = g_ptr_array_index(spots, i);

	if (data->freq <= centerfrequency - TOLERANCE)   /// condition A
	    below_qrg++;
	else
	    break;
    }

    /* check if current QRG is on a spot */
    if (below_qrg < spots->len) {
	data = g_ptr_array_index(spots, below_qrg);

	if (!(data->freq > centerfrequency + TOLERANCE))  /// condition B
	    on_qrg = 1;
    }
```

Condition A says that spots below or at QRG-100 Hz are the ones "below".
Condition B says (after mentally inverting it) that the next spot is "spot on" if it's below or at QRG+100 Hz.
All else is "above".

Now, if out spot is on 7042000 Hz, then for various QRGs
7041800 --> below_qrg=0 and on_qrg=0 --> ok, we are out of tolerance
7041900 --> below_qrg=0 and on_qrg=1 (spot is at QRG+100) --> ok, within tolerance (Condition B applies)
7042000 --> below_qrg=0 and on_qrg=1 --> ok
7042100 --> below_qrg=1(!) and on_qrg=0 --> this is wrong as we are still within tolerance (Condition A applies)
7042200 --> below_qrg=1 and on_qrg=0 --> ok, we are out of tolerance

So it seems in Condition A the less-than-equal should be simply less.

zcsahok:
Well, things are bit more complicated: with the less in Cond A if there is spot at QRG and also 100 Hz below, then the one below it is selected:
image

zcsahok (approved):

zcsahok:
The tolerance/rounding story looks more complicated than I thought, will open a separate issue to review it.

dl1jbe:
Ok, lets delay the the rounding problem for now and merge the actual version in.

PR #509, opened 2026-06-16
Fix DXCC lookup

zcsahok:
DXCC lookup uses a cache for 2-character prefixes. It has an issue as all IT calls resolved as Sicily (IT9).
It is caused by too greedy caching: prefixes from more that 2-char calls shall not be cached, just used for marking the prefix as ambiguous.

Current implementation:
IT -> IT9

Should be:
IT -> ambiguous (as it could be a IT9 or any other ITx)

First added tests to reproduce the issue.

### DIFF
--- a/src/dxcc.c
+++ b/src/dxcc.c
@@ -33,7 +33,7 @@
 GPtrArray *dxcc;
 GPtrArray *prefix;
 GHashTable *hashed_prefix;
-int two_char_prefix_index[36 * 36];
+static int two_char_prefix_index[36 * 36];
 bool have_exact_matches;
 char cty_dat_version[12];   // VERyyyymmdd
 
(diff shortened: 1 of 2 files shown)


zcsahok:
Added the fix in `prefix_add()`

dl1jbe (approved):
Looks good and makes a lot of sense.

Just to be curious: We did quite some performance test during implementation of the old behavior. Do we have some data how your PR affect that?

zcsahok:
I'll check it.

dl1jbe:
> I'll check it.

Please, do it only if you are interested yourself. It was just a question.

zcsahok:
Dug out the original PR #240: the speed test code is the `test_xspeed()` in `test/test_getctydata.c`, just `TEST_LOOKUP_SPEED` has to be defined. It makes 20M random lookups based from the callmaster file.
Testing on a 20 yr laptop (Intel Pentium M processor 1.70GHz) used regularly for portable logging gave these figures (usec/lookup):

- 2-char cache disabled: 0.70
- with 2-char cache, as before this PR: 0.54
- 2-char cache fixed as per this PR: 0.54

I made 50 runs of the test code per case. The std deviation of the data was below 0.01, so there is no significant difference due to the fix. In reality the fix removes ~40 false positives (mostly Italian prefixes) from the ~530 cache entries, but the related effect is minimal.

The ~20% improvement due the 2-char cache is consistent with the results of #240.

When at it, I also looked at adding some single letter prefixes (F,G,I,...) to the 2-char and it seems feasible. Will address this in a separate PR.

dl1jbe:
That are very good numbers. Thanks.
I must confess that from looking into cty.dat i had expected that the PR would have only a limited effect. But good to have real numbers anyway.

See the approval from some days ago.

PR #510, opened 2026-06-29
Extend 2-char prefix cache with some single character prefixes

zcsahok:
As a follow-up from #509 the 2-char prefix cache is completed with available single char prefixes.

The 2-char cache is filled with the prefixes from cty.dat as follows -

1. single char prefixes are ignored. The reason is that there could be a more specific 2-char prefix in the list (.e.g F vs. FG).
2. two char prefixes are evaluated and stored if they are unique
3. prefixes longer then two chars  are marked as ambiguous without evaluation, thus forcing a full lookup for that 2-char prefix (e.g. IU9 makes IU ambiguous).

The change in this PR is that once all entries from cty.dat are processed we make a loop over the cache (_post_fill_two_char_prefixes()_) and try to evaluate the empty slots. This is done by constructing a call _ab0X_, where _a_ and _b_ correspond to he actual cache index.
If the lookup succeeds that _ab_ is marked in the cache.

Example: there was no reference to _F0_ , the lookup of _F00X_ provides France, so it is stored in that slot.
For _X1_ the _X10X_ results in no DXCC, so it is kept as unknown.

Not all lookups succeed, but around 60 entries are added from F, G/M, I, K/N/W, and R/U lands.
It brings a minor, but measurable improvement to the lookup speed: the figure of 0.54 us/lookup is reduced to 0.50.

### DIFF
--- a/src/dxcc.c
+++ b/src/dxcc.c
@@ -260,16 +260,21 @@ void prefix_add(char *pfxstr) {
 
     /* build 2-char prefix hash */
     int pfxlen = strlen(pfxstr);
+    // single char prefixes are ignored for the moment, they are handled
+    // after all prefixes are loaded in post_fill_two_char_prefixes()
     if (pfxlen >= 2) {
 	int key = prefix_hash_key(pfxstr);
 	if (two_char_prefix_index[key] == TCPI_NONE && pfxlen == 2) {
-	    two_char_prefix_index[key] = index;     // unique 2-char prefix
+	    // it's a unique 2-char prefix, store it
+	    // note that it can still be set to ambiguous later on
+	    two_char_prefix_index[key] = index;
 	} else {
-	    two_char_prefix_index[key] = TCPI_AMB;  // ambiguous
+	    // it's a prefix that has been already seen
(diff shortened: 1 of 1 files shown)


dl1jbe:
On first run I had trouble understanding how it works. 
As I understand now it registers the single character prefix for every combination _char|digit_ with digit=0..9 if the according slot is unused. So for example prefix 'F' gets recorded for key from 'F0', 'F1' and so on (if the according entry is empty).

Not sure if we should add the logic described in your PR above into the file for further reference. Maybe just add a link to the PR. 

Otherwise I approve the PR.

zcsahok:
Thanks for the feedback. I'll rework my code a bit to gather explicitly the single char prefixes instead of implicitly deducting them from the unused slots.
But yes, it registers France for all free F0, F1, ... combinations.

zcsahok:
Now the scanning of `post_fill_two_char_prefixes()` is done more targeted on single char prefixes, looking for unchecked letter+number combinations. This made the mapping function `get_base36_char()` unnecessary. Also added some comments.
I hope it makes more sense like this.

dl1jbe:
Congratulation. That is much clearer (and simpler) now. I think it is easy enough to understand.

PR #432, opened 2024-05-10
Update workflow to use Node.js 20

zcsahok:
A quick one to fix the warning

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3.

See also https://stackoverflow.com/questions/77897660/github-actions-node-js-16-actions-are-deprecated-warning

PR #511, opened 2026-07-04
Update workflow to use latest action/checkout

zcsahok:
Action runs get this warning (see https://github.com/zcsahok/tlf/actions/runs/28259700334)

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

We had a similar one in #432 where the version of _actions/checkout_ (https://github.com/actions/checkout) was bumped.
Let's try now to use the latest version of it and hope that it always takes the current Node version.

### DIFF
--- a/.github/workflows/ci-build-ubuntu-22.yml
+++ b/.github/workflows/ci-build-ubuntu-22.yml
@@ -16,7 +16,7 @@ jobs:
       AM_COLOR_TESTS: always
 
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


dl1jbe:
I am not good with actions yet. Let us try if it solves the problem in the long run.

Fyi, will be away next week.

zcsahok:
Thanks, merging it in.

PR #513, opened 2026-07-11
Fix info page

zcsahok:
Fixing two minor issues with the info page:

1. search panel stays open
2. rig port name has a black tail when TLF was called without rig control

The screenshot below shows what I mean. To reproduce it

- start with rig control disabled (`-r` option)
- enter a callsign
- type `:inf`

image

The first commit addresses the search panel issue.

It is now explicitly hidden before calling `changepars()`. As the search panel is triggered from `callinput()`, the hiding of it is also done there.
Removed the extra clearing stuff from the end of `changepars()` as this is anyway done in the `callinput()` loop.
Moved resetting of the current callsign before the call of `changepars()`, so that does have to bother with it.
Made `changepars()` void, as the return value had no meaning.
`HideSearchPanel()` does nothing in case the search panel hasn't been initialized yet.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -419,8 +419,9 @@ int callinput(void) {
 
 	    // Colon, prefix for entering commands or changing parameters.
 	    case ':': {
-		changepars();
 		current_qso.call[0] = '\0';
+		HideSearchPanel();
+		changepars();
 		x = 0;
 		clear_display();
 		attron(COLOR_PAIR(C_LOG) | A_STANDOUT);
--- a/src/changepars.c
+++ b/src/changepars.c
@@ -136,7 +136,7 @@ void set_trxmode(int mode) {
 }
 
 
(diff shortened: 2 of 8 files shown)


zcsahok:
The second commit addresses the trailing black characters after rig port.
It is caused by not removing the newline character from `rigport`. It was the legacy behavior and has been kept for "compatibility". It was commented that all non-chomping config readers shall be removed. There are only 2 left: `rigport` and `clusterlogin`.
`rigport` was in fact stripped before actual use in `init_tlf_rig()`.

Now this happens at configuration parsing, thus it was dropped in `init_tlf_rig()`. The macro `CFG_STRING_NOCHOMP` could also be dropped.

The test failed on purpose. The next commit will fix it.

dl1jbe (approved):
Good to finally clean up the newline mess from past times.

PR #515, opened 2026-07-17
Add IARU-HF rules

fillods:
It does not count the points and multipliers correct in all situations, but it will get you through the contest.

Based on versions from Fred DH5FS, Nate N0NB, and Koos PE4KH.

It has initial exchange for HQ stations and IARU officials, and let you track them as multipliers.

Next step is to convert scoring to a python plugin.


zcsahok:
I will have a look into the scoring plugin.

rules/iaruhf/iaruhf
```diff
@@ -0,0 +1,95 @@
...
+# but it will get you through the contest.
+##############################
+#
+# The custom multiplier list should be stored in the tlf working directory
+CONTEST=iaru-hf
+LOGFILE=iaruhf.log
+CONTEST_MODE
+
+CABRILLO=UNIVERSAL
+CABRILLO-CONTEST= (IARU-HF-CW, IARU-HF-SSB, IARU-HF-MIXED)
```
zcsahok, line 19: My log submission was accepted with
```
CONTEST: IARU-HF
```

I'm not sure if IARU-HF-CW/SSB/MIXED are valid as a quick search found no references to them.
The rules do not specify it, but at least here https://www.contestcalendar.com/contestdetails.php?ref=67 it states "IARU-HF".

zcsahok:
Regarding ITUMULT: it would maybe make sense to add a configuration ITUEXCHANGE (or similar) that would enable just the initial exchange generation logic without the whole multiplier machinery. That would save us typing in the exchange.

zcsahok:
Please check my PR containing the scoring https://github.com/fillods/tlf/pull/1 so that we can progress with this one.

zcsahok:
Friendly reminder to check https://github.com/fillods/tlf/pull/1. Thank you.

rules/iaruhf/iaruhf
fillods: Indeed, it's safer to stay with `IARU-HF`, fixed by b15aa061.

fillods:
Thanks @zcsahok for the plugin PR and the reminder 🙂

This PR is good to go !

PR #516, opened 2026-07-17
Allow seamless editing of call input field

zcsahok:
Rethinking #495, focusing here on call input field only. Later on `getexchnage` can be changed the same way.
The goal is to achieve seamless editing of the callsign input field using the usual editing keys, while keeping the other special keys operational. Taking the approach of integrating editing into `callinput()` as mentioned as an option in #495.

First dropping the unused the 'int insert_char(int)' function. At the end the whole `calledit.c` file will be dropped.


zcsahok:
Adding two helper functions
- `insert_char` inserts a character into a string at a given position. The part of the string right from position (cursor) is shifted right. It returns the new cursor position. If the string would overflow the available size, then no action is taken.
- `delete_char` removes the character at the given position. The part of the string right from the position is shifted left (like for the Delete key).

zcsahok:
Now incorporating the editing functions into the switch statement of `callinput`. We have to track the cursor position (`pos`) and sync it if changed by an external action like grabbing, (re)storing, or cleaning.

Ctrl-A from editing clashes with the one from call input. One has to use Home key to navigate to the start of the input field.

The different color scheme used when editing is now gone. Should be no issue.

Apart from the generic working at least these special cases have to be checked

- autosend
- block_part (see searchlog.c)
- callsign grabbing in digi mode

zcsahok:
USEPARTIALS can also lead to modification of the call field. In this case we move the cursor to end of the call, as if it would my manually entered.

In order to allow editing when using USEPARTIALS we have to block further autofilling by setting `block_part` once we detect that the cursor was moved left from the end of the call.

zcsahok:
Memory operations (pop/swap) will try to keep cursor position.
This can be changed, if needed.

zcsahok:
Moving cursor to the end of the call on Fldigi grab.

zcsahok:
Not directly related to editing, but fixed USEPARTIAL triggering on a numbers-only entry.
Current MASTER.SCP contains the call 8C**356**PDG and this made entering 80m frequency 356x impossible.

dl1jbe (approved):
Looks good.
The critical point seems the lot of 'pos = ...' manipulations. Let us hope all relevant cases are handled.

Do you want to merge it in now and do the edit of exchange in a second PR or do plan to add it to these PR also?

zcsahok:
Keeping track of `pos` is complicated with direct accesses tot the call out of the loop. The one from fldigixmlrpc is even done from the background thread. To be fixed later, until then it works, provided one does not use the keyboard and the mouse at the same time.

Exchange will the addressed in a separate PR, but based on the solution worked out here.

PR #517, opened 2026-07-17
Fix missing files in rules subdir that are not distributed

fillods:
Some files in the `rules` subdir are not making it into the dist tar ball.

This PR also add some precious hints for the iota rule, gleaned from a past PR.
Not really related, it also documents of a precious tool for handling Cbr and ADIF.

### DIFF
--- a/doc/Manual.md
+++ b/doc/Manual.md
@@ -1157,12 +1157,21 @@ the moment there are at least three different versions:
 Q: The generated ADIF file misses some entries, e.g. the CONTEST_ID which is
 used by my log program during import. Any idea to work around it?
 
-A: You can use 'adifmerg' from OH7BF for it. It allows to post process the
+A: You can use [adifmerg](https://github.com/oh7bf/adifmerg) from OH7BF for it.
+It allows to post process the
 ADIF file. For instance you can add the missing CONTEST_ID with:
 
-```
+```sh
 adifmerg -f qso.adi -A CONTEST_ID=<my contest name> -o > file.adi
 ```
+
+With the [ADIF Multitool](https://github.com/flwyd/adif-multitool) from WT0RJ,
+you can do it (and more!) with:
+
+```sh
+adifmt edit --add 'contest_id=<my contest name>' mylog.adi > file.adi
+```
+
 Have a look at [www.adif.org]() for the already standardized contest IDs.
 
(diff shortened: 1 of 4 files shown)


zcsahok:
RULES_FILE shall automatically contain all non-Makefile files from rules. Will check how to achieve this. Else we have to update Makefile.am each time a new rule file is added.

fillods:
Are you sure you want to pick up all the littering^W working files in the `rules` subdir ?

Speaking of my tree, it often has various test files, pieces of documentation, wip rules, etc.

zcsahok:
At least the main branch in the main repo shall have no wip stuff. Well, ideally. In addition the files in `rules` are mostly templates to be customized by the user, so even the unlikely case of delivering a bit more would be OK in my view.

Anyway, here is how to achieve packaging of all files/dirs in `rules` starting with a lower case letter -- partically covering all rule files --

- add these 2 lines to `configure.ac`:
```
RULES_FILES=$(cd rules; echo [[a-z]]*)
AC_SUBST([RULES_FILES])
```

- remove `RULES_FILES` from `rules/Makefile.am`

dl1jbe:
> At least the main branch in the main repo shall have no wip stuff. Well, ideally. In addition the files in `rules` are mostly templates to be customized by the user, so even the unlikely case of delivering a bit more would be OK in my view.
> 
+1

> Anyway, here is how to achieve packaging of all files/dirs in `rules` starting with a lower case letter -- partically covering all rule files --
> 
>     * add these 2 lines to `configure.ac`:
> 
> ```
> RULES_FILES=$(cd rules; echo [[a-z]]*)
> AC_SUBST([RULES_FILES])
> ```
> 
>     * remove `RULES_FILES` from `rules/Makefile.am`

Good idea. There are still some problems which need some more work.

* 'make dist' builds an distribution package (lf-1.5.~git.tgz) with all files included.
* unpacking the tar ball, configuring it followed by 'make' and 'make install' shows problems. Any directory in 'rules' and all of their content will not be installed.
* Further problem - the final 'make distcheck' has wrong list of files in RULES_DIR and complains that these files can not be found.

zcsahok:
Looks like it's not easy to solve, autotools is more tricky.
I'll add a check to the CI build to make sure that all rules files are included.
This also allows using local build and dist creation as before.

zcsahok (approved):

PR #523, opened 2026-07-22
Check that all relevant files are included in distribution

zcsahok:
Add make dist + distcheck to CI build. And also check that all rules are in tar.gz.

Changed only one job to see it failing.
Will change the other too, and then after #517 both shall be OK again.


zcsahok:
Failed with the expected error:
```
...
rules/ukeidx/ukeidx					          <
rules/ukeidx/ukeidx.py			              <

---- Files marked with '<' are missing from RULES_FILES in rules/Makefile.am
```

zcsahok:
Both builds are OK after cherry picking the commit fixing RULES_FILES ( https://github.com/Tlf/tlf/pull/517/commits/8762fe4597519e281dec982f3579b9d574c3fe54) from #523.

Proposing to merge in both PRs.

dl1jbe:
Looks good. +1 
The 'make distcheck' makes also sure, that we do not forget to add new source files to the Makefile.am.

dl1jbe:
By the way I just stumbled about the fact that we have the same problem for the test/rules directory which is also missing from the distribution package.

It seems it needs similar checks.

zcsahok:
True. Other directories shall also be checked.

zcsahok:
Changed main CI build first to see if it fails as expected.

zcsahok:
OK, found the missing files in test/rules.
Changing the other yml.

dl1jbe:
So, with that PR the CI systems checks that all relevant files are in the distribution package.

New rule files have to be added to rules/Makefile.am by hand and the test/rules file should be added by generate_defs.pl.

Just checked out your branch and tried an 'autoreconf'. The 'defs.am' got not updated, so the TEST_RULES_FILES=... are still missing. Am I doing something wrong?

dl1jbe:
Sorry, should be 'autoreconf -f' and that works as expected.

Maybe we should have a similar script doing the work for the rules directory itself.

zcsahok:
Fixed out-of-source build done by distcheck: test/rules directory had to be copied to the build tree to make it work.
Now distcheck is successful, so we could even make a release... ;-)
```
tlf-1.5~git archives ready for distribution: 
tlf-1.5~git.tar.gz
```

For the tests we must have the generator script, so I simply reused it to create the list of files in test/rules.
It could be done in share dir also, but for the moment manual update would be fine. Adding another script just increases complexity.

dl1jbe:
Fine. Lets merge it in. I will check in coming days if 'make distcheck' also works standalone out of the CI framework,

dl1jbe:
> Fine. Lets merge it in. I will check in coming days if 'make distcheck' also works standalone out of the CI framework,

Confirmed. Works as expected.

PR #526, opened 2026-08-03
Fix linking of  test files

### LINKED ISSUE #524: Test rules for CQP and NEQP contests are out of sync
While most of the test rules for supported contests refer into the actual /rules directory via linking both above mentioned contests are using separate copies of the rules in the /test/rules directory. Both rule sets differ in content.

zcsahok:
Fixes #524 

For NEQP there was just a minor difference. Also linked the data files.


zcsahok:
For CQP there was a new, more functional version in the test directory. Merged it back to rules and removed cqp_ca* files.
Also considered the recent rule change: SSB counts also 3 points.

dl1jbe (approved):

PR #529, opened 2026-08-16
Add deprecation support for configuration keywords

zcsahok:
Added KeywordDeprecated() function and per config_t entry specification if a keyword is deprecated.

Migrated USE_CONTINENTLIST_ONLY to this approach.

Updated rules to use the recently introduced multiplier configuration.
While touching naqp also simplified it to use the build-in ONE_POINT scoring.


zcsahok:
Test failed due to pacc_dx using deprecated USE_COUNTRYLIST_ONLY

dl1jbe (approved):
Clean solution. Fine.

PR #532, opened 2026-08-30
Seamless editing of exchange field

zcsahok:
Sequel to #516: adding seamless editing to the exchange field.
Will supersede #495.

Using the approach as for the call input field: tracking cursor position and using helper functions for editing.

### DIFF
--- a/src/getexchange.c
+++ b/src/getexchange.c
@@ -175,7 +175,7 @@ int getexchange(void) {
 		x = KEY_LEFT;
 		continue;
 	    }
-	    case 19: {	// Ctl+s (^S)--Open QTC panel for sending QTCs
+	    case CTRL_S: {	// Ctl+s (^S)--Open QTC panel for sending QTCs
 		if (qtcdirection == 2 || qtcdirection == 3) {	// in case of QTC=SEND or QTC=BOTH
 		    qtc_main_panel(SEND);
 		}
@@ -289,17 +289,35 @@ int getexchange(void) {
 		break;
 	    }
 
-	    /* <Home>--edit exchange field, position cursor to left end of field.
-	     * Fall through to KEY_LEFT stanza if ungetch() is successful.
-	     */
+	    /* <Home>--move cursor to the beginning of exchange field */
 	    case KEY_HOME: {
-		if (current_qso.comment[0] == '\0' || ungetch(x) != OK)
-		    break;
+		pos = 0;
+		break;
 	    }
 
-	    case KEY_LEFT: {	/* Left Arrow--edit exchange field */
-		if (current_qso.comment[0] != '\0') {
-		    exchange_edit();
+	    /* Ctrl-E (^E) or <End>, move to the end of exchange field */
+	    case CTRL_E:
+	    case KEY_END: {
+		pos = strlen(current_qso.comment);
+		break;
+	    }
+
+	    // <Delete>
+	    case KEY_DC: {
+		delete_char(current_qso.comment, pos);
+		break;
+	    }
+
+	    case KEY_LEFT: {	/* Left Arrow--move cursor left */
+		if (pos > 0) {
+		    --pos;
+		}
+		break;
+	    }
+
+	    case KEY_RIGHT: {	/* Right Arrow--move cursor right */
+		if (pos < strlen(current_qso.comment)) {
+		    ++pos;
 		}
 		break;
 	    }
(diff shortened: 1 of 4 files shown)


zcsahok:
Moving duplicated serial number alignment logic to a separate function.

BTW the EU Sprint contest is now defunct: https://www.contestcalendar.com/contestdetails.php?ref=316
The latest results from the original contest are from 2009. Then RSGB tried to resurrect it in 2016, seemingly with not much success. So we could even drop it (SPRINT).

N0NB:
Based on my experience this weekend I found Ctrl-W only worked in the call field and if I wanted to wipe a contact that I had to tab out of the exchange field.  Not a deal breaker but a bit of frustration to have to remember to do an extra key tap.

N0NB:
BTW, I had built Tlf from master branch yesterday morning.

zcsahok:
@N0NB I'll have a look. Ctrl-W clears the current field only. At least in this branch it works, but does not update the cursor position.

zcsahok:
Fixed cursor update after Ctrl-W

N0NB:
My understanding was that Ctrl-W was supposed to clear all fields and place the cursor in the callsign field.

zcsahok:
It's Ctrl-U that clears both fields. Ctrl-W clears the current field only.

N0NB:
I guess I need to RTFM a bit more carefully and adjust my muscle memory!

src/getexchange.c
```diff
@@ -334,7 +352,9 @@ int getexchange(void) {
 	    }
 	}	// End switch
 
-	x = g_ascii_toupper(x);     // Promote to upper case
+	if (isascii(x)) {
```
dl1jbe: From GLib documentation: 'g_ascii_toupper - Convert a character to ASCII upper case. If the character is not an ASCII lower case letter, it is returned unchanged.'

So the change is not really needed.

src/getexchange.c
```diff
@@ -334,96 +353,38 @@ int getexchange(void) {
...
-		addch(x);
-		strcat(current_qso.comment, instring);
-		refreshp();
+		pos = insert_char(x, current_qso.comment, pos, contest->exchange_width);
 	    }
 	}
 
 	/* <Enter>, <Tab>, Ctl-K, '\' */
 	if (x == '\n' || x == KEY_ENTER || x == TAB
 		|| x == CTRL_K || x == BACKSLASH) {
```
dl1jbe, line 367: Just wondering about the CTRL_K here. It gets handled before by activating the keyer, so it will never trigger the condition here.

I looked into history: 
* In old TLF-0.9.10 there was no keyer while in exchange field but CTRL_K got handled after leaving get_exchange() in the logit() function. 
* In 2019 we moved keyer handling into get_exchange() but did not drop it in logit().

I think we should leave it for now and clean up in another PR.

dl1jbe (approved):

src/getexchange.c
zcsahok: There is subtle issue with `g_ascii_toupper`, also mentioned in the docs:

> Also unlike the standard library function, this takes and returns a char, not an int, so don’t call it on EOF ...

The problem is that our key `x` is an int that is silently cast to a char, losing the higher bits and potentially landing in the ASCII range. Then g_ascii_toupper converts it to an upcase char that has no relation to the original key.
An example: KEY_END = 0550 --> 104 + 256. A plain g_ascii_toupper call delivers "H", since 104 is "h".
The `if` avoids calling g_ascii_toupper on a non-ascii key.
Maybe `toupper()` from ctype.h would be a better choice, provided it is really locale independent.

src/getexchange.c
zcsahok: Interesting. Ctrl-K is handled in the switch (line 333) and x is set to zero. So at line 368 x can never be CTRL_K, as far as I see.

src/getexchange.c
zcsahok: `callinput.c` uses a somewhat similar construct:
```
        if (valid_call_char(x)) {
            x = g_ascii_toupper(x);
```

dl1jbe: > The problem is that our key x is an int that is silently cast to a char, losing the higher bits and potentially landing in the ASCII range.

Ahh, yes I remember. We had that problem some time ago already in another context.
I just checked the 'isascii()' function. Man page (see CAVEATS) explicitly states that it works only for EOF or a value re-presentable as unsigned char. So it may fail on KEY_END and similar also.

src/getexchange.c
dl1jbe: Yes that was what brought me to my comment above. It is an leftover from old versions - see my comments wrt the history above.

zcsahok:
Added own `ascii_toupper()` that can handle ints without issues.

The open issues for later are then

1. review Ctrl-K handling; eventually drop it from logit.c + move to `handle_common_key()`
2. review usage of `g_ascii_toupper()`; migrate to `ascii_toupper()` is needed/possible

dl1jbe:
Fine. That addresses all open points. Lets merge it in.

PR #533, opened 2026-09-07
Simplify keyer activation

dl1jbe:
As discussed in PR #532 this is a rework of Ctrl-K and ',' handling in callinput() and getexchange().

It is handled now in one place - handle_common_keys().
Ctrl-K or ',' no longer leaves the according input loops.

### DIFF
--- a/src/callinput.c
+++ b/src/callinput.c
@@ -884,8 +884,9 @@ int callinput(void) {
 	    }
 	}
 
+	// leave callinput() function
 	if (x == '\n' || x == KEY_ENTER || x == SPACE || x == TAB
-		|| x == CTRL_K || x == ',' || x == BACKSLASH) {
+		|| x == BACKSLASH) {
 	    break;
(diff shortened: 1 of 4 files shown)


zcsahok:
Now comma in the keyer window opens another keyer window on top of the original one due to comma being handled in common code.
I see two options: either map comma to Ctrl-K (use for closing the keyer) or map to a special value in order not be consumed in common handler and then map back to comma for sending.

zcsahok:
For option 2 (keeping comma as normal character) would be maybe clearer not to call common key handling:
```
if (x != ',') {  // comma has no special function
   x = handle_common_key(x);
}
```

dl1jbe:
> ```
> if (x != ',') {  // comma has no special function
>    x = handle_common_key(x);
> }
> ```
Looks like the best/simplest solution.

zcsahok (approved):

PR #535, opened 2026-09-16
Add YUDX contest rules

zcsahok:
+ ignore pycache directory (created when running tests)


dl1jbe:
Looks good and works as expected.

One point to consider is the inconsistency for the location of the inital_exchange file. For most contests we distribute these file via the 'share' directory. In last time we added some of theses initial_exchange files (e.g. for 'mcd' or 'iaru_hf' in the contest specific 'rules' directory.
We should consider to harmonize that usage in either way.

zcsahok:
Good point regarding the structuring of rules and support files.

For the time being I'd keep it like it is, but there is definitely an improvement potential here.
On one hand rules (+ other files) could be also read from a subdir of rules/ to keep them localized.

The other topic is more generic. Currently our rules serve practically as samples to be edited by the user. So even though say `RULE=cwops_mini_cwt` would load that rule, it is of little use as it has the 'JO 2225' exchange hardcoded.
A typical logger user would want to select the contest and then customize it: select the category details and set the sent exchange (if applicable).
Category setting can be done in TLF by adding further configuration after RULES, e.g.
```
RULE=yudx
CABRILLO-BAND=40M
```
if only 40m operation is planned.

What one can't do is simply setting (overriding) the sent exchange, unless one changes F3, S&P_TU_MSG, and CABRILLO_EXCHANGE configs keeping them in sync.
Much easier would be if the `#` macro in the messages could refer to the actual value of CABRILLO_EXCHANGE (i.e. the value to be put into the log), with default value being the QSO sequence number. That would allow us to keep the definition of F3+S&P unchanged as configured by the user, and just by change a single config control the sent exchange.
Hope this makes sense.

It would also make possible to include both sides in asymmetric contests, like YU-DX (or typically all xx-DX).

PR #536, opened 2026-09-17
Add SAC contest

zcsahok:
For non-Scandinavian stations.


dl1jbe:
would 'sac_dx' a better name as it is for non-scandinavian stations? Otherwise looks fine.

zcsahok:
will add Scandinavian side too, the rules are quite similar

zcsahok:
Extended for Scandinavian stations and added a test for it. Sc/non-Sc distrinction is done automatically based on the call used.
The rules are a bit unique: a Sc-Sc QSO is worth 0 points, but the DXCC multiplier for it is valid.

dl1jbe (approved):
Looks good and complete. Just merge it in.

